Issue #248, opened 2015-09-16
statusDetails Field missing


timgdavies:
The documentation for the [Award Status](http://ocds.open-contracting.org/standard/r/1__0__0/en/schema/codelists/#award-status) codelist entry for 'unsuccessful' states that: 

> "This award could not be successfully made. If items or supplier details are included within the award section, then these narrow the scope of the unsuccessful award (i.e. the award of noted items, or an award to the noted supplier, was unsuccessful, but there may be other successful awards for different items listed in the tender, or to different suppliers). The reason that the award did not suceed will be given in the statusDescription field."

Apart from the type in succeed, we also have the issue that no '`statusDescription`' field exists in the schema. 

This appears to be a bug - and such a field should have been created.

timgdavies:
(statusDescription was proposed in #143 but it appears it was not incorporated into the final schema)

jpmckinney:
It's not clear that #143 had any concensus about `statusDescription` - hence it wasn't added (I think you were the only person talking about it, Tim :) )

jpmckinney:
But, FWIW, `statusDetails` would be more consistent with other field names.

timgdavies:
See pull request: #257

timgdavies:
Ok - renaming this to statusDetails and pushing to 1.1 discussions.

timgdavies:
The phrase "The reason that the award did not suceed will be given in the statusDescription field." has been removed from the codelist in 1.0.1 bugfix release.

Views for, or against, inclusion of `statusDetails` in 1.1 are welcome. 

We would need to see interest in this field, and examples of data that would fill it, to take forward.

timgdavies:
There have been no voices in favour of ```statusDetails```, nor further requests for this field. 

This will not be actioned as part of 1.1. Closing.

PR #257, opened 2015-10-27
There is no statusDescription field in any version of the standard

gabelula:
It would be useful to have an awardStatusDetails (as free text) to add information about why an award was unsuccessful or anything related with the status but so far we don't have it.


timgdavies:
Thanks for the pull request on this. I'll flag this for the 1.1 discussion, as we'll need to agree this before the pull request is accepted.

timgdavies:
Reviewed this, and understood that notes change should be made, whilst description being included should be an issue in 1.1. @bjwebb to sort out pull request.

Bjwebb:
New pull request for the codelist change at https://github.com/open-contracting/standard/pull/307. There's also an issue about actually adding the new field at https://github.com/open-contracting/standard/issues/248

PR #281, opened 2015-12-21
Fix some links in the Reference documentation

jpmckinney:


timgdavies:
Thanks @jpmckinney. Just working on updated docs so this will get reflected on master now, and in main docs in Jan.

PR #294, opened 2016-02-05
Fixing links to documentation in the schema

timgdavies:
I've updated the links to documentation in the schema to point to the new docs.

I've not updated the internal links between schema (E.g. links to http://ocds.open-contracting.org/standard/r/1__0__0/release-schema.json in `record-package-schema.json`).

Not sure if these need to be updated or not at this point, and if so, any considerations we should have for making those updates.

### DIFF
--- a/standard/schema/record-package-schema.json
+++ b/standard/schema/record-package-schema.json
@@ -74,7 +74,7 @@
             "properties": {
                 "ocid": {
                     "title": "Open Contracting ID",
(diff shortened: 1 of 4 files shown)


Bjwebb:
I'd leave the internal schema links as is unless we change the schema IDs.

Also Sphinx doesn't like `//` in urls, so I've removed those (guessing this is a find and replace artefact).

PR #299, opened 2016-02-15
Correcting link to validator

timgdavies:
The link to the validator on the 'Validate' page was wrong.


Bjwebb:
Fixed on live now http://standard.open-contracting.org/latest/en/getting_started/validation/

PR #360, opened 2016-08-02
Update CSV serialization section

Bjwebb:
Preview at http://standard.open-contracting.org/1.0-dev-serialization/en/implementation/serialization/#csv

### DIFF
--- a/standard/docs/en/implementation/serialization.md
+++ b/standard/docs/en/implementation/serialization.md
@@ -4,7 +4,7 @@ The Open Contracting Data Standard provides a **structured data model** for capt
 
 The current canonical version of this data model is provided by a **[JSON Schema](../../../schema/release)** which describes field names, field definitions and structures for the data. The compliance of data with the Open Contracting Data Standard will be assessed against this schema.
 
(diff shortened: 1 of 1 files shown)


standard/docs/en/implementation/serialization.md
```diff
@@ -94,14 +94,18 @@ Instead, data with a one-to-many relationship can be represented using a multi-t
...
 
-The OCDS approach to multi-table flat serializations is currently under review. Contact the [helpdesk](../../../support/) if you wish to make use of this. 
+An example multi-table template can be found [in the sample data repository](https://github.com/open-contracting/sample-data/tree/master/flat-template).
 
-### Packaging files with meta-data
+For further information on multi-table serializations please see the [flatten tool documentation](http://flatten-tool.readthedocs.io/en/latest/).
+
+
+
+## Packaging files with meta-data
```
Bjwebb, line 107: Changed from 3rd level to second level section as it's not CSV specific. http://standard.open-contracting.org/1.0-dev-serialization/en/implementation/serialization/#packaging-files-with-meta-data

@timgdavies Could you check this makes sense. I did mention it in our working document, but I don't think it was clear what I meant.

Issue #303, opened 2016-02-23
The versioned release schema should be consistently sorted


Bjwebb:
This would allow more easily understood diffs of that schema.

Bjwebb:
Looks like https://github.com/open-contracting/standard/issues/285#issuecomment-171366799 may also solve this.

timgdavies:
@bjwebb Can you confirm if this is covered by #285, and if so, close this issue?

Bjwebb:
No, there was still some more to do, but it's happened as part of https://github.com/open-contracting/standard/pull/412

PR #412, opened 2017-01-06
Make make_validation_schema simpler and useful for extensions

kindly:
Make it use the merge stratagies properly so that it does not hard code
the field name.

Make the unversioned definitions generate automatically, so are not
hardcoded either.

Try and maintain order so diff is not so bad, even though there are few
reordings that were too much work to avoid.


Bjwebb (approved):
This looks good, I've added a commit to ensure it's consistently sorted (see also https://github.com/open-contracting/standard/issues/303) as discussed last week.

I think this is now ready to merge onto 1.1-dev, but could you double check @kindly (and merge if you want).

timgdavies (approved):

PR #418, opened 2017-01-31
Check for duplicate keys in the schema

Bjwebb:
Creating a PR so I can find this again.

Tests fail because 1.1-dev branch fails.

@timgdavies I think you're planning to do some work on the 1.1-dev branch (updating your script, so that release-schema.json is correct) before we merge this, so assigning to you to ping me once that's done, and I can rebase.

### DIFF
--- a/standard/schema/tests/test_validate_schema.py
+++ b/standard/schema/tests/test_validate_schema.py
@@ -2,6 +2,25 @@
 import json
 from path import path
 from jsonschema import validate
+from collections import UserDict
+
+class RejectingDict(UserDict):
+    """
+    A dict that only allows a key to set once.
+    Lets us raise an error on duplicate keys in JSON.
+    """
+    def __setitem__(self, k, v):
+        if k in self.keys():
+            raise ValueError("Duplicate key in JSON")
+        else:
+            return super().__setitem__(k, v)
+
+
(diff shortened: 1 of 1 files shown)


timgdavies:
I think the issue here was not that release-schema.json was wrong, but that I'd not run the:

````
cd standard/schema/utils
./make_validation_schema.py
````

stage. 

I think that script might need some attention anyway with the upgrade, as it looks like it has some custom handling for organisations etc. that we need to be aware of. 

However, for now I've run ```make_validation_schema.py``` and then pushed an updated commit that passes the tests.

Bjwebb:
The lack of validation schema is why the tests were failing.

However, release-schema.json is missing this update https://github.com/open-contracting/standard/commit/18c80d3dd787ba3b1077aeb6b3ffdff3f283c16b

Bjwebb:
Additionally, I think the work that needs to be done to update that script is in this pull request https://github.com/open-contracting/standard/pull/412 (so done, just waiting to be reviewed and merged).

Bjwebb:
I've rebased this, and I think its now ready to merge, to make sure this test gets run for future schema changes.

timgdavies (commented):
Looks good to me. Merge when ready.

Issue #288, opened 2015-12-22
Stricter Validation on Strings and Lists


kindly:
This is under consideration for a core update in version 1.1 of OCDS.
## The issue

If a string or a list is a requiredProperty jsonschema only checks to see if the key is there.
So if there is an empty string or list against that key then it does not give you the errors you want.

Examples:
- ocid
- tags
- ids (all of them) 
## The proposal

We should check for length > 0 for these cases.

The schema should be updated to require this.

timgdavies:
I would support this proposal.

timgdavies:
The version 1.1 schema now includes minLength properties for required properties

PR #431, opened 2017-03-06
[#288] stricter-lists-strings

kindly:


timgdavies:
We've missed the official deadline for 1.1 updates for review... but I'll see if we can get this into the final version.

timgdavies:
Good to merge but might be some new strings that need this too now?

kindly:
I reran the script this morning so unless the script is not picking up ones it should then it should be up to date.

PR #437, opened 2017-04-12
Update sphinxcontrib-jsonschema commit to add collapse directive

Bjwebb:


Bjwebb:
Build currently fails because there is an Unresolvable JSON pointer in one of the extensions.

PR #439, opened 2017-04-19
Draft of updates to relatedProcess

timgdavies:
In this edit I've suggested refactoring relatedProcess so that it has schema and identifier, rather than OCID, and I've updated a number of other definitions.

We would then layer onto this an extension to add 'description' and 'estimatedValue' fields required by EU and some PPP use-cases.

### DIFF
--- a/standard/docs/en/implementation/index.md
+++ b/standard/docs/en/implementation/index.md
@@ -11,5 +11,6 @@
    serialization
    hosting
    registration
+   related_processes
    
 ```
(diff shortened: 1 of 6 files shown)


duncandewhurst (changes requested):
Shouldn't the enum for `relatedProcess/scheme` be replaced with a codelist csv and an opencodelist boolean?

To meet the use case for Red Compartida - scheme would need to be an open codelist, so for example compranet could be listed as the scheme with the compranet ID given.

Also recommend updating the description of `relatedProcess/uri` to read "machine or human readable"

timgdavies:
@duncandewhurst Thanks for those points. I've made those changes and added some documentation pages also.

duncandewhurst (approved):

PR #441, opened 2017-04-19
Update sphinxcontrib-jsonschema commit to change table layout

Bjwebb:
Example of what this looks like: http://standard.open-contracting.org/jsonschema-directive/en/implementation/ppp_framework/

(Edit: this is a PR against the ppp-extension branch, 1.1-dev now has a more up to date commit on it. This PR probably wants updating to the latest commit before merging, or closing and 1.1-dev merging into ppp-extension.)


Bjwebb:
PPP docs are now at https://github.com/open-contracting/public-private-partnerships and https://github.com/open-contracting/public-private-partnerships/blob/master/requirements.txt#L29 has a later commit pinned.

PR #454, opened 2017-05-10
Fix translation support for jsonschema directive tables

Bjwebb:
It's patchy atm, because of the work to do translations for 1.1, but you can see there's now some translations at http://standard.open-contracting.org/1.1-dev-jsonschema-directive/es/schema/reference/


timgdavies:
Great. I hadn't spotted that we'd need to do this - so thanks for working that through.

PR #462, opened 2017-05-26
Merging 1.1-dev into 1.1 to prepare for deployment. 

timgdavies:


timgdavies:
I've realised I need to do this using an 'ours' merge strategy, so will try at the command line locally.

Issue #251, opened 2015-10-05
Should schema allow amendment: null


timgdavies:
In validating a draft data file I've come across a case where, rather than containing an ammendment object, `amendment` is set to `null`.

This fails validation, as "None is not of type 'object'"

However, the [guidance on emptying fields](http://ocds.open-contracting.org/standard/r/1__0__0/en/schema/reference/#empty-fields) would appear to suggest that it is valid to have an object set to null.

Should we allow null objects?

gabelula:
Yes. I think it is easier to read (and download) to have a lighter file. Null is not adding any information to the contract. I would remove the null object.

timgdavies:
@bjwebb Could you have a quick check on whether this issue has been addressed, or needs to be taken forward into 1.1?

Bjwebb:
@timgdavies This has not yet been address, so needs to be taken forward.

timgdavies:
I think I need to run through this with someone. 

With the changes in #357 the amendment object is deprecated, and replaced with an amendments array - and as I understand our approach to blanking out array contents is to include an entry with just ```id``` 

Does this dodge the issue? Or do we still need to do something to allow amendment objects themselves to be null?

Bjwebb:
I'm not sure, but I think this problem might exist for other objects in the standard, e.g. value, not just amendment.

timgdavies:
What would it take to test for this and check if any further fixes required?

Bjwebb:
I've looked through the schema again, and the only non-nullable, but not required field I can see is roles. https://github.com/open-contracting/standard/pull/482

PR #482, opened 2017-06-21
[#251] Make roles nullable

Bjwebb:
#251


timgdavies:
Handled in separate commit to 1.1.1-dev.

Issue #484, opened 2017-06-23
OrganizationReference id fields are required in 1.1 schema, which means valid 1.0 data can be invalid against it


Bjwebb:
I ran into this whilst looking at this faked sample data - https://github.com/open-contracting/extension_tester/blob/c16bdd581c18af760dca015ba8230b683564917d/ocds_extension_tests/fixtures/fullfakedata.json - which validates for 1.0 but not 1.1.

timgdavies:
Ok - for backwards compatibility then we should remove that one requirement. I'll look into that now.

robredpath:
@timgdavies is this backwards-compatibility or forwards-compatibility?

timgdavies:
I've been working on the assumption that anything valid against 1.0 should be valid against 1.1 - with at most deprecation warnings on things to change. 

I've put in a branch and pull open a pull request which we can review as part of a collection of 1.1.1 issues.

robredpath:
Cool, thanks @timgdavies

PR #486, opened 2017-06-26
Removing minLength and required from id on OrganizationReference. Clo…

timgdavies:
…ses #484


timgdavies:
Handled in a separate commit on 1.1.1-dev now.

Issue #485, opened 2017-06-26
Make sure that all text is translated correctly


Bjwebb:
Using [podebug](http://docs.translatehouse.org/projects/translate-toolkit/en/latest/commands/option_rewrite.html) we can create a pseduo translation, which we can use to check for gaps.

Here's a build of what this looks like: http://standard.open-contracting.org/pseudo/en_pseudo/

Not translated properly:
- [x] Note boxes
- [x] Framework types box http://localhost:8023/build/en_pseudo/implementation/related_processes/#framework-types
- [x] Additional blue boxes on schema reference page [Note: community extension descriptions are deliberately not translated in transifex]
- [x] Extensions tables
- [x] Bid statistics codelist shouldn't translate codes...

Translating some of the above properly requires upgrading Sphinx and recommonmark, which requires other changes:
- [x] Fix the uses of csv-table
- [x] Push fixes to extensions back to the extension repos

Even with this upgrade, we're currently stuck on Sphinx 1.5.1 due to this issue with headings: https://github.com/rtfd/recommonmark/issues/77, and a slightly old version of recommonmark due to changes to how links work.

Pull request now at https://github.com/open-contracting/standard/pull/487

Bjwebb:
Pull request now at https://github.com/open-contracting/standard/pull/487

Bjwebb:
I've updated the bid extension repo with my changes: https://github.com/open-contracting/ocds_bid_extension/commit/85fc7eb6a34506facc916141946b0570429918d1

The other change was to the location extension, but the table there is missing from the repo, so I've not tried to do anything with it.

PR #487, opened 2017-06-27
Make sure that all text is translated correctly

Bjwebb:
#485 

I've been checking this using an auto-generated pseudo translation - http://standard.open-contracting.org/pseudo/en_pseudo/ - https://github.com/open-contracting/standard/commit/6f15cb54f3583eeac52c12645950d966d0453ba8

ToDo after merge:
- [x] Push the new strings up to transifex.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,5 +7,6 @@ venv
 
 standard/docs/field_definitions
 standard/schema/codelists_translated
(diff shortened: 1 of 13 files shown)


Bjwebb:
Merged, because translations are happening on transifex, and I wanted these translations pushed sooner rather than later.

PR #488, opened 2017-07-03
Allow breaking words for urls in extension json-highlight

edugomez:
This avoids an odd layout in the json box for `extensions` when adding urls by selecting from the core and community tables.


timgdavies:
Reviewed. This looks good.

PR #497, opened 2017-07-18
Fix schema links

duncandewhurst:
The links to the json schemas in the live version of 1.1 are broken - this PR fixes them

@Bjwebb @timgdavies checking that merging this into 1.1-dev is the right way to get it into the 1.1.1 bugfix


Bjwebb (approved):

Bjwebb:
Looks right to me. Since this is docs changes they could also go out before 1.1.1

duncandewhurst:
@Bjwebb yes please - do I need to do anything to make that happen?

PR #503, opened 2017-07-21
Add docstrings to utils

jpmckinney:
This PR is to help any new developers find their way around the existing scripts.

It doesn't seem like `upgrade_to_multilingual_string.py` has been run? Given that uncertainty, I didn't add a docstring.


timgdavies:
This looks good. It also shows up some scripts we can now remove from 1.1 as they are no longer used: so I'll do that now.

PR #504, opened 2017-07-21
Use Flake8

jpmckinney:
I ran flake8, corrected all warnings, and added it to Travis to catch future warnings.


Bjwebb (approved):
Looks good.

I'm not the greatest fan of some of the line truncation: I try and avoid breaking with backslashes, and use Django's line-length of 119 if enforcing that.

That's splitting hairs though, would be happy to see this merged.

jpmckinney:
@Bjwebb Ah, yeah, I hadn't configured autopep8 to control for that. We can configure `tox.ini` so that flake8 doesn't issue warnings if line-length is <= 119 in the future.

PR #505, opened 2017-07-22
Updates to readme and docstrings; removing old utility script

timgdavies:
Building on @jpmckinney's updates, I've made a few further docstring updates and removed a script that is no longer used.


standard/schema/utils/README.md
```diff
@@ -14,26 +14,4 @@ This can be created by running:
 
 There are tests set-up to ensure this has happened. 
 
-### make_field_definitions.py
```
jpmckinney, line 17: Since `make_field_definitions.py` is still used by `build_docs.sh`, should we add these docs to its file, if we're removing them from the readme?

jpmckinney: Oh, oops, just saw #506 :)

Issue #475, opened 2017-06-15
Review inclusion and validation of enum lists in schema


timgdavies:
Based on #471 and #336 we need to:

* [x] Check that we are not including enum lists in the schema for open codelists
* [ ] Check we **are** including enum lists in the schema for closed codelists
* [ ] Check that our closed codelists in the schema match the csv files that define the codes

PR #509, opened 2017-07-24
Adding script to keep closed codelists in sync

timgdavies:
To address #475 I've put together a small script which tries to manually sync a number of codelist. 

Will merge into 1.1.1 this afternoon.


jpmckinney:
Be sure to run flake8 on the new script :) This seems to correct the flake8 issues:

    autopep8 --max-line-length 119 --in-place standard/schema/utils/update-closed-codelist-enums.py

I'm guessing make_validation_schema.py needs to be run to get tests to pass.

Also, we use underscores instead of dashes for other scripts' filenames.

jpmckinney:
I may be reading the diff incorrectly, but did the `milestoneType` codelist vanish from `Milestone.type` in the validation schema? (On second look, maybe it's fine.)

timgdavies:
I believe so - but ```Milestone.type``` is an open codelist, so we don't want the enum.

PR #513, opened 2017-07-25
Updates to changelog.md

duncandewhurst:
All docs links in the changelog were pointing to 1.1-dev

This update points links in the 1.1.1 section to 1.1.1-dev and links in the 1.1.0 section to 1.1


timgdavies:
To avoid this requiring lots of manual edits to all the translations - we need to either:

* Edit the .po files for the languages as well
* Push the changes to transifex and manually go through the Spanish and French to update the links

Can you have a go at option (1)?

timgdavies:
Ok - so option (2) doesn't avoid lots of manual edits - it means we do them rather than asking the translators to... which we would certainly want to avoid for the French at least

Bjwebb:
If we edit the .po files, then we'd want to push those edits to transifex. And try to pull from transifex shortly before doing so, to avoid clobbering stuff.

duncandewhurst:
@timgdavies I've manually updated the .po files in this branch with the links fixes

Those .po files haven't been updated to include the new strings from the 1.1.1 changelog

Let me know if I should download the latest .po files from the Transifex interface and update those instead

timgdavies:
I've merged and pushed the updated .po files for the translations to Transifex.

PR #518, opened 2017-07-26
BidStatistic should be BidsStatistic

jpmckinney:
See https://github.com/open-contracting/ocds_bid_extension/pull/9


duncandewhurst:
thanks @jpmckinney 

I merged your PR against the bids extension repository which fixes this. The build process will pull the updated readme.md in to this repository, so no need to update it here. I'll leave this PR open so we don't lose the updates to the .po file

Issue #501, opened 2017-07-20
Links in schema in /latest point to /1.1 and links in /1.0 point to /latest


duncandewhurst:
In the /latest version of the docs, links in the schema point to the /1.1 version of the docs, e.g. in [/latest/en/schema/reference](http://standard.open-contracting.org/latest/en/schema/reference/#release) the link in the description of ```ocid``` points to [/1.1/en/schema/identifiers](http://standard.open-contracting.org/1.1/en/schema/identifiers/)

Whilst in the /1.0 version of the docs, links in the schema point to the /latest version of the docs, e.g. in [/1.0/en/schema/reference](http://standard.open-contracting.org/1.0/en/schema/reference/) the link in the description of ```ocid``` points to [/latest/en/schema/identifiers](http://standard.open-contracting.org/latest/en/schema/identifiers/)

This behaviour seems back to front, i.e. I would expect latest -> latest, 1.0 -> 1.0 etc.

timgdavies:
@bjwebb Can you give a quick estimate on how complicated it would be for us to add a search/replace process to the documentation build? So that links get updated to reflect the branch they are on?

Bjwebb:
This should be reasonably straightforward to do, a couple of hours work.

timgdavies:
Thanks - I've opened a non-sprint ticket to see if we can tackle this in the next couple of weeks.

timgdavies:
We also want to check the links in changelog are turned into relative links (and ideally to do this in a way that does not upset all the translations)

timgdavies:
This is a nice-to-have, but not a blocker for 1.1.1

Bjwebb:
Remaining tasks:
- [x] Sort out translations
- [x] Work out how we need update the deploy, so that latest is built on a latest branch.

PR #519, opened 2017-07-28
Add a replace of version/lang in links to schema translation 

Bjwebb:
From https://github.com/open-contracting/standard/issues/501

This should now be ready to merge.
I've pushed the updates to transifex now as well.


timgdavies (approved):
Tested through the Travis build, and reviewing the strings in Transifex.

PR #522, opened 2017-07-31
Fix broken link to identifiers guidance

duncandewhurst:
Fixing broken link in the first line of: http://standard.open-contracting.org/latest/en/schema/reference/#identifier

@timgdavies not sure whether this can be merged into 1.1.1-dev now the announcement has gone out?


timgdavies:
Yes. I don't think we need process for simple link fixes like this. Only for anything that impacts upon the meaning and interpretation of the standard.

PR #534, opened 2017-08-07
Pinning all extensions to 1.1.1 versions

timgdavies:


duncandewhurst:
@timgdavies as far as I can see this PR points the main docs to build at the [v1.1.1 branch of the extension registry](https://github.com/open-contracting/extension_registry/tree/v1.1.1)

However that branch of the extension registry doesn't look like it has been updated to point to the v1.1.1 releases of the individual extensions.

So the 1.1.1 build will be missing the typo fixes from @jpmckinney (and other changes) in the individual extension repos.

**Example:**

In the v1.1.1 branch of the extension registry, [extension.json for the participation fee extension](https://github.com/open-contracting/extension_registry/blob/v1.1.1/participation_fee/extension.json) points to `"https://raw.githubusercontent.com/open-contracting/ocds_participationFee_extension/v1.1/"` rather than the [v1.1.1 release of the participation fee extension](https://github.com/open-contracting/ocds_participationFee_extension/tree/v1.1.1)

Let me know if I misunderstood, or if I should go through and update the extension.json files in the 1.1.1 branch of the extension registry (cc. @kindly @Bjwebb )

timgdavies:
@duncandewhurst Ah - I thought the 1.1.1 extension registry branches had been updated.

I think if you can go through https://github.com/open-contracting/extension_registry/tree/v1.1.1 and update **core** extensions to point to 1.1.1 versions. 

The same should also happen in the master branch. 

We'll then need to re-build the live docs.

duncandewhurst:
@timgdavies see https://github.com/open-contracting/extension_registry/pull/55 for the update to the V1.1.1 branch

For the master branch of the extension registry, currently the `entry.json` files are pointed at the master branches of the individual extension repos, do we definitely want to change that?

duncandewhurst:
> For the master branch of the extension registry, currently the entry.json files are pointed at the master branches of the individual extension repos, do we definitely want to change that?

Discussed with @timgdavies and we don't need to change these.

PR #520, opened 2017-07-31
1.1.1 Release

timgdavies:
You will find OCDS 1.1.1 bug-fix release to preview at http://standard.open-contracting.org/1.1.1-dev/en/ with a [changelog here](http://standard.open-contracting.org/1.1.1-dev/en/schema/changelog/) and pasted below. 

You can also see a log of all change commits below.

Please use this pull request/issue to raise any questions or concerns about these changes. 

Patch releases are backwards compatible (or correct unintentional errors in earlier schema or documentation releases), and so do not go through the full upgrade process. **This patch release will become the main versions for 1.1 in 7 days, unless substantive objections are raised.**

This should not impact on existing published data (apart from where schema errors have been causing validation problems), although publishers may wish to review the changelog to identify areas where they can improve the alignment of values to updated open codelists.

### Bug fixes

* [#482](https://github.com/open-contracting/standard/issues/482) **[Allow parties.role to be set to null](https://github.com/open-contracting/standard/pull/502/commits/475abf598063aae5c22e07baba015d960fcc3a95)** - required by the [merging approach](http://standard.open-contracting.org/1.1.1-dev/en/schema/merging/). 
* [#422](https://github.com/open-contracting/standard/issues/422) **[Including currency codelist in documentation and schema](http://standard.open-contracting.org/1.1.1-dev/en/schema/codelists/#currency)** to support validation of currency values.
* [#479](https://github.com/open-contracting/standard/issues/479) **[Fixing incorrect requirement to include releases in record-package-schema.json](http://standard.open-contracting.org/1.1.1-dev/en/schema/record_package/)** 
* [#475](https://github.com/open-contracting/standard/issues/475) **Adding enum arrays to all fields in the schema with a closed codelist** to support validation.

### Minor revisions

* [#471](https://github.com/open-contracting/standard/issues/471) **[Updating milestoneType codelist](http://standard.open-contracting.org/1.1.1-dev/en/schema/codelists/#milestone-type)** replacing 'planning' with 'preProcurement' and 'adjudication' with 'assessment' and introducing codes for 'approval' and 'financing'. This is an open codelist, so previous codes remain valid, but publishers able to update to the new codes should do so. 
* [#473](https://github.com/open-contracting/standard/issues/473) **[Updating definition of contractSchedules in documentType codelist](http://standard.open-contracting.org/1.1.1-dev/en/schema/codelists/#document-type)**

### Documentation improvements

* **Fixing typographic errors throughout the documentation and codelist descriptions**
[...]

PR #535, opened 2017-08-07
Merge 1.1-dev into 1.1

timgdavies:
@Bjwebb Can you review and merge, and then update this ticket with a checklist of all the stages you carry out as part of making this the latest version, so that we can capture that in the documentation.

I believe this should include creating a 1.1.1 release, as well as updates to the version switcher. 

PR for consensus was #520


Bjwebb:
From https://github.com/open-contracting/standard/blob/1.1-dev/DEPLOYMENT.md

Before merge:
- [x] [Push translations to Transifex if any text has changed](https://github.com/open-contracting/standard#translations)
- [x] [Pull completed translations from Transifex](https://github.com/open-contracting/standard#translations) _[see note below]_
- [x] If the version has changed, remember to updated the schema ids and refs
- [x] Set up a dev instance of Cove using the new schema, and run tests against it _https://travis-ci.org/OpenDataServices/cove/builds/261861531_

After merge:
- [x] Copy the files from dev to live following the instructions at the top of https://github.com/OpenDataServices/opendataservices-deploy/blob/master/salt/ocds-docs-live.sls
- [x]  ~~For a new translation, edit http://standard.open-contracting.org/robots.txt~~
- [x] Create x__y__z schema folder - https://github.com/OpenDataServices/opendataservices-deploy/blob/master/salt/ocds-docs-live.sls#L25
- [x] Update live Cove to use the new schema _https://github.com/OpenDataServices/cove/pull/808_



....
- [x] Update version

Bjwebb:
Note: French translations are not completed. Spanish translations are completed, except for additions to the changelog (I've added a few translations to other pages, where it's obvious from the suggestions what the text should be). My understanding is that we're happy to go live like this, particularly since French translations weren't previously complete for 1.1 anyway.

timgdavies:
Correct. French translations will come in a few weeks.

For outstanding changelog items in Spanish we can get those covered in the same timeframe.

PR #542, opened 2017-08-10
Markdown linting

jpmckinney:
More importantly than the style fixes, the added testing will prevent Markdown errors.


jpmckinney:
Will re-do at a later date, possibly with more relaxed rules.

PR #550, opened 2017-08-23
Fetch markdown files from v1.1.1 READMEs of core extensions

Bjwebb:
- [x] Create metadata section
- [x] Check that the built docs look okay

TODO after merge:
- [x] Document script in the standard development handbook

### DIFF
--- a/standard/docs/en/extensions/bids.md
+++ b/standard/docs/en/extensions/bids.md
@@ -1,21 +1,6 @@
 Bid statistics and details
 ==========================
 
-## Metadata
-
-To use this extension, include its URL in the ```extension``` array of your release or record package. 
-
-```json
-{
-    "extensions":["https://raw.githubusercontent.com/open-contracting/ocds_bid_extension/v1.1/extension.json"],
-    "releases":[]
-}
-```
-
-This extension is maintained at [https://github.com/open-contracting/ocds_bid_extension](https://github.com/open-contracting/ocds_bid_extension)
-
-## Documentation 
-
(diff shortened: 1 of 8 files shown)


jpmckinney:
@Bjwebb Is there a script that does this? I can't seem to remember seeing relevant documentation.

Bjwebb:
I think this was previously manual.

I've started writing a script for this pull request https://github.com/open-contracting/standard/pull/550/files#diff-938ec24ad4e51596153df9258530240f, but it looks like there's some extra work to do to create the Metadata section.

jpmckinney:
Ah, I missed it. Can we add that script to the upgrade workflow documentation? https://github.com/open-contracting/standard-development-handbook/issues/4

Bjwebb:
I think this is now ready for merge, could @timgdavies and/or @duncandewhurst double check?

The one thing I don't like is the TODO sections are pulled in, but I think this is preferable to being out of sync with the extension repos.

duncandewhurst (changes requested):
Changes to the markdown files for the extensions look good, but noting that the links to the individual extension pages on http://standard.open-contracting.org/extensions-1.1.1-md/en/extensions/#core-extensions are set up to point to the /1.1/ build of the docs, rather than the build the user is browsing

Bjwebb:
@duncandewhurst Looks like that's not a new issue, so I'm not sure this pull request is the best place to fix it. The current behaviour is what it is because the url comes from the extension registry (e.g. https://github.com/open-contracting/extension_registry/blob/v1.1.1/bids/extension.json#L3).

duncandewhurst (approved):

PR #555, opened 2017-08-30
Pull French translations from transifex

Bjwebb:
http://standard.open-contracting.org/1.1-dev-fr/fr/


Bjwebb:
There's some broken markdown:
http://standard.open-contracting.org/1.1-dev-fr/fr/schema/merging/
http://standard.open-contracting.org/1.1-dev-fr/fr/implementation/related_processes/
http://standard.open-contracting.org/1.1-dev-fr/fr/implementation/amendments/
http://standard.open-contracting.org/1.1-dev-fr/fr/extensions/milestone_documents/
http://standard.open-contracting.org/1.1-dev-fr/fr/extensions/process_title/

Bjwebb:
Here's a full list from running:

```sh
find build/fr | grep '\.html$' | xargs grep -E --color=always '(\] +\()|\*\*' | aha
```

https://pste.eu/p/RuUu.html

Here's the English for comparison: https://pste.eu/p/2hxJ.html

timgdavies:
Thanks @bjwebb. I've run through on Transifex and I think I've fixed all of these. Can you re-check tomorrow.

Bjwebb:
There were a couple missed, but I've fixed those.

PR #562, opened 2017-09-05
Update code for Zambian Kwacha

### LINKED ISSUE #554: currency: Incorrect code for Zambian currency
We list ZMK as the code for Zambian currency in http://standard.open-contracting.org/latest/en/schema/codelists/ , however the code listed under ISO4217 is ZMW: https://www.currency-iso.org/en/home/tables/table-a1.html

duncandewhurst:
to reflect ISO4217:2015 - this code was updated effective 2013

closes #554


jpmckinney:
@duncandewhurst Can you add a note to the changelog?

Bjwebb:
There's other changes that are missing e.g. https://www.currency-iso.org/dam/isocy/downloads/dl_currency_iso_amendment_161.pdf

Also, shouldn't our codelist contain all historic values, to support historic data?

jpmckinney:
Can we have a script to generate the codelist?

For historic values, would we list them as deprecated?

jpmckinney:
Also, don't we need to update `release-schema.json`?

Bjwebb:
I've worked on a script to generate the codelist. There's a new pull request for this https://github.com/open-contracting/standard/pull/607, so closing this PR as superseded by that one.

PR #565, opened 2017-09-08
Add note on required fields in schema viewer

duncandewhurst:
Required fields are shown in bold in the schema viewer but we don't explain that anywhere


Bjwebb (approved):
I'm guessing this was mostly assigned to me due to the test fail, which I've fixed (was due to changes to the community extensions, that the tests weren't handling properly).

As for the content of this pull request, looks good to me.

PR #567, opened 2017-09-18
Typographic updates to contributors agreement

timgdavies:
The contributors agreement was missing a cross reference in §3 to §1, and a signature panel to make clear the information that should be submitted as part of signing.

This PR adds this missing information to 1.1-dev.

@Bjwebb If we have a good way to applying this change to other branches as well, we should do so.


Bjwebb (approved):

Bjwebb:
I've added the change to 1.0-dev.
I'll merge the changes onto 1.1 and 1.0 tomorrow.
I think that's all the branches we most care about.

PR #597, opened 2017-10-12
Updating directive to use plural awards and contracts. Closes #596

### LINKED ISSUE #596: Bug: missing awards/contracts in release reference
Reported by @juanpane, in the table at http://standard.open-contracting.org/latest/en/schema/reference/#release 'awards' and 'contracts' are not displaying. 

This appears to be because of a typo in the directive:

```eval_rst

.. jsonschema:: ../../../../build/current_lang/release-schema.json
    :include: 
    :collapse: planning,tender,award,contract,parties,buyer,relatedProcess

```

instead of the plural

```eval_rst

.. jsonschema:: ../../../../build/current_lang/release-schema.json
    :include: 
    :collapse: planning,tender,**awards**,**contracts**,parties,buyer,relatedProcess

```

I'll work on a PR for this.

timgdavies:
#596

@Bjwebb I don't think this has any wider implications for translation or other processes, but shout if you think it might.

Once this build has completed and I've tested, this could be merged into 1.1-dev, and deployed when we have translations for the other recent changes complete.

### DIFF
--- a/standard/docs/en/schema/reference.md
+++ b/standard/docs/en/schema/reference.md
@@ -45,7 +45,7 @@ All new information about a contracting process is described within a release.
 
 .. jsonschema:: ../../../../build/current_lang/release-schema.json
     :include: 
-    :collapse: planning,tender,award,contract,parties,buyer,relatedProcess
+    :collapse: planning,tender,awards,contracts,parties,buyer,relatedProcess
 
 ```
 


Bjwebb:
>  I don't think this has any wider implications for translation or other processes

That's correct. This table is built off the relevant schema, which is translated separately.

Issue #566, opened 2017-09-16
$schema should point to meta-schema incorporating our changes to JSON Schema


jpmckinney:
Originally reported via email by @devgateway, quoted below:

The `$schema` section should define the meta-schema of the current schema. This is used by validation tools that comply with the standard, to correctly identify entities that are part of the schema. Further read on the standard page:

http://json-schema.org/latest/json-schema-core.html#rfc.section.7

and another third party page linked from json-schema.org, they say it is a good read:

https://spacetelescope.github.io/understanding-json-schema/reference/schema.html

The issue is OCDS defines draft-04 as the reference meta-schema: `"$schema": "http://json-schema.org/draft-04/schema#"`

Because extra properties are defined in the OCDS schema files, the line above should read something like this:

`"$schema": "http://standard.open-contracting.org/schema/1__1/meta-schema.json"`

which should be an extension of draft-04, adding the `deprecated` keyword, and others that were added on top of draft-04.

This meta-schema extension should allow:

* validating the schema itself against the meta-schema (which is self-validating). For example if someone introduces now a typo in the definition of deprecated for a new deprecated field, say `deprecatedVersoin` instead of `deprecatedVersion`, this will be spotted easily.

* make validation tools know what rules to apply for validation, right now they correctly ignore all these extra keywords because `$schema` says it only has entities defined from draft-04 (which in the case of OCDS release files is false).

So for me, because I use one of these validating libraries in jOCDS, I need to hack the OCDS schema and either remove `$schema` altogether (which is a problem in itself), or write my own `$schema` and define your extra keywords there...

jpmckinney:
It may be possible to simply re-use this (reformatting that info as JSON): https://github.com/open-contracting/standard-maintenance-scripts/blob/master/tests/test_json.py#L42-L57

kindly:
@jpmckinney thanks for the link that was helpful.

In pull request #611 in the schema directory there is an adaptation of the draft-04 schema which is more strict (uses additionalProperties: false) and covers our use.  It is also configured to run in the tests so travis will pick up issues.  There is still some work todo on adding title and description tags to the fields we have added which @timgdavies said he would look at. Also, $ref and format properties had to be added to make the additionalProperties: false work.

I nonetheless have concerns about using this meta-schema in the the $schema tag.
[...]

PR #611, opened 2017-11-03
[#566] Add meta-schema.json and test using it.

kindly:
related #566


standard/schema/meta-schema.json
```diff
@@ -0,0 +1,190 @@
...
+                    "minItems": 1,
+                    "uniqueItems": true
+                }
+            ]
+        },
+        "allOf": { "$ref": "#/definitions/schemaArray" },
+        "anyOf": { "$ref": "#/definitions/schemaArray" },
+        "oneOf": { "$ref": "#/definitions/schemaArray" },
+        "not": { "$ref": "#" },
+        "$ref": {
```
jpmckinney: For `$ref`, draft-06 has:

```
        "$ref": {
            "type": "string",
            "format": "uri-reference"
        },
```

though we'd use `"format": "uri"` for draft-04.

standard/schema/meta-schema.json
```diff
@@ -0,0 +1,190 @@
...
+        },
+        "versionId": {
+            "type": "boolean",
+            "default": false
+        },
+        "wholeListMerge": {
+            "type": "boolean",
+            "default": false
+        }
+
```
jpmckinney: Extra line.

jpmckinney (commented):
Made a few comments. It might be good for easier change-tracking to first commit the draft-04 schema, and then commit our changes, so that we can easily see the divergence when reviewing history.

standard/schema/meta-schema.json
kindly: I think the issue here is that internal $ref i.e '#/definitions/Awards' would not pass "uri" validataion. Which I think is why they added the new format.

jpmckinney: Aha, makes sense.

standard/schema/tests/test_validate_schema.py
```diff
@@ -31,14 +31,15 @@ def object_pairs_hook(pairs):
 
 
 DRAFT_SCHEMA_PATH = path(__file__).parent / 'json-schema-draft-4.json'
```
jpmckinney: This constant's value needs to be updated because the file moved in another commit, but if we aren't using the constant anymore, we can just remove it.

PR #647, opened 2018-01-15
Update sphinxcontrib-jsonschema to warn if collapse doesn't exist

Bjwebb:
https://github.com/OpenDataServices/sphinxcontrib-jsonschema/issues/1


jpmckinney:
I figure we should fix the newly uncovered errors before merging. Can you do that @Bjwebb ?

PR #648, opened 2018-01-15
Fix functional tests

Bjwebb:


jpmckinney:
Seems `make_validation_schema.py` needs to be run?

jpmckinney (approved):

PR #649, opened 2018-01-15
Quiet build

jpmckinney:
Finding the errors and warnings from Sphinx, etc. becomes quite a challenge when unnecessary informative log messages crowd the output.


jpmckinney:
Ah, that output is much shorter: https://travis-ci.org/open-contracting/standard/builds/329094631

Failing until #648 is merged.

Bjwebb (approved):

PR #659, opened 2018-01-23
 Replace triple backticks with single backticks

jpmckinney:
I made replacements in both the source language and translated languages.

Fixes https://github.com/open-contracting/standard-development-handbook/issues/32

Includes #658


jpmckinney:
I ran a test ([as documented here](https://github.com/open-contracting/standard-development-handbook/issues/141)) with this branch, and it looks like we can safely `tx push -s -t -f --no-interactive` after merging. Have a look: https://www.transifex.com/open-contracting-partnership/open-contracting-standard-1-1-dev/dashboard/ The strings to translate are from extensions resources, new title/descriptions in core schema, and new descriptions for codes.

PR #661, opened 2018-01-30
Miscellaneous fixes

jpmckinney:
Includes #659


jpmckinney:
@timgdavies It's likely easier to pick specific commits to review, rather than look at all files changed.

jpmckinney:
The only normative change is the update to the currency codelist, which is mentioned in the changelog, so merging.

PR #662, opened 2018-01-31
Extension update script and including codelists

timgdavies:
@jpmckinney I fear I ended up undertaking some slightly parallel work to you on synchronising the extensions, though I'm not sure the approach in https://github.com/open-contracting/standard/blob/1.1.3-dev/standard/docs/en/extensions/get-readmes.py#L11 is correct - as I understood our current approach was to [read core extensions from a 1.1.3 branch of the extension registry](https://github.com/open-contracting/extension_registry/tree/v1.1.3) - rather than to just auto replace in-line with 1.1.3. 

I've also in my version of the get-readme.py (renamed to represent full goal) at https://github.com/open-contracting/standard/blob/1.1.3-dev-extension-scripts/standard/docs/en/extensions/fetch_updated_extension_docs.py I've included fetching all the extension codelists, as I anticipate that readme files *should* really display these, and in future iterations we might want to better automate creation of core extension documentation pages to include all their schema additions and codelists. 

This PR just to show the approach I've taken to support discussion. May need some more work before merging.


jpmckinney:
I changed the base to 1.1.3-dev as we won't be merging to 1.1-dev directly.

jpmckinney:
My extra line in get-readmes.py is just to be able to quickly test things without creating an extension_registry branch.

I don't think we want to distribute the core extensions' codelists in the standard unless required for display.

I committed "Correct typo in suppliers definition." to 1.1.3-dev directly, and I'll copy the changes to `fetch_updated_extension_docs.py ` to a new branch, to avoid unnecessarily fixing conflicts here.

Issue #660, opened 2018-01-29
Bug in versioned-release-validation-schema.json


timgdavies:
.@kindly has identified a bug in [versioned-release-validation-schema.json](https://github.com/open-contracting/standard/blob/1.1-dev/standard/schema/versioned-release-validation-schema.json#L10) which is omitting the object type from each versioned property. 

As JSON Schema validators only check for properties of objects, this means the fields of each versioned object have not been correctly checked. 

I.e., in the example below, the schema has correctly checked that `amount` is a property within the schema, but it has not validated the `releaseDate`, `releaseID`, `releaseTag` and `value` properties, such that an implementation that had, for example, misspelled these properties, or otherwise omitted them, would incorrectly be evaluated as valid. 

```json
{
  "amount": [
    {
      "releaseDate": "2016-01-01T09:30:00Z",
      "releaseID": "ocds-213czf-000-00002-01-tender",
      "releaseTag": [
        "tender"
      ],
      "value": 1000
    }
  ]
}
```

This is a result of a missing line in the [make_validation_schema.py](https://github.com/open-contracting/standard/blob/1.1-dev/standard/schema/utils/make_validation_schema.py#L14) script - which has been the case since the first version.

version-release-validation-schema.json is used within record-package-schema.json for publishers who provide `versionedRelease`.  

The impact of correcting this bug would be that record packages containing `versionedRelease` would be more strictly validated than has been the case to date. 

We are not aware of any currently published `versionedRelease` data. 

Our options here are to:

* **Fix versioned-release-validation-schema.json generation as part of a bug fix release (1.1.3)**. On a strict interpretation, this would involve a backwards incompatible change, as it would mean data that was incorrect against the documentation, but valid against 1.1.2 schema, could now become invalid against the schema also. However, it is still clearly a bug fix. 

* Fix versioned-release-validation-schema.json generation as part of an integer release (i.e. wait until 2.0). In this instance, we should introduce a warning to the OCDS validator for this edge case in the meantime. 

I'm leaning towards fixing in 1.1.3 - given *in practice* we don't think this will have any backwards compatibility issues for existing publishers, and any publisher who followed the documentation and guidance should not have data that would be invalid against the corrected schema. 

@jpmckinney thoughts?

jpmckinney:
We've had some schema bugs that are non-obvious to implementers (e.g. not requiring `id` fields on objects in arrays), and which we therefore postponed to 2.0.

[...]

PR #663, opened 2018-01-31
[#660] Add type: "object" to each version block

kindly:
This makes each version block validate.


jpmckinney:
When working on a release, start from the appropriate branch (1.1.3-dev) instead of 1.1-dev, as documented here: http://ocds-standard-development-handbook.readthedocs.io/en/latest/standard/technical/repository/#branches-and-tags

I merged 1.1.3-dev, which contains the fix for chrome, and changed the base for the PR.

jpmckinney:
No harm in adding "type": "object", but I'm still not sure that this actually changes the behavior of validators (see comment in #660).

PR #664, opened 2018-01-31
Updates to scripts

jpmckinney:
Work in progress. Follow-up to #662


jpmckinney:
Should change requirements.txt after OpenDataServices/sphinxcontrib-jsonschema#6 is merged.

PR #666, opened 2018-02-01
Better document the build

jpmckinney:
Also, as far as I can tell, the building of .pot files is unnecessary to building the documentation (though it's necessary for collecting the source files that get pushed to Transifex). Is that correct?

I find build_docs.sh much easier to read now; the interdependency between commands is clear, and the script no longer requires jumping in and out of the `standard` directory.


jpmckinney:
Quite happy with this now. I can update the documentation in the handbook to use the Makefile, but now you can:

```
make clean # remove build directory and .mo files
make extract # generates .pot files
make push # generates .pot files then pushes to Transifex
make pull.es_MX # pull a specific translation
make pull # pull supported translations (es,fr)
make source # build the source language's docs
make es # build a specific translation's docs (as well as the source language's)
make # build supported translations' docs (as well as the source language's)
```

translate_codelists.py now uses the build directory and doesn't overwrite the same files in different languages (which could cause issues if commands were run out-of-order), instead writing the translated codelists to different language directories.

Note that you can use `make -n` to see the commands that would be run, without running them.

Bjwebb (approved):
From a quick look through, this all seems to be good.

jpmckinney:
* [x] Merge https://github.com/open-contracting/ocds_bid_extension/pull/30 after this.

Issue #639, opened 2017-12-06
Buyer name required in OrganizationReference


timgdavies:
In OCDS 1.1 the OrganizationReference block has two non-deprecated properties:

* `id`
* `name`

It currently specifies that `name` is a required property, which causes data without a `name` to fail validation. 

This appears to be a bug as:

* (a) We generally require `id` values, but not strings;
* (b) There are legitimate cases where no `name` can be provided;

Following investigation to confirm that this is a bug, and check if there was any rationale for the original requirement being introduced that we've overlooked here, we should look at addressing this in a patch release. 

Removing the required property from `name` is anticipated to have minimal impact on existing applications.

timgdavies:
Tagging @kindly who worked on required properties for OCDS 1.1, and @bjwebb who might have a quicker way than I do to work out how to 'git blame' and work out when we (I?) introduced this.

jpmckinney:
It was done in #368.

On the other hand, it's not clear to me that requiring `id` was backwards-incompatible, as `id` was introduced to `Organization` in 1.1 and doesn't exist in 1.0 (whereas `name` already existed in 1.0). The evidence in #484 for the backwards-incompatibility of requiring `id` seems to be spurious?

jpmckinney:
Wait – I had the logic backwards. Since `id` didn't exist in 1.0, it can't be required in 1.1! So that's fine.

PR #667, opened 2018-02-02
1.1.3 Release, including fix to tag codelist in record package

timgdavies:
A bug fix release for the OCDS Schema, Documentation and Core Extensions has been drafted and is available to preview at http://standard.open-contracting.org/1.1.3-dev/en/

This release adds a number of missing definitions to schema elements and code lists, as well as correcting the use of “null” as a value within the schema. 

As a bug-fix release, this does not introduce any additional fields or objects, nor should it affect the validation of existing data, with the exception of data that was incorrectly failing validation due to issue #639 or when using null values to unset fields. 

The [changelog is available here](http://standard.open-contracting.org/1.1.3-dev/en/schema/changelog/). 

Users are invited to feedback on the changes before 15th February 2018. Unless substantive objections are raised, this will become the live version of OCDS 1.1 shortly afterwards. 

In-line with the new translation policy, schema translations into Spanish and French will be made available at the same time the updated version goes live. 

**Notes**

Through improved testing, we have also identified an issue with `id` values for objects within arrays that publishers should be aware of: 

At a number of points in the main schema and in extensions we have found objects contained within an array that do not have a required `id` field. 

Whenever OCDS data contains an array of objects, the `id` field is required to facilitate the releases-records merging model, and to allow objects in the array to be ‘unset’ by later releases. 

Making the `id` fields required on all these objects may cause data that was valid against 1.1.0 to be invalid against 1.1.3, and so we have not introduced a required property in this release. However, a warning will be added to the validator for cases where objects within arrays are missing an `id` property.


jpmckinney:
@timgdavies Can we merge this onto 1.1-dev and then merge a later PR with the updated *.po files? Merging should cause several issues to close, which will make it easier to see the issues that are truly open.

jpmckinney:
Per [governance](http://standard.open-contracting.org/latest/en/support/governance/#versions) and [handbook](http://ocds-standard-development-handbook.readthedocs.io/en/latest/standard/technical/repository/#branches-and-tags) I see no issue with merging onto 1.1-dev, so merging.

PR #668, opened 2018-02-02
Add OCP favicon to standard docs

edugomez:


jpmckinney:
I guess we can squeeze one more commit into 1.1.3 :)

edugomez:
@jpmckinney Thanks. A quick comment on contributions to the standard repo. The handbook does not specify very clearly the targets for PRs. I assume that currently bug fixes should be submitted against 1.1 or perhaps 1.1-dev and then from there merge into 1.1. I decided to send this PR to 1.1.3-dev as it is where more of the action is and adding a favicon doesn't feel like a bug fix. Anyway, my point is that workflow is a convention after all and should be more clearly specified.

jpmckinney:
The documentation is here: http://ocds-standard-development-handbook.readthedocs.io/en/latest/standard/technical/repository/ but any improvements or clarifications to that documentation are welcome.

PR #674, opened 2018-03-09
Expand the meaning of versions in OCDS

jpmckinney:


jpmckinney:
Feedback:

> When describing semantic versioning, please include a description of what consuming implementations that have been implemented to a particular version (eg 1.1.1) are expected to do with previous and following major, minor and path versions. Don't just describe it from a publisher perspective.

jpmckinney:
Noting from today's call with @duncandewhurst and @timgdavies that we can add some text here about how core extensions are versioned. We can also explain that core extension version numbers may not match OCDS version numbers.

timgdavies:
I've added a commit to update the governance page with a section on the versioning of core extensions.

I also added a minor change to the text on versioning, to avoid any misinterpretation of what we mean about validation of data (i.e. validation against the schema, rather than any other sort of review or quality assurance of data that might have taken place).

I believe that the explanation that core extension version numbers may not match OCDS version numbers should be shown as part of core extensions page in the updated extension registry. I will work with @kindly to make sure this comes up in the interface.

standard/docs/en/support/governance.md
```diff
@@ -200,5 +200,15 @@ Anyone who is a current or potential publisher or user of the standard can be co
 
 "The principle of consensus has its origins in the desire to achieve the general acceptance and application of a Standard within its intended sphere of influence. This entails trying to ensure that the interests of all those likely to be affected by it are taken into account, and that individual concerns are carefully and fairly balanced against the wider public interest." [BSI, 2012](http://www.bsigroup.com/Documents/about-bsi/NSB/BSI-pocket-guide-to-standards-development-UK-EN.pdf) 
 
-   
+## Versioning of extensions
+
+Extensions are governed and versioned independently of the standard, with the exception of core extensions, where:
+
+* A MAJOR or MINOR release of a core extension will only take place as part of the governance process for a MAJOR or MINOR release of OCDS.
```
jpmckinney: I think it's more future-proof to use the present tense, instead of the future tense, in these paragraphs. We can say, "starting with the version after OCDS 1.1.3" if desired.

standard/docs/en/support/governance.md
```diff
@@ -200,5 +200,15 @@ Anyone who is a current or potential publisher or user of the standard can be co
...
-   
+## Versioning of extensions
+
+Extensions are governed and versioned independently of the standard, with the exception of core extensions, where:
+
+* A MAJOR or MINOR release of a core extension will only take place as part of the governance process for a MAJOR or MINOR release of OCDS. 
+
+* Each version of the standard will explicitly declare the version number of each core extension for that version of the standard. 
+
+Changes to core extensions between standard versions may be staged with -alpha, or -beta suffixes to indicate these are under development or proposed for adoption in the next version of OCDS, but that their adoption has not yet been formally agreed.
```
jpmckinney: I suggest "with labels (like `-alpha` or `-beta`)" instead of "with -alpha, or -beta suffixes", to allow for `-rc.1` etc.

standard/docs/en/support/governance.md
```diff
@@ -200,5 +200,15 @@ Anyone who is a current or potential publisher or user of the standard can be co
...
+
+Extensions are governed and versioned independently of the standard, with the exception of core extensions, where:
+
+* A MAJOR or MINOR release of a core extension will only take place as part of the governance process for a MAJOR or MINOR release of OCDS. 
+
+* Each version of the standard will explicitly declare the version number of each core extension for that version of the standard. 
+
+Changes to core extensions between standard versions may be staged with -alpha, or -beta suffixes to indicate these are under development or proposed for adoption in the next version of OCDS, but that their adoption has not yet been formally agreed. 
+
+
```
jpmckinney: Extra newlines.

duncandewhurst:
@jpmckinney I've just stumbled across this, do you still need me to review it, or is the action to address your review comments?

jpmckinney:
The comments need to be addressed, and it's been long enough that we should then re-review.

standard/docs/en/support/governance.md
duncandewhurst: Done in https://github.com/open-contracting/standard/pull/674/commits/ef99fb58b6213da3d9e35b4b87a54b2a78697990

standard/docs/en/support/governance.md
duncandewhurst: Done in https://github.com/open-contracting/standard/pull/674/commits/ef99fb58b6213da3d9e35b4b87a54b2a78697990

standard/docs/en/support/governance.md
duncandewhurst: Done in https://github.com/open-contracting/standard/pull/674/commits/ef99fb58b6213da3d9e35b4b87a54b2a78697990

docs/governance/index.md
```diff
@@ -58,6 +58,20 @@ This documentation website is composed of normative content (the prescriptive pa
 
 ### Versions
 
+We use **version numbers** to indicate the nature of changes made to normative content, and to fit the governance process to each change.
+
+* **A MAJOR version (e.g. 1.0.0 -> 2.0.0)** makes backwards-incompatible changes. Systems written to input or output data with a previous major version might need substantial updates. Data that is valid against the 1.X schema might not be valid against the 2.X schema, for example.
+
+* **A MINOR version (e.g. 1.0.0 -> 1.1.0)** makes backward-compatible changes. Systems written to input or output data with a previous minor version should be reviewed to assess whether they can implement any concepts added in the new minor version. Data that is valid against the 1.0 schema will be valid against the 1.1 schema, for example. Where relevant, changes should apply a principle of progressive enhancement, such that updated systems can take advantage of the added concepts, while non-updated systems experience no substantive changes in behavior.
```
duncandewhurst: Use present tense.

```suggestion
* **A MINOR version (e.g. 1.0.0 -> 1.1.0)** makes backward-compatible changes. Systems written to input or output data with a previous minor version should be reviewed to assess whether they can implement any concepts added in the new minor version. Data that is valid against the 1.0 schema is also valid against the 1.1 schema, for example. Where relevant, changes should apply a principle of progressive enhancement, such that updated systems can take advantage of the added concepts, while non-updated systems experience no substantive changes in behavior. 
```

docs/governance/index.md
```diff
@@ -58,6 +58,20 @@ This documentation website is composed of normative content (the prescriptive pa
...
+
+* **A MINOR version (e.g. 1.0.0 -> 1.1.0)** makes backward-compatible changes. Systems written to input or output data with a previous minor version should be reviewed to assess whether they can implement any concepts added in the new minor version. Data that is valid against the 1.0 schema will be valid against the 1.1 schema, for example. Where relevant, changes should apply a principle of progressive enhancement, such that updated systems can take advantage of the added concepts, while non-updated systems experience no substantive changes in behavior. 
+
+* **A PATCH version (e.g. 1.1.0 -> 1.1.1)** updates open codelists or external closed codelists, or makes backwards-compatible bug fixes or improvements to normative content. Users are encouraged to upgrade to the latest patch version. Unless they were affected by a fixed bug, they should not see any substantive changes after the upgrade. Data that is valid against the 1.1.0 schema will be valid against the 1.1.1 schema, for example.
```
duncandewhurst:
```suggestion
* **A PATCH version (e.g. 1.1.0 -> 1.1.1)** updates open codelists or external closed codelists, or makes backwards-compatible bug fixes or improvements to normative content. Users are encouraged to upgrade to the latest patch version. Unless they were affected by a fixed bug, they should not see any substantive changes after the upgrade. Data that is valid against the 1.1.0 schema is also valid against the 1.1.1 schema, for example.
```

duncandewhurst:
@jpmckinney I've addressed the review comments, re-reviewed the content added in this PR and made a couple of suggestions.

jpmckinney:
Thanks @duncandewhurst, I've committed your suggestions. Please approve so that it can be merged.

jpmckinney:
Ah, nevermind – the blocker was a CI error. As such I approve your changes, and I take your comment as approval as well.

PR #688, opened 2018-04-17
Update versioning updates for RFC2119 alignment

timgdavies:
I've made minor suggested edits to the suggested expanded versioning text, taking into account the work we're doing tidying up our use of normative language.


jpmckinney:
Failures seem due to chrome driver again.

PR #692, opened 2018-04-19
Fix typos reported by translators

jpmckinney:
This PR fixes a number of issues in the source language (English) reported by translators.

* [x] After merging, I will force push the translations to Transifex, which should remain 100% translated, except for the change to serialization.md


duncandewhurst (commented):
should the change to serialization.md also be updated in the corresponding .po files?

jpmckinney:
Aha - that's one that will need to be translated (but that's not an issue). The others I was able to translate easily as the changes were small.

Anything else before merging?

duncandewhurst (approved):
nope, good to go

Issue #677, opened 2018-03-22
Minor issues in package schemas


jpmckinney:
Missing title/description:

* [x] release-package-schema.json is missing /properties/releases/description
* [x] record-package-schema.json is missing /properties/publisher/title
* [x] record-package-schema.json is missing /definitions/record/description
* [x] record-package-schema.json is missing /definitions/record/properties/releases/oneOf/0/items/properties/url/title

Need to check whether it makes sense for these to be nullable, etc.:

* release-package-schema.json has optional but non-nullable array at /properties/extensions
* record-package-schema.json has optional but non-nullable array at /properties/extensions
* record-package-schema.json has optional but non-nullable array at /properties/packages
* record-package-schema.json has required but nullable ['string', 'null'] at /definitions/record/properties/releases/oneOf/0/items/properties/url
* record-package-schema.json has optional but non-nullable array at /definitions/record/properties/releases/oneOf/0/items/properties/tag

duncandewhurst:
Whilst working on this I noted a typo in the record reference documentation which I fixed in https://github.com/open-contracting/standard/pull/713

duncandewhurst:
I don't think that fields in the release and record packages should be nullable, since merging only applies to the fields in the release itself, not the package metadata. Maybe @kindly can confirm?

jpmckinney:
@duncandewhurst Ah, yes, package schema follow different rules. I do notice that most of the fields are nullable or non-nullable based on whether they are required or not, with only the above exceptions. 

While on the topic, It's not clear to me that `null` should be a valid value for any of these, though it's not a significant issue (and we can't go back and change that now anyway).

I'll remove the nullability check from standard-maintenance-scripts for package schema – though if we want to follow rules for nullability going forward, I can restore that functionality.

kindly:
@duncandewhurst @jpmckinney 
Yes I agree the release/record package fields should probably not be nullable as the merging does not happen to them.

jpmckinney:
Resolved in #712.

PR #712, opened 2018-05-21
Add missing properties

duncandewhurst:
Addresses https://github.com/open-contracting/standard/issues/677

@timgdavies please could you check you're happy with the description for `record`


jpmckinney:
Are records only published when a process is complete? If not, then a record won't provide "all the OCDS releases relating to a single contracting process" until that process is completed. We can perhaps add a clause to account for that temporal issue.

Also, are the MUST and SHOULD statements already part of the documentation, or are those new normative statements? I know they reflect our intention, but I just want to check that they have precedence.

duncandewhurst:
> Are records only published when a process is complete? If not, then a record won't provide "all the OCDS releases relating to a single contracting process" until that process is completed. We can perhaps add a clause to account for that temporal issue.

No we expect a record to be published when the first release about a contracting process is published and to be updated throughout the process. I updated the description to read "all the existing OCDS releases"

> Also, are the MUST and SHOULD statements already part of the documentation, or are those new normative statements? I know they reflect our intention, but I just want to check that they have precedence.

Yes, these are on the [record reference](http://standard.open-contracting.org/latest/en/schema/records_reference/#versioned-release) page of the docs.

jpmckinney:
PR looks good to me, then!

timgdavies (approved):
The definition works well.

PR #713, opened 2018-05-21
fix reference to compiledRecord

duncandewhurst:
Fixing typo


jpmckinney:
Ignoring whitespace: https://github.com/open-contracting/standard/pull/713/files?w=1

PR #719, opened 2018-05-25
fix extensionlist in party details docs

duncandewhurst:
Addresses https://github.com/open-contracting/standard/commit/c601cb36da251e7cc76946d46aeb4acdd75c4389#commitcomment-29108248

### DIFF
--- a/standard/docs/en/extensions/party_details.md
+++ b/standard/docs/en/extensions/party_details.md
@@ -7,6 +7,6 @@ Each of the entries in the [parties](../../../../schema/reference/#parties) arra
 A list of available extensions (tagged as 'partyDetail') can be found below:
 
 ```eval_rst
-.. extensionlist:: The following extensions are available for parties
-   :list: parties
+.. extensionlist:: The following party detail extensions are available:
+   :list: partyDetail
 ```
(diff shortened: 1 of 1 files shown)


standard/docs/en/extensions/party_details.md
```diff
@@ -7,6 +7,6 @@ Each of the entries in the [parties](../../../../schema/reference/#parties) arra
 A list of available extensions (tagged as 'partyDetail') can be found below:
 
 ```eval_rst
-.. extensionlist:: The following extensions are available for parties
-   :list: parties
+.. extensionlist:: The following party detail extensions are available:
```
jpmckinney: None of the other calls to extensionlist use a colon. Given that this text appears in the title section of an admonition, I think it looks okay without the colon. If we want a colon, we should be consistent.

We use the construction "The following extensions are available for …" in most other cases, but I'm fine with the present construction. Just FYI if you see an opportunity to align.

Examples of `extensionlist`'s: http://standard.open-contracting.org/latest/en/schema/reference/

duncandewhurst:
@jpmckinney thanks, that makes sense. I added another commit with those changes

PR #724, opened 2018-06-12
Use new registry format

jpmckinney:
In my local testing, all three Sphinx directives from sphinxcontrib-opencontracting continue to work after these changes, as does extensions.js, which adds community extensions to some tables created by those directives. `fetch_core_extensions.py` also produces the same results as before (with the exception of a small change to the Markdown template).

Once this PR is built, they can be tested at:

* http://standard.open-contracting.org/registry/en/schema/reference/#tender
* http://standard.open-contracting.org/registry/en/extensions/#community-extensions
* http://standard.open-contracting.org/registry/en/extensions/community/
* http://standard.open-contracting.org/registry/en/extensions/bids/#bid-statistics

Note that the documentation URLs for core extensions have changed, which should be corrected by closing https://github.com/open-contracting/ocds-extensions/issues/65

### DIFF
--- a/standard/docs/en/_static/extensions.js
+++ b/standard/docs/en/_static/extensions.js
@@ -3,13 +3,17 @@ $(function () {
   var isUsingExtensionsPage = window.location.pathname.indexOf('/extensions/') >= 0;
   var isCommunityPage = window.location.pathname.indexOf('/extensions/community/') >= 0;
   var isSchemaReferencePage = window.location.pathname.indexOf('/schema/reference/') >= 0;
+  var blankListEl = '<dl><dt><a class="reference external" href=""></a></dt><dd></dd></dl>';
 
   if (!isUsingExtensionsPage && !isCommunityPage && !isSchemaReferencePage) {
     return
   }
 
-  // Core extensions only
+  // The "Extension" column values must be transformed to links, e.g.
+  // "Bid statistics and details::https://github.com/open-contracting/ocds_bid_extension"
+  // TODO: Why do we transform the links in JavaScript and not in Sphinx?
   if (isUsingExtensionsPage) {
+    // Get the table's data rows, which will only be core extensions.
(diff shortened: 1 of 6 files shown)


jpmckinney:
The URLs above work as expected. Requesting review. Something for @Bjwebb or @kindly ?

jpmckinney:
I'll merge this PR next Monday if no word, as it causes no visible changes to the docs besides the above noted documentationURL issue, for which there are other linked PRs.

standard/docs/en/_static/extensions.js
```diff
@@ -20,17 +24,16 @@ $(function () {
...
     // append an empty list for community part of the section
     $('.extension_list p.last').after('<br><dl class="docutils hide community-list"></dl>');
     $('.extension_list p.last').wrap('<small></small>');
   }
 
+  // Get the community extensions to add to the documentation.
   $.ajax({
-    "url": "http://standard.open-contracting.org/extension_registry/master/extensions.js", 
-    "jsonpCallback": "extensions_callback",
+    "url": "https://raw.githubusercontent.com/open-contracting/extension_registry/master/build/extensions.json",
```
kindly, line 34: We were trying to make sure we did not have a reliance on github.

kindly: We also have to commit a compiled version of the extensions to the extension_registry which was also something that we tried to avoid.

kindly (approved):
This seems ok as a stopgap till the extension explorer website is made but I am not sure if it is needed at all as it will only effect 1.1-dev till it's backported to latest/1.1.

standard/docs/en/_static/extensions.js
jpmckinney: Indeed, this is a stopgap.

jpmckinney:
It may indeed not be needed, but it's good implementation experience for the new registry format, and until the explorer is ready, such changes are appropriate.

PR #752, opened 2018-09-20
Fix old or broken links

jpmckinney:
e.g. URLs containing `ocds.open-contracting.org`, `1.1.1-dev`, etc.

Translations have also been updated to minimize impact.

Qualifies as non-normative change based on [exceptions](https://docs.google.com/document/d/1xjlAneqgewZvHh6_hwuQ98hbjxRcA2IUqOTJiNGcOf8/edit#heading=h.3js2okx1x4fu) in policy.


jpmckinney:
To ignore whitespace changes: https://github.com/open-contracting/standard/pull/752/files?w=1

jpmckinney:
Two weeks since review requested for non-normative changes. Merging.

PR #753, opened 2018-09-20
Remove update_schema_ids.py (will use find command instead)

jpmckinney:
See https://github.com/open-contracting/standard-development-handbook/pull/199

This is a non-normative change, as it changes [non-normative content](https://docs.google.com/document/d/1xjlAneqgewZvHh6_hwuQ98hbjxRcA2IUqOTJiNGcOf8/edit).


jpmckinney:
Two weeks since review requested for non-normative changes. Merging.

PR #806, opened 2019-01-28
Add Italian translation of 1.1.3

jpmckinney:
FYI, @duncandewhurst, @groundrace

Given that 1.1.4 will involve more text changes and is bigger than originally anticipated, and given that 1.1-dev includes some normative changes, this PR builds on 1.1 to add the Italian translation.

I'll sort out the merge conflicts with 1.1-dev afterwards.

This branch should look the same as the old branch http://standard.open-contracting.org/1.1-italian/it/ which built on 1.1-dev (which at the time already included things that went beyond 1.1.3).

Once this branch builds successfully, we should follow [these steps](https://ocds-standard-development-handbook.readthedocs.io/en/latest/standard/technical/deployment/#copy-the-files-to-the-live-server) including updating `robots.txt`.

@robredpath Please give me SSH access to perform the necessary steps. Otherwise, who among @Bjwebb and @kindly (who, as far as I know, have been the only people to deploy the docs) would be available this week to do so?


jpmckinney:
It seems to be taking more than 10mins for the deploy, and so Travis is erroring. I can't tell but I suspect that re-indexing is taking too long. We might want to add some echo statements to the build script so that Travis receives output between steps (and restarts its 10min countdown).

I added docs for adding community extensions here: https://ocds-standard-development-handbook.readthedocs.io/en/latest/standard/translation/technical/#add-a-community-translation

jpmckinney:
Fixed the build by making the mirror command verbose. Preview here: http://standard.open-contracting.org/1.1-italian-bis/it/

Bjwebb:
I can take a look at this on Thursday.

This PR:
* Moves the deploy script into this repo
* Pins a commit of the standard theme

Previously, it was possible to update the deploy script or theme without a commit to this repo, and to just hit rebuild on Travis. After this change, a commit will be needed on each branch to be updated. (But, this change does make builds/deploys more exactly reproducible). Is this okay?

jpmckinney:
Thanks! Please let me know if there's anything you need.

After this build I'll pin the theme to the branch. The problem was that the theme had layout changed, with matching changes in the standard on 1.1-dev. However, since in this PR I'm just trying to make the minimal changes to build 1.1.3 with the Italian translation, I had to pin to the commit on which it was built.

For the deploy script, I don't have write access to the deploy repo (which makes sense, as it affects non-OCP repos), but it also means that OCP has less control over its deployment. I think it's fine to leave in this repo, as it changes very infrequently.

The only critical branches for deployment are `1.0` and `1.1`, so I don't think it's too much trouble to have to update both if we change the deploy script in a meaningful way. The present change was just to increase output. And as I don't expect any further releases of 1.0, we'll likely leave `1.0` as-is.

Bjwebb:
This is deployed now http://standard.open-contracting.org/latest/it/

jpmckinney:
Thanks, @Bjwebb !

PR #809, opened 2019-01-31
Fix Italian typo

jpmckinney:
If no more corrections within a week, we should merge and deploy.


jpmckinney:
@robredpath @Bjwebb Can I be given sufficient access to the production server, so that I can deploy small changes like from this PR?

PR #808, opened 2019-01-31
Merge 1.1 into 1.1-dev

jpmckinney:
This applies the changes needed to deploy OCDS 1.1.3 in Italian to 1.1-dev.

Essentially:

* Inline the deployment script
* Update the language dropdown
* Add a localization-note
* Add Italian translations

Some other commits just replicated changes that were already in 1.1-dev.

PR #810, opened 2019-02-01
Fix versioned release schema and merge rules

### LINKED ISSUE #770: tag should be omitWhenMerged
The behavior of OCDS Merge is to set the tag to `compiled` for compiledRelease, and to move the tag onto the versioned object for versionedRelease. In other words, the original tag field is omitted.

Assigning to @kindly to confirm. The docs don't describe the first behavior, which we can fix as part of this issue.

### LINKED ISSUE #771: Add missing documentation on merging
I don't see where in the documentation the following OCDS Merge behaviors are prescribed:

* [x] The `id` of the compiledRelease is the `id` of the latest release (#810 #834 set `{ocid}-{date}`)
* [x] The `date` of the compiledRelease is the `date` of the latest release #810 #834 
* [x] The `tag` of the compiledRelease is `["compiled"]` #770 #810
* [x] The `ocid` of the versionedRelease is not versioned (only specified in schema) #810
* [x] The `tag` of the versionedRelease is omitted #770 #810
* [x] If a value in a release is the same as in its preceding release, don't add a versioned object #810

@kindly Did I miss these somewhere?

* [x] As part of this we can clarify precedence, i.e. that `omitWhenMerged` precedes all other rules.

### LINKED ISSUE #798: Versioned release schema is inconsistent with merge rules
## Issue

The [merge rules](http://standard.open-contracting.org/latest/en/schema/merging/#versioned-data) state that all fields are versioned "with the exception of the `id` property of objects within an array". The OrganizationReference definition appears in both the context of an object in an array (e.g. `suppliers`) and in the context of a plain object (e.g. `buyer`). According to the rules, in one case its `id` property should be versioned (`buyer`, etc.), in the other it shouldn't (`suppliers`, etc.).

However, the versioned release schema uses the same definition for both contexts, which states that `id` isn't versioned. But, we *do* want the `id` in `buyer` and other single organization references to be versioned.

Classification and Identifier also appear in the context of both an object in an array (`additionalClassifications`, `additionalIdentifiers`) and in the context of a plain object (`classification`, `identifier`). However, those arrays are whole list merge, and objects in that context already have no versioned fields. That said, the versioned release schema has a bug in that it states that their `id` isn't versioned. Compared to the above, this is an easy fix.

## Discussion

**Note: Ignore this discussion, as a simpler solution has been found.**

I believe that the merge rules are correct and that the versioned release schema is incorrect. The three broad solutions are:

1. Change the versioned release schema to merge these objects as a whole, if not in an array
1. Change the versioned release schema to make `id` versioned if not in an array of objects
1. Remove the versioned release schema from the standard

### Merge the objects as a whole

We either:

1. Add a new `wholeObjectMerge` property, to indicate that the `buyer`, etc. object should be merged as a whole. (For non-`$ref` objects, we can simply add a `type` of "null" to achieve the same behavior. However, we can't do that for Classification, Identifier, and OrganizationReference, because these definitions are references from arrays, and we don't want to allow `null` in arrays.)
1. Use different definitions when referring to an object in array-context than when referring to it in non-array-context. This allows us to use the strategy of adding a `type` of "null" mentioned before, which has the advantage of not adding a new merging rule. There's some discussion in #606 about what sort of tool support we'll need to overcome the disadvantages of a bigger, repetitive schema.

### Make `id` versioned if not in an array of objects

As above, we either (1) add a new merge property to indicate that this object's `id` field should be versioned, or (2) make a copy of the definition and set `"versionId": true` for its `id` field.

### Remove the versioned release schema from the standard

This has been mentioned before, but over time I've appreciated the functionality that the versioned release provides. It makes it very easy to see changes over time, which is a common use case. That said, hardly anyone has implemented it, and it is always possible to calculate from individual releases. We also, evidently, have trouble defining its schema, documenting it, and implementing it.

## Proposal

Within the first two approaches, I prefer the second sub-option as it helps solve #606 as well, which will become an increasingly severe issue as the number of extensions increases. It's likely more work than the first option, however. The first option can be pursued in a patch release. The second option *might* be pursued in a patch release – it depends if it affects any extensions.

I'm open to the third approach (removing the versioned release schema), but it can't be pursued without a minor version (1.2), as it's a significant change.

cc @mpostelnicu, @duncandewhurst, @kindly, @LindseyAM

jpmckinney:
Notable changes:

* Merging documentation is rewritten
* `tag` is omitWhenMerged #770 
* `Budget.id`, `Identifier.id` and `Classification.id` are versioned #798 
* `buyer.id`, `Tender.procuringEntity.id`, `Transaction.payer.id`, and `Transaction.payee.id` are versioned, using a special `OrganizationReferenceVersionedId` definition #798 
* Add versionId to Identifier and Classification
* All unused `*Unversioned` definitions are removed
* Add a dereferenced release schema

closes #770, closes #798, closes #771


jpmckinney:
Ready for review - please see linked issues for context.

jpmckinney:
@mpostelnicu The versioned release schema in this pull request should now be corrected.

standard/docs/en/schema/changelog.md
```diff
@@ -8,13 +8,15 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Advisories
 
 * [#780](https://github.com/open-contracting/standard/pull/780) All extensions have been moved to the [open-contracting-extensions](https://github.com/open-contracting-extensions) organization. No change is required, but we recommend that publishers update the URLs of extensions in release packages and record packages.
-* [#744](https://github.com/open-contracting/standard/pull/744) OCDS now has a [translation and localization policy](http://standard.open-contracting.org/1.1/en/support/governance/#translation-and-localization-policy).
+* [#744](https://github.com/open-contracting/standard/pull/744) OCDS now has a [translation and localization policy](http://standard.open-contracting.org/1.1/en/support/governance/#translation-and-localization-policy). [#808](https://github.com/open-contracting/standard/pull/806) [#808](https://github.com/open-contracting/standard/pull/808) The documentation is now available in Italian.
```
duncandewhurst: PR #806 incorrectly labelled as #808. Also, should the italian translation have it's own bullet point? (currently it follows on from translation and localization policy bullet)

jpmckinney:
@duncandewhurst This PR is still WIP as I need to further clarify merging.md

duncandewhurst:
Ah, OK - let me know when I should restart my review.

jpmckinney:
@duncandewhurst The PR is again ready for review. merging.md was rewritten. I recommend instead reading it at http://standard.open-contracting.org/merge-rules/en/schema/merging/ (once it's built)

standard/docs/en/schema/merging.md
```diff
@@ -51,106 +52,163 @@ If the record is complete with versioning information, then the versioning secti
...
+* For a versioned release:
+  * Merge the field's values in **output** and **input** according to the merge routine; if there is a result, add the field to the object in **output** if not already added, and set it to the result
 
-### Example
+#### Literal values
+
+If the **input** value is neither an object nor an array, then:
+
+* For a compiled release:
+  * If the **input** value is not `null` and is different from the **output** value, replace the **output** value with the **input** value
```
duncandewhurst: > If the **input** value is not `null`...

Since a literal value will never be encountered other than as a property of an object, I think this case is already covered by the "If the field in **input** has a value of `null`..." rule in the object value section.

We should either remove the "If the **input** value is not `null`..." clause here, or if repeating it for clarity, we should also repeat what to do if the field is `null`.

standard/docs/en/schema/merging.md
```diff
@@ -51,106 +52,163 @@ If the record is complete with versioning information, then the versioning secti
...
+  * If there is no **output** value, set the **output** value to an empty JSON array, convert the **input** value to a [versioned value](#versioned-values), and append it to the new array of versioned values in **output**
+  * If the **input** value is different from the value of the `value` field of the most recent versioned value in **output**, convert the **input** value to a [versioned value](#versioned-values), and append it to the array of versioned values in **output**
+
+#### Array values
+
+If the **input** array contains anything other than objects, treat the array as a literal value. Otherwise, there are two sub-routines for arrays of objects: whole list merge and identifier merge.
+
+##### Whole list merge
+
+An **input** array should be treated as a literal value if the corresponding field in a [dereferenced copy](https://jsonref.readthedocs.io/en/latest/) of the release schema has `"array"` in its `type` and if any of the following are also true:
```
duncandewhurst: We could provide a compiled schema in OCDS for the convenience of reimplementations of merging written in something other than python.

There's some work happening in BODS to add generate a dereferenced schema as part of the standard build process which we might be able to re-use.

standard/docs/en/schema/merging.md
```diff
@@ -1,30 +1,31 @@
...
+**Merging** is the process of combining individual releases into a compiled or versioned release, described in more detail below.
 
-A publisher provides a tender release on 1st January, with a planned contract value of $1000.
+The versioned release makes it easy to see how field values have changed over time: for example, how contract values have been modified or how milestones have been re-scheduled during implementation.
 
-On 5th February, the publisher provides an amended tender release updating the planned contract value to $2000.
+<div class="example hint" markdown=1>
 
-After assessing bids, it is decided to award the contract in two lots.
+<p class="first admonition-title">Worked Example</p>
```
duncandewhurst, line 15: It would be good to expand the example (or add another example) demonstrating null-ing for values, objects and arrays.

I don't think we need to do that now, but I added it to the list of suggested examples for OCDS 1.1.5

duncandewhurst:
I reviewed the built version of `merging.md` and suggested that @mrshll1001 reads through the [built page](http://standard.open-contracting.org/merge-rules/en/schema/merging/) too as a fresh pair of eyes, since I already understand merging.

I'll continue reviewing the other changes

standard/docs/en/schema/merging.md
```diff
@@ -1,30 +1,31 @@
...
 
-<div class="example hint" markdown=1>
+* a compiled release, which expresses the current state of the contracting process, by showing only the most recent field values
+* a versioned release, which expresses all historical states of the contracting process, by showing all the field values over time
 
-<p class="first admonition-title">Worked Example</p>
+**Merging** is the process of combining individual releases into a compiled or versioned release, described in more detail below.
 
-A publisher provides a tender release on 1st January, with a planned contract value of $1000.
+The versioned release makes it easy to see how field values have changed over time: for example, how contract values have been modified or how milestones have been re-scheduled during implementation.
```
mrshll1001: This is a bit jarring in context of the previous paragraphs. We go from "Records are this." to "Records make use of merging" to "Merging is used for compiled and versioned releases and here are details" to "Here's a repeat of the versioned release details".

Recommend merging this content with line 6 to avoid duplication and creating a better flow for the introduction.

standard/docs/en/schema/merging.md
```diff
@@ -51,106 +52,163 @@ If the record is complete with versioning information, then the versioning secti
...
+  "releaseDate": "2016-01-01T09:30:00Z",
+  "releaseTag": [
+    "tender"
+  ],
+  "value": 1000
+}
+```
 
-i.e. if the list exists in **new**, the entire list in **new** will overwrite the list in **old**. 
+In a **versioned release**, with a few exceptions, a field's value is replaced with an array of versioned values, which should be in chronological order by `releaseDate`.
```
mrshll1001, line 128: This information is critical to provide to users and is well explained here :-) I suggest it be signposted earlier in the document in the introduction. I'll make a comment where I think it would be good to include this.

standard/docs/en/schema/merging.md
```diff
@@ -1,30 +1,31 @@
...
 
-In OCDS, merging involves combining individual [releases](../getting_started/releases_and_records.md) of data during a contracting process into a [record](../getting_started/releases_and_records.md) which provides an overview of the current state of that process. Including a versioned release as part of the merged record provides structured information on the history of the process: showing when updates were made, or values were changed. 
+An OCDS [record](../getting_started/releases_and_records.md) aggregates all the releases available for a contracting process at a given time, and optionally includes:
 
-<div class="example hint" markdown=1>
+* a compiled release, which expresses the current state of the contracting process, by showing only the most recent field values
+* a versioned release, which expresses all historical states of the contracting process, by showing all the field values over time
 
-<p class="first admonition-title">Worked Example</p>
+**Merging** is the process of combining individual releases into a compiled or versioned release, described in more detail below.
```
mrshll1001, line 9: As per my later comment about essential information, I suggest signposting the technical processes described later in detail.

The bullet points on lines 5 and 6 do a good job of explaining the differences between compiled and versioned releases. We should also indicate at a high level what the implementation of merging will look like.

I suggest something to the effect of "A compiled release is produced by taking the latest version of each field from the releases in a contracting process" and "A versioned release is produced by replacing a field's value with an array of versioned values, which should be in chronological order by `releaseDate`"

standard/docs/en/schema/merging.md
```diff
@@ -51,106 +52,163 @@ If the record is complete with versioning information, then the versioning secti
...
 
 ```
 
 </div>
 
-## Merging rules
+## Merging specification
 
-Order all the releases that share an `ocid` by their release `date`. Starting from the oldest release (old), compare it with the next oldest release (new), and apply the following rules.
+### Discarded fields
```
mrshll1001, line 87: Suggest adding a brief explanation of why these fields are discarded. It *should* be obvious to people who know the standard enough to follow the merging advice, but worth addressing explicitly.

standard/docs/en/schema/merging.md
```diff
@@ -51,106 +52,163 @@ If the record is complete with versioning information, then the versioning secti
...
 
-If the value of a key-value pair is a list of objects, the list merge rules below should be used; otherwise, a list should be treated as a single value.
+```eval_rst
+.. note::
 
-### Lists
+  The compiled release presently uses the same schema as the release schema, which means that the ``id``, ``date`` and ``tag`` fields are required in a compiled release. We invite discussion on whether to change these requirements in a separate compiled release schema in issue `#330 <https://github.com/open-contracting/standard/issues/330>`__, and on how to identify and date compiled and versioned releases in issue `#834 <https://github.com/open-contracting/standard/issues/834>`__.
 
-There are two merge patterns for lists of objects: identifier merge, and whole list merge. 
+  In the meantime, an intermediate solution is to set ``tag`` to ``["compiled"]``, ``date`` to the date of the most recent release, and ``id`` to ``{ocid}-{date}``, like in the `reference implementation <#reference-implementation>`__ of the merge routine.
```
mrshll1001, line 102: Suggest every time "merge routine" is mentioned, it is hyperlinked back to the relevant heading. This supports users who make heavy use of swapping in their short-term memories ;-)

standard/docs/en/schema/merging.md
```diff
@@ -51,106 +52,163 @@ If the record is complete with versioning information, then the versioning secti
...
-### Example
+#### Literal values
+
+If the **input** value is neither an object nor an array, then:
+
+* For a compiled release:
+  * If the **input** value is not `null` and is different from the **output** value, replace the **output** value with the **input** value
+* For a versioned release:
+  * If there is no **output** value, set the **output** value to an empty JSON array, convert the **input** value to a [versioned value](#versioned-values), and append it to the new array of versioned values in **output**
+  * If the **input** value is different from the value of the `value` field of the most recent versioned value in **output**, convert the **input** value to a [versioned value](#versioned-values), and append it to the array of versioned values in **output**
```
mrshll1001, line 182: Lot of use of the word *value* here. Might be unavoidable due to the `value` field. I suggest looking at potential rephrasing as it's a bit jarring to read.

standard/docs/en/schema/merging.md
```diff
@@ -51,106 +52,163 @@ If the record is complete with versioning information, then the versioning secti
...
+   :jsonpointer:
+   :expand: records, versionedRelease, tag, tender, awards
    :title: versioned
 
 ```
 
 </div>
 
-## Merging rules
+## Merging specification
```
mrshll1001, line 84: Suggest reviewing structure of subsections here to make it easier to find information pertaining to specific types of release (compiled vs versioned).

Current subsections describing specific areas are good, and would be enhanced by a consistent structure ie

```markdown
### Objects
#### In a Compiled Release
....
#### In a Versioned Release
```

mrshll1001: I did notice that this is somewhat already present in some subsections. I think we could be more consistent.

duncandewhurst (approved):
Of my comments, only https://github.com/open-contracting/standard/pull/810#discussion_r269912163 is important to resolve before merging.

standard/docs/en/schema/changelog.md
jpmckinney: Fixed in f3c05cdab953293d305fbb6ca86677d0a3db246c

standard/docs/en/schema/merging.md
jpmckinney: Removed in fb0aff6

standard/docs/en/schema/merging.md
jpmckinney: I added a dereferenced release schema and linked to it in fb0aff6

standard/docs/en/schema/merging.md
jpmckinney: Removed in fb0aff6

standard/docs/en/schema/merging.md
jpmckinney: I don't know how to rephrase, but we can revisit at another time. Replacing "versioned value" with another term would help, but I can't think of one.

standard/docs/en/schema/merging.md
jpmckinney: Yeah, that last paragraph was on the original page and I didn't know where to put it. The worked example demonstrates the utility of the versioned release, so I deleted this paragraph in 50ee7f9

standard/docs/en/schema/merging.md
jpmckinney: Added bullet points in 50ee7f9

standard/docs/en/schema/merging.md
jpmckinney: Added in 50ee7f9

standard/docs/en/schema/merging.md
jpmckinney: Done throughout the merge routine section in 50ee7f9

standard/docs/en/schema/merging.md
jpmckinney: "Discarded fields" is the only section in which both types of releases are mentioned but not segmented visually using indented lists. Fixed in 50ee7f9

jpmckinney:
I've resolved all comments, except two (example to be part of 1.1.5, and repetition of value to be worked on later https://github.com/open-contracting/standard/pull/810#discussion_r270074822) that don't block merging.

jpmckinney:
I added another commit relating to #831, in which all norms are changed to MUST except for the chronological order of versioned values, which was never communicated anywhere before, so it can be at most a SHOULD.

duncandewhurst (approved):
Changes look good.

I realised that implicit in the ordering part of the merge routine is a requirement that `date` is unique within the scope of a contracting process - I've opened https://github.com/open-contracting/standard/issues/848 to suggest we make that explicit in the schema.

This actually came up as an issue in the Australia data where the publisher was updating the release date for all releases relating to a contracting process whenever anything changed.

jpmckinney:
Thanks! We'll have an opportunity to further refine this page in 1.1.5, so merging.

PR #824, opened 2019-03-21
Generate translated versions of codelist files for download

### LINKED ISSUE #673: Publish codelists with the standard
We presently publish schemas at standard.open-contracting.org, but not codelists. Codelists are presently only accessible at GitHub URLs.

~~Core extensions’ codelists should be included among the published codelists.~~

jpmckinney:
closes #673

I ran a diff before and after, and saw no difference in the HTML pages.

Deployment of codelists and ZIP file is covered in https://github.com/open-contracting/standard-development-handbook/commit/09c4f3655de2c42a23f5ef7ef65403bffb0de9ad


duncandewhurst (approved):
I checked http://standard.open-contracting.org/publishcodelists/es/getting_started/building_blocks/ and the procurement method codelist displays OK in both english and spanish docs

PR #822, opened 2019-03-20
Integrate Extension Explorer

jpmckinney:
Fixes #803

* Merge extensions/party_details into schema/reference and extensions/developing into extensions. Use extensionexplorerlinklist on extensions page ([preview](http://standard.open-contracting.org/extensionexplorer/en/extensions/#using-extensions)).
* Remove community extensions page, page translation, toctree item, .tx/config entry, test. Remove extensionselectortable tables and related JavaScript.
* Remove core extension pages, page translations, codelists, toctree items, .tx/config entries. Update core extensions URLs in changelog.
* Update references to bids and location extensions on schema reference page.

Also, fixes #743

PR #837, opened 2019-03-25
Add link to OCDS for PPPs

jpmckinney:
* [x] Includes #822 so merge that first.


duncandewhurst (approved):
I checked the changes from the `extensionexplorer` branch using https://github.com/open-contracting/standard/compare/extensionexplorer...issue459

Issue #605, opened 2017-10-23
Improve packaging documentation


jpmckinney:
We introduce the concept of [releases and records](http://standard.open-contracting.org/latest/en/getting_started/releases_and_records/) on one page, and we briefly talk about how to package them [halfway down another page](http://standard.open-contracting.org/latest/en/getting_started/publication_patterns/#packaging-releases-and-records). There's also some discussion of [APIs](http://standard.open-contracting.org/latest/en/implementation/hosting/#bulk-downloads). This is insufficient, based on common errors and questions we receive.

I intend to do a full audit of the documentation at a later date, but my initial thoughts on this specific issue is that the current documentation expects the reader to proceed linearly through the "getting started" section, reading everything. We know that this is not how many people approach the documentation.

One potential solution is to author a new page (reusing existing content as applicable), which describes:

1. A release (the fundamental and first thing a publisher will produce)
2. A release package, and the various reasons/processes/patterns for which a publisher might want to package releases (i.e. when/why you would want to use a release package)
3. A record, and how it relates to releases
4. A record package, hitting the same points as for a release package

My anecdotal evidence is that fewer publishers use records/record packages that we would anticipate, especially considering that a record is an important concept (I think all publishers should have records). We aren't communicating these concepts clearly enough, and I think a stepped approach as above that more closely matches the incremental implementation of OCDS would better serve publishers.

timgdavies:
Agreed. I think more worked examples, and some really good documentation design can help us here to better communicate the key points.

jpmckinney:
Documenting a couple possible patterns to include in the documentation:

1. If a publisher intends to provide a single bulk download of all *releases*, they should publish a single release package containing all the releases.
1. If a publisher intends to provide a single bulk download of all releases *organized by contracting process*, they should publish a single record package containing one record per contracting process, with each record containing all the releases for that contracting process.

We can provide sample JSON to illustrate each pattern.

jpmckinney:
cc @juanpane

jpmckinney:
Related: #389

duncandewhurst:
I added a related issue on how the package schemas are structured: #716 

> My anecdotal evidence is that fewer publishers use records/record packages that we would anticipate, especially considering that a record is an important concept (I think all publishers should have records). We aren't communicating these concepts clearly enough, and I think a stepped approach as above that more closely matches the incremental implementation of OCDS would better serve publishers.

I agree with this but it's worth noting that not all publishers can have records.

[...]

PR #838, opened 2019-03-25
Clarify uniqueness of release, award and contract identifiers

### LINKED ISSUE #455: Clarify the scope of uniqueness of Release ID
In OCDS 1.0 & 1.1 we define ```release.id``` as:

> "An identifier for this particular release of information. A release identifier must be unique within the scope of it’s related contracting process (defined by a common ocid), and unique within any release package it appears in. A release identifier must not contain the # character." 

However:

(a) This is conceptually subtle and difficult to communicate;
(b) As we see people starting to aggregate data from multiple sources, this creates a challenge, as an aggregator of information might end up with releases from multiple sources, which have distinct ```ocid``` values, but the same ```release.id```. Strictly then, the re-publisher would have to manipulate the release ids if it wanted to present both in the same package.

This leads to a suggestion from @edugomez that we should just make ```release.id``` globally unique. 

This would be simply enough for people to implement, using their OCDS prefix, and OCID. 

It would likely be a 2.0 issue.

### LINKED ISSUE #682: Correct the explanation of 'unique' with respect to release IDs and tender, award and contract IDs
http://standard.open-contracting.org/latest/en/schema/identifiers/

> The tender, award and contract IDs must be unique within the contracting process it is part of.

> For any given ocid, there should be no two tender, award or contract IDs that refer to different tender, award or contract sections respectively.

The second paragraph is confusing, because two (different) IDs can of course refer to two different sections.

The second paragraph either needs to be rephrased or removed.

Originally reported by @colinmaudry via Transifex.

jpmckinney:
closes #682, closes #455 

There is only one `tender/id` field, so I think it makes more sense to omit anything about the uniqueness of that field.

Another significant change is described in the changelog: Eliminate the conflicting rule that release identifiers must be unique within release packages. Uniqueness within the scope of a release package either implies that release identifiers must be globally unique, or imposes an arbitrary restriction on the contents of release packages, as there is no rule preventing release packages from containing releases from different contracting processes. Release identifiers are only required to be locally unique within the scope of a contracting process. The `ocid` and `id` values of a release can be used together to disambiguate releases within a release package.


duncandewhurst:
I think the reason that global uniqueness for release identifiers was suggested in #455 is because the description of the `url` field in the linked releases part of the record schema is as follows:

> The URL of the release which contains the URL of the package with the releaseID appended using a fragment identifier e.g. http://standard.open-contracting.org/latest/en/examples/tender.json#ocds-213czf-000-00001

jpmckinney:
Hmm, I think it's very likely that release IDs are not, in practice, globally unique, as that was never stated anywhere clearly, and can only be determined by following the implications of multiple disparate pieces of information. So, with respect to the suggestion in #455, I believe that making the rule that release IDs must be globally unique is not an acceptable resolution to the lack of clarity that this PR intends to resolve.

The only other mention I find of this way of identifying releases inside release packages is in `records_reference.md`: "In order to locate the specific release inside a release package the releaseID of the release should be appended to the package URL using a fragment identifier."

Rather than forcing release IDs to be globally unique, we can simply clarify that, if the `url` of a linked release matches multiple releases, then the `ocid` (which is required in the record package schema) must be used to disambiguate the releases.

If that's fine, I'll create an issue for 1.1.5 about that – as I think we should explore other solutions, like changing the URL pattern to not be ambiguous.

For 1.1.4, I think it's still important to clarify that release IDs need only be locally unique, as that is the practice and what I consider to be the intention (it's not clear that the implication of global uniqueness was understood by the authors, as otherwise it would have been much simpler to simply say so, rather than have the complex combination of unclear statements that we have now).

duncandewhurst:
Yes, that sounds good. Just flagging that those other references will need updating but I think it's fine to do that 1.1.5.

This reminded me why publishing releases in packages is a must (so that linked releases can be fetched in a consistent way) so I added a note to #605

duncandewhurst (approved):

jpmckinney:
I created an issue about those references and set it to the 1.1.5 milestone #844

PR #842, opened 2019-03-26
Update currency codelist for ISO4217 amendment 169

jpmckinney:


duncandewhurst (approved):
I checked against https://www.currency-iso.org/en/shared/amendments/iso-4217-amendment.html

We might want to update the changelog entry to explain what the change adds (since there is no explanation in the linked PR)

jpmckinney:
In previous versions, we simply mentioned the most recent amendment number. I updated the changelog to link to each amendment's description to avoid repeating ISO's own messages.

The update is automated using `standard/schema/utils/fetch_currency_codelist.py`

PR #852, opened 2019-03-29
Fix normative language in codelists

jpmckinney:
Addresses #830

* [x] Includes #831, so merge that first. Diff from #831 is `https://github.com/open-contracting/standard/compare/normative...normativeschema`

No surprises here. Please check the spreadsheet for any rows where "Status" has "Disagree" (you can use the spreadsheet's filters in the header row, or manually search the spreadsheet).

PR #831, opened 2019-03-22
Implement norms review in documentation

jpmckinney:
addresses documentation part of #830

This iteration focuses on using the correct keyword. A second PR will copy/move statements into normative sections if the norm only appears in a non-normative section. I haven't found a reasonable way to link between re-statements of the same norm, so that suggestion won't be implemented until a way is found.

Regarding normative keywords:

* Please check the spreadsheet for any rows where "Status" has "Disagree" (you can use the spreadsheet's filters in the header row, or manually search the spreadsheet).
* It didn't come up in the review, but the docs have "publishers are encouraged to register" an ocid prefix. I changed this to MUST. In a later iteration, I need to copy this into a normative section.
* I avoided rewriting sentences and paragraphs – as that would be a very large project (for 1.1.5) – and instead focused as much as possible on just cleaning up the normative words. I did rewrite sentences when it was harder to keep the current sentence.

Besides changing normative/non-normative keywords, this PR also covers these points of the norms review summary:

* (8) Explicit statements of optional fields
* (9) Explicit statements of optional extensions

This PR also includes these changes:

1. governance.md has a new paragraph about this normative content policy.
1. The documentation about null values at the [bottom of reference.md](http://standard.open-contracting.org/latest/en/schema/reference/#emptying-fields-and-values) is… wrong. It's replaced with a reference to the merging documentation.
1. A sentence in reference.md referred to a non-existent 'notice' code, so I deleted it.
1. Replaced "publisher prefix" with "ocid prefix" (involves rephrasing) in contracting_process.md
1. The example in releases_and_records.md for iterating a release `id` was too confusing. The description of compiled and versioned releases was also copied from the new merging.md
1. Remove link to closed issue about community discussion of API specification.
1. The last paragraph in deprecation.md was unclear ("decimal versions" is not defined or used anywhere else).
1. `releaseDate` and `releaseID` were used to refer to the `date` and `id` fields in the release. This has been corrected throughout.
1. Fixed release-package-schema.json in line with #838

----

This paragraph is wrong (we have multiple unit classification schemes now) and ought to be rewritten:

> Although no codelist for units has been established in the current release of the standard, publishers may consider using the Units provided by the [Quantities, Units, Dimensions and Data Types Ontologies](http://www.qudt.org/qudt/owl/1.0.0/unit/) in the `unit.name` field (drawing on the CamelCase unit names, such as SquareMile), in order to provide detailed information the cost per unit of a line-item.

Issue #830, opened 2019-03-22
Use non-normative language in non-normative sections


jpmckinney:
Following an internal review of documentation, schema, codelists and those of core extensions, we will be more consistent in our use of normative terms, based on our [normative and non-normative content policy](https://docs.google.com/document/d/1xjlAneqgewZvHh6_hwuQ98hbjxRcA2IUqOTJiNGcOf8/edit). Steps anticipated under the policy:

* Clearly mark documentation pages or sections as normative or non-normative. [#846]
* Use RFC 2119 keywords (MUST, SHOULD, MAY, etc.) where appropriate. [this issue]
* Expand the conformance statement with a summary of OCDS norms. [#580]
* Update the Governance page with more detail on versioning. [#674]
* Update the Governance page with detail on how the governance process itself can be updated, including changes to the license and copyright holder of the standard. [#832]

The `must|require|shall|should|recommend|may|optional` (case-insensitive) regular expression can be used to find normative keywords.

jpmckinney:
* [x] Write PR to implement norms review in documentation #831
* [x] Write PR to implement norms review in codelists #852
* [x] Write PR to implement norms review in schema #855
* [x] Write PR to copy/move statements into normative sections if the norm only appears in a non-normative sections

PR #855, opened 2019-04-01
Fix normative language in schema

jpmckinney:
Addresses #830

* [x] Includes #831 and #852, so merge those first. Diff from #852 is `https://github.com/open-contracting/standard/compare/normativeschema...normativeschema2`

No surprises here. Please check the Schema sheet of the spreadsheet for any rows where "Status" has "Disagree" (you can use the filters in the header row, or manually search the spreadsheet).


jpmckinney:
@duncandewhurst WIP, so no review yet, but I noticed that codelist fields are inconsistent in saying things like:

* This is a closed codelist. Local XYZ should be mapped to this list. (2x)
* This is an open codelist. Local XYZ can be included in this list.
* Milestone codes is an open codelist, and codes should be agreed among data producers and the applications using that data.
* Values from the provided codelist should be used wherever possible, though extended values can be provided if the codelist does not have a relevant code. (2x)

I think it's better to omit such information, and instead ensure that the codelists documentation is clear on the rules around open and closed codelists (perhaps borrowing the improved language from OC4IDS). What do you think? It will save cleaning up the normative keywords in these clauses, which are mostly wrong ("should" instead of "must").

duncandewhurst:
Yes, this makes sense to me.

I think we should state whether they are open or closed in the schema descriptions but keep the explanations of what that means in the docs.

jpmckinney:
@duncandewhurst Ready for review.

standard/schema/release-schema.json
```diff
@@ -136,7 +136,7 @@
   "definitions": {
     "Planning": {
       "title": "Planning",
-      "description": "Information from the planning phase of the contracting process. Note that many other fields may be filled in a planning release, in the appropriate fields in other schema sections, these would likely be estimates at this stage e.g. totalValue in tender",
+      "description": "Information from the planning phase of the contracting process. Note that many other fields can be filled in a planning release, in the appropriate fields in other schema sections; these would likely be estimates at this stage, e.g. totalValue in tender.",
```
duncandewhurst: `totalValue` doesn't exist, the field should be `value`

duncandewhurst (approved):
PR looks good, I left one comment on an error we should correct and noted one other thing:

The [related process](http://standard.open-contracting.org/1.1-dev/en/schema/codelists/#related-process) section of the codelist reference still mentions PQQs:

> The related process block is used at the release level to point backwards to prior processes, such as planning, PQQ or framework establishment...

This PR removes the mention of pre-qualifications from the related process schema, so we should remove it from the reference page too.

Neither are blockers, so I've approved the PR

jpmckinney:
Good catches! I made the two changes and will merge.

Issue #440, opened 2017-04-19
Handling multi-stage contracting processes


timgdavies:
This issue builds on the review comments in #371, and discussions with the technical team and implementers, to set out an approach for OCDS 1.1 for handling multi-stage contracting processes, and sets a direction towards 2.0. 

### Current situation

The concept of a **contracting process** is central to OCDS. By bringing together multiple releases of data and information related to a single contracting process, OCDS enables citizens, companies and governments to get a clearer view of contracting activities.

The standard defines a contracting process as:

> All the planning, tendering information, awards, contracts and contract implementation information related to a single initiation process.

Each contracting process is identified by an Open Contracting ID (```ocid```). 

The standard currently contains a single ```tender``` block, and then arrays of one or more ```award``` and ```contract``` (and nested within contract, ```implementation```) blocks. 

This pattern, of a single ```tender``` block, was developed to avoid publication mistakes where tenders from different contracting processes could end up mixed together under the same ```ocid```. 

### Limitations

The current model has major limitations in the case of contracting processes where multiple tender notices or solicitations are issued. This includes:

1. Two-stage processes pre-qualification shorlisting, and then a full invitation to tender;
2. Competitive dialogues - which may have repeated rounds;
3. Processes where the initial tender does not lead to an award, and so a second call for proposals is made;
4. Framework contracts with mini-competitions;

### Initial proposal and review

In proposals for OCDS 1.1, we sought to address this by creating a ```relatedProcess``` block that could link two or more contracting processes. This would allow, for example, one contracting process to contain the pre-qualification tender, and award onto a shorlist, and a second, related process, to contain the invitation to qualified bidders and the final award of the contract. 

However, peer-review feedback noted that this breaks the idea of bringing together all information about a contracting process under a single ```ocid```. 

The reviewer rejected the introduction of ```relatedProcess``` to deal with cases of frameworks, pre-qualification, competitive dialogue and other multi-stage processes. 

We agree with the reviewer that the handling of pre-qualification and other multi-stage processes requires attention. 

We believe that ```relatedProcess``` still has a role in some forms of framework, particularly cases where one agency establishes the framework, and other agencies are then running their own independent contracting processes to buy from that framework. 

### Alternative models and their implications

In this section we scope out alternative approaches to allow us to represent multi-stage contracting processes. This is guided by:

* Maintaining compatability with existing OCDS tools and data; 
* Avoiding multiple ways of expressing the same thing;
[...]

PR #890, opened 2019-07-08
Add examples for framework agreements

jpmckinney:
Making in a PR so that I can review.

### DIFF
--- a/standard/docs/en/examples/frameworks/example_mini-competition/005_framework_mini-competition_tender.json
+++ b/standard/docs/en/examples/frameworks/example_mini-competition/005_framework_mini-competition_tender.json
@@ -1,0 +1,65 @@
+{
+  "ocid": "ocds-r6ebe6-minicompetiion_from_other_publisher_framework",
+  "id": "ocds-r6ebe6-minicompetiion_from_other_publisher_framework-tender-2019-05-01T00:00:00Z",
+  "date": "2019-05-01T00:00:00Z",
+  "tag": [
+    "tender"
+  ],
+  "initiationType": "tender",
+  "buyer": {
+    "name": "Edinburgh",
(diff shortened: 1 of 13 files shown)


jpmckinney:
This might need significant changes, depending on how #440 is resolved (see latest comment there).

yolile:
Some comments from discussions/retreat;

- It is preferable to always use a new process for each call off even if it is a direct call off without competition and then, for direct call offs:
  * Add an `Award` object to `awards` and a `Contract` object to `contracts`
  * Set `awards.id` and `contracts.id` to the same value
  * Populate the `awards.value`, `awards.contractPeriod` and `awards.items` fields with the initial contract value, period and items respectively
  * If the contract value, period or items are subsequently updated, populate the `contracts.value`, `contracts.period` and `contracts.items` fields with the updated value

- We need a field to distinguish the FA tender from the traditional ones: we can use the techniques extension https://github.com/open-contracting-extensions/ocds_techniques_extension
- We need a field to distinguish the mini competition tenders from the direct call offs: the direct call offs are going to be modeled as a new process without the tender block.
- FA duration: the duration is mappend in the tender.contractPeriod field
- To map all the buyers who participate in the FA dont map the buyer object but include multiples parties with 'buyer' role
- For indicate who buyer is buying what, as each call off is a new process, use the buyer object in that new process.
- As each mini competition or call of is a new process, use the planning block to indicate the budget for that purchase.


cc @jpmckinney

jpmckinney:
@yolile: Can you work with @duncandewhurst next week to rewrite this page based on the discussion at the retreat?

jpmckinney:
Closing this PR as I think the updates will be easier to review as a new PR.

I assume @yolile's notes from the retreat will all be reflected in the new PR. If not, please create/update an issue with any outstanding notes.

PR #905, opened 2019-08-07
Standard Search Server has moved

ghost:


deploy.sh
```diff
@@ -22,4 +22,4 @@ echo "Copy the built files to the remote server..."
 $LFTP -c "set sftp:connect-program \"ssh -i id_rsa\"; connect sftp://ocds-docs:xxx@staging.docs.opencontracting.uk0.bigv.io; mirror -eRv build web/$TRAVIS_BRANCH"
 
 echo "Update the search index..."
-curl "www.standard-search.default.opendataservices.uk0.bigv.io/v1/index_ocds?secret=${SEARCH_SECRET}&version=${TRAVIS_BRANCH}"
+curl "www.live.standard-search.opencontracting.uk0.bigv.io/v1/index_ocds?secret=${SEARCH_SECRET}&version=${TRAVIS_BRANCH}"
```
jpmckinney, line 25: Like in https://github.com/open-contracting/standard_theme/pull/6, do we want the FQDN to be `www.live.…` instead of just `live.…`?

PR #915, opened 2019-08-12
Standard Search server has moved to a CNAME

ghost:


ghost:
[Use standard-search.open-contracting.org instead of bigv.io FQDN](https://trello.com/c/q5gRnAIz/65-use-standard-searchopen-contractingorg-instead-of-bigvio-fqdn)

ghost:
@jpmckinney Is there a problem with this one so it can't be merged ( but https://github.com/open-contracting/standard_theme/pull/7 can? )

jpmckinney:
@odscjames No, I just missed it :)

Issue #759, opened 2018-10-12
Improve publication policy guidance


jpmckinney:
The present guidance is: http://standard.open-contracting.org/latest/en/implementation/publication_policy/

There's an opportunity to improve this based on the recent [community call](https://drive.google.com/drive/folders/1x-aboyQ7ppvWHHKyYCRrbwc9hAJkHuGy) and guidance given privately to publishers (e.g. CRM issue 3554 update 22).

cc @LindseyAM

pindec:
The new  [publication policy template is now published](https://docs.google.com/document/d/1IEZhotGcgdC7XBmLepdPQRVN5y4b5l9H0y0h32HAQyo/view) so we should add a link to it from [standards docs](http://standard.open-contracting.org/latest/en/implementation/publication_policy/) and/or otherwise update the standards docs page.

jpmckinney:
Added 1.1.5 milestone.

jpmckinney:
@LindseyAM Here's the OC4IDS page for 0.9.2: https://standard.open-contracting.org/infrastructure/0.9-dev/en/guidance/data_user_guide/

jpmckinney:
Fixed in #986.

PR #929, opened 2019-10-15
Update "Publication Policies"

aguilerapy:
Change the name from "Publication Policies" to "Publication Policies: Data User Guide".

Based on CRM issue 5167.

### DIFF
--- a/standard/docs/en/implementation/publication_policy.md
+++ b/standard/docs/en/implementation/publication_policy.md
@@ -1,17 +1,26 @@
-# Publication policies: Data User Guide
+# Publication policies
+
+This guidance helps users to understand a publishers OCDS data, its call Publication Policies: **Data User Guide**.
 
 Publishing Open Contracting data involves making a number of choices about what data and documents to include and how to classify and categorize different items. To allow users to understand the choices that a publisher has made, a publisher ought to provide a public web page or document that details the decisions they have made concerning the issues listed below.
 
 This document ought to also contain information about future plans for changes to the data and who to contact with enquiries about the data.
 
-Publication policy pages ought to include information about:
+Publication Policies: Data User Guide pages ought to include information about:
 
-* Who is responsible for providing this data - including appropriate contact details and locations of where to offer feedback.
-* How the data is generated, and how often - for example, a brief summary describing the systems the data is generated from, how often the source data is updated, and any processes that operate between the source data and the published open contracting data and documents.
-* Any exclusions from the data - such as information which is not disclosed for reasons of privacy and security. This can describe the fields which are excluded for a selection of records, and/or the policy applied to decide if any information is withheld from a public OCDS release. 
-* Any custom codelists used - or custom codes added to existing codelists. 
-* Future development plans - brief notes on the future plans of the publisher to improve their data.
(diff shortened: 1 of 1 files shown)


jpmckinney:
@aguilerapy This will make the menu item in the left-hand sidebar very long (maybe longer than the sidebar): https://standard.open-contracting.org/latest/en/implementation/publication_policy/

Also, I think we'd want to edit some of the text on the page to more consistently use 'data user guide' instead of (or along with) publication polic*.

aguilerapy:
2019-10-18-140152_1121x670_scrot

@jpmckinney The left-hand sidebar don't change, the text "Guide" ran to the second line. 

2019-10-18-140210_998x678_scrot

We can also leave "Data User Guide" as subtitle in publication_policy.md what do you think?

jpmckinney:
Yes, my concern is that it goes onto two lines, whereas we try to keep each menu item on one line.

The subtitle will unfortunately create an extra layer of hierarchy, which isn't something we want.

I think we should focus on editing the text to make it clearer that this policy should function as a data user guide (we can bold those words). 

We should work to align the content from the newer publication policy guidance ([EN](https://www.open-contracting.org/resources/ocds-1-1-publication-policy-template/), [ES](https://www.open-contracting.org/resources/ocds-1-1-plantilla-de-politica-de-publicacion-es/)) with this page.

That would be a bigger project, but it will have more impact than adding a few words to this page.

aguilerapy:
Okay, I will work on that

standard/docs/en/implementation/publication_policy.md
```diff
@@ -1,17 +1,26 @@
...
+4. Any exclusions from the data - such as information which is not disclosed for reasons of privacy and security. This can describe the fields which are excluded for a selection of records, and/or the policy applied to decide if any information is withheld from a public OCDS release. Highlighting how their legal frameworks and concepts map onto their OCDS publication covering.
+5. Any custom codelists used - or custom codes added to existing codelists.
+6. Who is responsible for providing this data - including appropriate contact details and locations of where to offer feedback on the data or seeking clarification on its use.
+7. Details of the license under which the data is published.
+8. Any relevant additional documentation that the reader should be aware of when using a publishers OCDS dataset.
+9. Future development plans - brief notes on the future plans of the publisher to improve their data.
+10. The limits of the publishers legal responsibilities for the data.
```
aguilerapy: Are the summary sections that are in the template drive document.

jpmckinney:
@aguilerapy Let me know when there is a new version ready for review (no rush).

aguilerapy:
@jpmckinney free to review!

jpmckinney:
@LindseyAM to review

LindseyAM:
Thanks both! I've pulled together some (even shorter) text in this google doc.

My understanding it will fall under the Publish subheading beneath the Guidance Heading in 1.1.5

James, feel free to make any edits and incorporate into documentation

https://docs.google.com/document/d/1cArKYbzYpfVt7WuIPTZz2Ot2JLahhESgc_q6SFeGSnA/edit#

jpmckinney:
After merging, check if #759 is closed.

jpmckinney:
The text from the linked document will likely be merged as part of #986. Once that's done, I can review the changes here, to see if we missed anything.

jpmckinney:
Superseded by #986.

PR #934, opened 2019-11-08
Merge 1.1-dev into 1.1

jpmckinney:
Waiting on translations.


jpmckinney:
Nice – using rsync instead of lftp brings test/deploy down to 4 mins from 15 mins.

PR #948, opened 2019-12-26
Worked example - Organization Identifiers

romifz:


standard/docs/en/implementation/organization-identifiers.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+Note that it is possible to use more than one organization register, in which
+case any additional identifiers can be included in the `additionalIdentifiers`
+section.
+
+Publishers should consider if/when to include [local
+identifiers](../schema/identifiers#choosing-an-identifier).  Local identifiers
+may not be registered in [org-id.guide](http://org-id.guide), provided that the
+organization list is described in a publication policies document.
```
pindec: I think this should be "a publication policy document", and linked to the 'publication policy' section https://standard.open-contracting.org/latest/en/implementation/publication_policy/#publication-policies.

standard/docs/en/implementation/organization-identifiers.md
```diff
@@ -0,0 +1,39 @@
...
+provided in OCDS by identifying the **organization lists** used in the source
+data, choosing an appropiate **organization register prefix** for each list,
+and identifying the organizational ID for each list and organization in the
+data.
+
+It is strongly recommended to use an organization register prefix from
+[org-id.guide](http://org-id.guide), or to [propose a new
+entry](http://docs.org-id.guide/en/latest/contribute/#proposing-a-new-entry)
+for the organization register used if an appropiate entry does not exists. The
+instructions in the [Meta-data
```
pindec: I think it would be clearer if this link title is [org-id meta-data guide] (so it's clear it's an external link).

pindec (commented):
I think we need to clarify how this worked example fits in with the one given in 'getting started' - https://standard.open-contracting.org/latest/en/schema/identifiers/#organization-ids

They should at least be linked in the first paragraph, and we should consider removing the simpler worked example from the 'getting started' in favour of linking to this more complex example.

The implementation index.md page also needs a link to this.

romifz:
Hi @pindec! I updated the example with your suggestions and made two additional changes: 

* Update the supplier information to a real company instead of a fictitious one
* Introduce a small local list example

Please let me know what you think.

standard/docs/en/implementation/organization_identifiers.md
```diff
@@ -0,0 +1,36 @@
...
+    :jsonpointer: /releases/0/parties/1
+    :expand: identifier, additionalIdentifiers
+```
+
+The publisher also collects the VAT identification number for suppliers. The additional identifier is included as an entry in the `additionalIdentifiers` array, using the "GB-VAT" scheme. Note that the VAT registry is not present in [org-id.guide](http://org-id.guide/), but the publisher followed the instructions in the [org-id meta-data guide](http://docs.org-id.guide/en/latest/metadata/#assigning-a-code) to build the "GB-VAT" code used in the `scheme` field: the two-letter country prefix ("GB") plus a short abbreviation for the registry ("VAT").
+
+In this fictional example, the publisher decides to add identifiers from an internal supplier list used in Contracts Finder, the platform used in England to publish opportunities and contract information. A new entry is included in the `parties/additionalIdentifiers` section that looks as follows:
+
+```
+"additionalIdentifiers": [
```
pindec: My only question would be whether this should be presented in the same way as the other examples for consistency.

pindec (commented):
Great, thanks very much, all looks good to me apart from whether the local list example should be presented in the same way as the others.

Then it's ready for @jpmckinney for final review.

standard/docs/en/implementation/organization_identifiers.md
romifz: I added the local list example to the main json file.

PR #950, opened 2019-12-27
Update amendments worked example

romifz:


duncandewhurst:
@jpmckinney do you know why the docs build is breaking?

It'll be easier to review if the built version of the example is available too

jpmckinney:
I forgot to merge an earlier branch that updated the version of commonmark we're using. It should build now.

jpmckinney:
I also had to run `ocdskit indent -r standard/docs/en/examples`

standard/docs/en/implementation/amendments.md
```diff
@@ -16,21 +16,43 @@ The nature of a change can be made explicit using:
 
 * **The amendments** building block. This can contain an array of amendment explanations, and clearly identify the releases that contain before and after values. 
 
-## Worked example
+## Worked examples
 
-In the example below:
+Each example includes sample JSON files located at the bottom of the page. Use the combobox at the end of the page to display the files for each example and step.
```
duncandewhurst: I think it would be better to thread the sample JSON files in between each step, as with the [easy releases example](https://standard.open-contracting.org/docs-easy-releases/en/implementation/easy-releases/) and use expanding/collapsing to highlight the relevant sections.

standard/docs/en/implementation/amendments.md
```diff
@@ -16,21 +16,43 @@ The nature of a change can be made explicit using:
...
-* The tender is then formally amended, with an increased value. This is published in a release with the `tag` 'tenderAmendment', and an `amendments` block under `tender`.
+### Example 1: Tender
 
+This example goes through updates during the **tender** stage in a contracting process.
+
+1. A tender is issued for a "Data merge tool" in a release with the `tag` 'tender'. 
+2. Information is updated in a new release with the `tag` 'tenderUpdate' to provide an expanded description of the item. The publisher did not consider this to be a formal 'amendment' to the tender, so did not publish any amendment information.
+3. Then there is a new update, where the value of the tender increases and the deadline for bid submissions is extended. This is considered as an 'amendment' by the publisher (depending of jurisdiction, both changes require an official approval), and so the new release has the `tag` 'tenderAmendment' and an `amendments` block under `tender`. The release reflects the updated value (USD 2000 instead of USD 1000) and the updated closing date for bid submissions (`2012-02-20` instead of `2012-02-15`).
```
duncandewhurst: depending of jurisdiction, both changes require an official approval -> depending on jurisdiction, certain changes can need to be disclosed as amendment

standard/docs/en/implementation/amendments.md
```diff
@@ -16,21 +16,43 @@ The nature of a change can be made explicit using:
...
+
+Note in this example that:
+
+* **The amendments block does not contain amended values (as in previous versions of OCDS)**. Values are updated in the same fields and using a new release.
```
duncandewhurst:
```suggestion
* **The amendments block does not contain data on what was changed**. Changes are recorded by updating the properties of the `tender` block in a new release. 
```

standard/docs/en/implementation/amendments.md
```diff
@@ -16,21 +16,43 @@ The nature of a change can be made explicit using:
...
+3. Then there is a new update, where the value of the tender increases and the deadline for bid submissions is extended. This is considered as an 'amendment' by the publisher (depending of jurisdiction, both changes require an official approval), and so the new release has the `tag` 'tenderAmendment' and an `amendments` block under `tender`. The release reflects the updated value (USD 2000 instead of USD 1000) and the updated closing date for bid submissions (`2012-02-20` instead of `2012-02-15`).
+
+Note in this example that:
+
+* **The amendments block does not contain amended values (as in previous versions of OCDS)**. Values are updated in the same fields and using a new release. 
+
+* **The publisher chooses in the tenderAmendment release to repeat a fragment of the original 'tender' release**. This is not necessary when a full archive of releases is made accessible, but can be a convenient way of publishers making structured information on the before and after values of a field more accessible to users.
```
duncandewhurst: Add quotes to tenderAmendment

duncandewhurst: > but can be a convenient way of publishers making structured information on the before and after values of a field more accessible to users. 

Unless I misunderstood, a amendment release should not contain the before values?

standard/docs/en/implementation/amendments.md
```diff
@@ -16,21 +16,43 @@ The nature of a change can be made explicit using:
...
-## Worked example
+## Worked examples
 
-In the example below:
+Each example includes sample JSON files located at the bottom of the page. Use the combobox at the end of the page to display the files for each example and step.
 
-* A tender is issued for a "Data merge tool" in a release with the `tag` 'tender'. 
-* This is updated in a release with the `tag` 'tenderUpdate' to provide an expanded description of the item. The publisher did not consider this to be a formal 'amendment' to the tender, so did not publish any amendment information.
-* The tender is then formally amended, with an increased value. This is published in a release with the `tag` 'tenderAmendment', and an `amendments` block under `tender`.
+### Example 1: Tender
```
duncandewhurst:
```suggestion
### Example 1: Tender updates and amendments
```

standard/docs/en/implementation/amendments.md
```diff
@@ -16,21 +16,43 @@ The nature of a change can be made explicit using:
...
+
+Note in this example that:
+
+* **The amendments block does not contain amended values (as in previous versions of OCDS)**. Values are updated in the same fields and using a new release. 
+
+* **The publisher chooses in the tenderAmendment release to repeat a fragment of the original 'tender' release**. This is not necessary when a full archive of releases is made accessible, but can be a convenient way of publishers making structured information on the before and after values of a field more accessible to users. 
+
+* **In the merged record**, the information in the `amendments` array can be used to lookup information in the `versionedRelease` section, to see where changes are explained by an amendment `rationale`. 
+
+### Example 2: Contract
```
duncandewhurst:
```suggestion
### Example 2: Contract amendment
```

standard/docs/en/implementation/amendments.md
```diff
@@ -16,21 +16,43 @@ The nature of a change can be made explicit using:
...
+
+Note in this example that:
+
+* **The amendments block does not contain amended values (as in previous versions of OCDS)**. Values are updated in the same fields and using a new release. 
+
+* **The publisher chooses in the tenderAmendment release to repeat a fragment of the original 'tender' release**. This is not necessary when a full archive of releases is made accessible, but can be a convenient way of publishers making structured information on the before and after values of a field more accessible to users. 
+
+* **In the merged record**, the information in the `amendments` array can be used to lookup information in the `versionedRelease` section, to see where changes are explained by an amendment `rationale`.
```
duncandewhurst: Please update `versioned-tender.json` to reflect the changes in `merge-tender-3.json`

> * **In the merged record**, the information in the `amendments` array can be used to lookup information in the `versionedRelease` section, to see where changes are explained by an amendment `rationale`. 

What information from the amendments array would be looked up in the versionedRelease?

standard/docs/en/implementation/amendments.md
```diff
@@ -16,21 +16,43 @@ The nature of a change can be made explicit using:
...
+
+* **The amendments block does not contain amended values (as in previous versions of OCDS)**. Values are updated in the same fields and using a new release. 
+
+* **The publisher chooses in the tenderAmendment release to repeat a fragment of the original 'tender' release**. This is not necessary when a full archive of releases is made accessible, but can be a convenient way of publishers making structured information on the before and after values of a field more accessible to users. 
+
+* **In the merged record**, the information in the `amendments` array can be used to lookup information in the `versionedRelease` section, to see where changes are explained by an amendment `rationale`. 
+
+### Example 2: Contract
+
+This example shows an update in the value and scope of a contract.
```
duncandewhurst:
```suggestion
This example shows an update to the value and scope of a contract.
```

standard/docs/en/implementation/amendments.md
```diff
@@ -16,21 +16,43 @@ The nature of a change can be made explicit using:
...
+
+1. A contract notice is published for the purchase of domestic appliances. The release uses the 'contract' `tag`.
+2. The scope of the contract is increased to include the purchase of one additional appliance. A single item is added in the `contracts/items` block and the value of the contract is increased.
```
duncandewhurst: Explain that a 'contractUpdate' release is published.

duncandewhurst (changes requested):
Thanks Romina, updates are looking good!

I made some suggestions in `amendments.md`, also:

1. Is there a reason for having separate merged and versioned files for each example, can we just have one record for each example, which has both the compiled and versioned releases?

2. I found it helpful to generate a record with embedded releases for each example and to review them in OCDS-Show, in order to easily see what had changed between each release. Can you add the embedded releases to the records for each example and include an admonition directing the reader to review the examples in OCDS-Show via the Data Review Tool (e.g https://standard.open-contracting.org/review/?source_url=). Currently amendments are missing from the OCDS-Show interface, but I made a PR to correct that!

3. Can we add an extra example explaining how to preserve detail of individual amendments in an implementation which uses the easy releases model?

e.g. in Australia the source system does not store a change history of individual contract notices, however contract amendments are stored as separate notices, therefore even in the easy releases model it is possible to generate an OCDS release for the contract notice (which is kept up to date with the latest value of the contract) **and** an OCDS release for each amendment to the contract notice.

The advantage of this approach (over just maintaining a single release and updating it's properties and release ID each time there is an amendment) is that users have access to the details of each amendment, i.e. the individual changes to contract value or duration, rather than just the state after the most recent amendment.

There is a worked example already documented for this on [Gist](https://gist.github.com/mrshll1001/6ef17e6990587b8da13114a1050257c2) and the notices the example is based on are available in the AusTender front end ([contract notice](https://www.tenders.gov.au/Cn/Show/6dcfcc43-c5cf-7a3e-30ae-7561f054e523), [amendment notice](https://www.tenders.gov.au/Cn/Show/48661147-ac3c-b9b4-7228-6b852602425e))

standard/docs/en/implementation/amendments.md
romifz: This is present in the current text. I don't think it means to say the amendment release contains the before values, but that it might be easier to users to have full blocks of data to compare. I'm not sure this is true though, so I will change the sentence.

standard/docs/en/implementation/amendments.md
romifz: This was in the original example as well. `releaseID` and `amendsReleaseID` can be used to find changes in the `versionedRelease`, I'll clarify this in the text.

duncandewhurst: Do you think that it would make more sense to say that `releaseID` and `amendsReleaseID` can be used to find the releases with the before and after values in the `releases` array?

I am wondering which is easier from a user's perspective: checking through every field in the versioned release to find where both `releaseID` and `amendsReleaseID` appear, or comparing the two releases in the releases array.

standard/docs/en/implementation/amendments.md
duncandewhurst: Great, thanks!

romifz:
Hi @duncandewhurst, thanks for the review. I've included your suggestions except for the Data Review Tool link review using the `source_url` param (I couldn't figure how to build the link using markdown, and I'm not sure about having a fixed link in the docs), so I included instructions to download the records instead. I also added more information to the Australia example.

I changed the location of the JSON files we want to make downloadable to `docs/en/_static/json`, I'm not sure if there is a better location.

I wanted to add a link to the probable location of the Easy Releases guidance, but the checks failed. I've put a comment in the docs to add the link once #951 is merged.

PD: I realize I edited some files I shouldn't have in this branch, it is fixed now.

standard/docs/en/implementation/amendments.md
duncandewhurst: Outstanding typo:

'depending of jurisdiction' > 'depending on jurisdiction'

standard/docs/en/implementation/amendments.md
duncandewhurst: Did you have a view on this?

If we don't know which is easier, we can just update the sentence to mention both, e.g.

> * **In the merged record**, the information in the `amendments` array can be used to lookup information in the `versionedRelease` section or in other releases, to see where changes are explained by an amendment `rationale`.

standard/docs/en/implementation/amendments.md
```diff
@@ -16,57 +16,169 @@ The nature of a change can be made explicit using:
...
-   :title: versionedRecord
+   :expand: records, releases
+   :title: Record
 
 ```
 
+### Example 3: Amendments in a Easy Releases scenario
 
-Note in this example that:
+The Easy releases guidance (<mark>add link</mark>) explains how to build releases when there is no access to the change history for contracting processes. In a scenario when little or no change history is present, it may or may not be possible to identify and build amendments inside releases.
```
duncandewhurst:
```suggestion
The Easy releases guidance (<mark>add link</mark>) explains how to publish releases without storing or publishing a full change history. Depending on the source system, it may still be possible to publish a history of amendments when using this model.
```

standard/docs/en/implementation/amendments.md
```diff
@@ -16,57 +16,169 @@ The nature of a change can be made explicit using:
...
 
-* **The publisher chooses in the tenderAmendment release to repeat a fragment of the original 'tender' release**, so that the before and after values of the tender can be accessed in the single file. This is not necessary when a full archive of releases is made accessible, but can be a convenient way of publishers making structured information on the before and after values of a field more accessible to users. 
+However, it is possible for a publisher to store partial history of events that may be relevant to them, like contract amendments. This example shows how amendment releases can be built when amendment history is present.
```
duncandewhurst:
```suggestion
Where the source system stores a history of contract amendments, either as separate notices or as properties of the original contract notice, contract amendments can be published as separate releases in OCDS.
```

standard/docs/en/implementation/amendments.md
```diff
@@ -16,57 +16,169 @@ The nature of a change can be made explicit using:
...
 
-* **In the merged record**, the information in the `amendments` array can be used to lookup information in the `versionedRelease` section, to see where changes are explained by an amendment `rationale`. 
+A publisher does not register changes in the contracting process. However, since contract amendments are a more frequent ocurrence, these are stored as new contract notices with a reference to the contract being modified.
```
duncandewhurst: I think this is now covered by the suggested change above, but it would be good to replace it with a link to a real life example.

[CN3350299](https://www.tenders.gov.au/Cn/Show/03a3c53e-b3bd-eac1-558a-4c659e44a516) from AusTender has many amendments.

We can add sentence to make it clear that if this was published using only a single release in OCDS, then only the latest contract value and duration would be available to users, however if each amendment is published as a separate release then the full history of changes to the value and duration is available.

duncandewhurst:
Thanks Romina! Looks good - I added a couple of comments to the Australia example and commented on a couple of outstanding bits from the initial review. Once those are all resolved then I'm happy for James to do a final review.

duncandewhurst:
@romifz is this ready to review again? Thanks

romifz:
@duncandewhurst yes it is!

duncandewhurst (approved):

duncandewhurst:
Thanks Romina, a few comments/typos were unresolved so I've added commits to fix those.

Ready for @jpmckinney to review.

jpmckinney:
Okay to merge, but blocked by the presence of `<mark>add link</mark>` while waiting for #951.

PR #971, opened 2020-01-21
Remove the suggestion to use X prefixes

### LINKED ISSUE #872: Remove suggestion of using X prefixes
On schema/codelists, like:

> or by using a codes prefixed with 'x\\_' to indicate that it is a local 'eXtensions' to the codelist

https://tools.ietf.org/html/rfc6648 explains how HTTP, FTP, email, SIP, etc. have moved away from this pattern for the reasons in https://tools.ietf.org/html/rfc6648#appendix-B

jpmckinney:
… for the reasons expressed by the IETF. Closes #872. Putting changelog entry here to avoid conflicts:

* [#971](https://github.com/open-contracting/standard/pull/971) Codelists page: Remove the suggestion to use X prefixes, for the reasons expressed by the [IETF](https://tools.ietf.org/html/rfc6648#appendix-B).

### DIFF
--- a/docs/examples/planning.json
+++ b/docs/examples/planning.json
@@ -151,7 +151,6 @@
         "documents": [
           {
             "id": "0003",
-            "documentType": "x_consultationDocument",
             "title": "Consultation on cycle provision",
             "description": "A consultation document inviting citizen input into cycle provision.",
             "url": "http://example.com/consultations/cycle-provision/",
@@ -162,7 +161,6 @@
           },
           {
             "id": "0004",
-            "documentType": "x_map",
             "title": "Map of affected areas",
             "description": "A map showing areas affected by the planned highway updates. Available from local libraries.",
             "datePublished": "2015-02-15T00:00:00Z",
--- a/docs/examples/record.json
+++ b/docs/examples/record.json
@@ -332,7 +332,6 @@
               "dateModified": "2015-02-15T00:00:00Z",
               "datePublished": "2015-02-15T00:00:00Z",
               "description": "A consultation document inviting citizen input into cycle provision.",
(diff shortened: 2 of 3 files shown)


jpmckinney:
I'll let this one wait for review before merging into the 1.1.5-dev branch. It doesn't change any norms (the original text didn't use normative terms), but it's still a change in suggested behavior.

duncandewhurst:
Should the example files just show the new codes without the x_ prefix?

Should the documentation give any extra guidance for publishers on how to check if there is an existing additional code which they could use (e.g. by checking for issues in the Github repo?) or what to do if they add a new code?

The data feedback template includes the following guidance for publishers with additional codes:

> Where additional codes cannot be mapped to existing OCDS codes:
> 
> * Create an extension to document the title and description for the code.
> * Document the title and description of the code in the publication policy.

> Open a Github issue on the standard issue tracker so that the additional codes used in your data can be considered for inclusion in the core standard.

jpmckinney:
I'm worried that readers will see the codes and assume they belong to a codelist; my preference is for examples in OCDS to only use codes from codelists – unless the example is specifically to demonstrate the use of a new code that is not in a codelist.

For your second question, that's the subject of #972, which adds [this commit](https://github.com/open-contracting/standard/pull/972/commits/ded50c97f95b58dd66376c54fe61c5ad6d617988). This PR is just about eliminating the suggestion(s). #972 is about making the desired publisher behavior clearer. Can you review #972 and suggest any changes there?

With that clarification, is this PR ready to merge?

duncandewhurst (approved):
Good to merge, I hadn't seen #972

PR #992, opened 2020-03-23
Fix link to UNSC CPC codelist

Inova-MPRJ:


jpmckinney:
Thank you! I've added that commit to the 1.1.5-dev branch.

Issue #854, opened 2019-04-01
Clarify release.date field description


jpmckinney:
One publisher was updating the release date for all releases relating to a contracting process whenever anything changed. Other publishers have used it for the date when they generated the release, not when the process was made or first published in their local system. The current description is ambiguous, allowing for the above behaviors:

> The date this information was first released, or published

The field value should be the date on which the information contained in the release was first created or published in any system.

Related #848 and #849

jpmckinney:
I might also suggest a slight rewording: "first recorded in, or published by, any system".

PR #994, opened 2020-03-25
Update release.date description

yolile:
ref #854


jpmckinney:
@yolile Can you make a new branch starting from the 1.1.5-dev branch, and cherry-pick this commit onto it? I'm keeping all normative changes on that branch, so that 1.1-dev only includes non-normative changes until we're ready to release 1.1.5.

PR #995, opened 2020-03-25
Clarify the instruction for setting the release.date field

### LINKED ISSUE #854: Clarify release.date field description

yolile:
closes #854

### DIFF
--- a/docs/history/changelog.md
+++ b/docs/history/changelog.md
@@ -26,6 +26,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
 ### Schema
 
 * [#969](https://github.com/open-contracting/standard/pull/969) Clarify the instruction for setting the `Unit.id` field.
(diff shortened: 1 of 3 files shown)


jpmckinney:
I think you need to run `util/make_dereferenced_release_schema.py` whenever changing the release schema, otherwise tests will fail.

As it happens, it looks like a GitHub rate-limiting issue also caused the tests to fail (never saw that before).

yolile:
@jpmckinney done

jpmckinney:
Beautiful, tests pass (at least sometimes!).

jpmckinney:
I forgot to ask: Can you add a bullet point to the changelog under the Schema heading?

PR #996, opened 2020-03-25
Update documentation on linked release URLs

### LINKED ISSUE #844: Improve documentation on linked release URLs
record-package-schema.json:

> The URL of the release which contains the URL of the package with the releaseID appended using a fragment identifier e.g. http://standard.open-contracting.org/latest/en/examples/tender.json#ocds-213czf-000-00001

records_reference.md:

> In order to locate the specific release inside a release package the releaseID of the release should be appended to the package URL using a fragment identifier.

After #838, which clarified the set of rules about the uniqueness of release `id` values, the method above can match multiple releases. For this issue, it should be clarified that, from the perspective of a record package, the `ocid` is used to disambiguate matching releases.

We might also want to consider alternative methods of linking to releases to make the `url` values unambiguous; however, that doesn't need to be resolved as part of this issue.

yolile:
closes #844


jpmckinney:
Hmm, the old, surrounding text is not very clear. A short JSON snippet can demonstrate the overall structure of the record. There can be a better description of the two options (embedded releases and linked releases), and perhaps an explanation of their different benefits. The format of linked releases can be better described (perhaps with a short jsonschema directive of the three fields). A discussion of how linked releases refer to release packages can follow.

Do you want to update that entire section?

yolile:
@jpmckinney sure! I will update the record_reference.md file with your suggestions, but should I leave record-package-schema.json as it is now? (with this PR changes)

jpmckinney:
Let's update the Markdown first. It might give us some ideas for improving the schema description. Please also add a changelog entry (you can put it under the Schema heading, even though it also changes the Markdown).

yolile:
@jpmckinney maybe something like this?

jpmckinney:
Yes, looks great! Let me know when ready for review.

yolile:
@jpmckinney ready for review

jpmckinney:
@yolile Can you review my new changes this week? I'd like to merge and announce the 1.1.5 draft before I take vacation next week.

Link: https://standard.open-contracting.org/update-linked-releases-url-doc/en/schema/records_reference/

docs/schema/records_reference.md
```diff
@@ -28,81 +13,112 @@ The versioned release format is designed to support analysis of how data has cha
...
+The compiled release follows the [release schema](reference), with the exception of any fields on which `"omitWhenMerged": true` is declared.
 
-A versioned release contains the history of all the data in the compiled release, including which fields have changed, when they were changed, and the release which updated them.
+The process for creating a compiled release from individual releases is described in the [merging](merging) reference.
 
-This versioned information is valuable for many use cases relating to contract monitoring. However, it can add substantially to file sizes, and so where publishers do generate this at source, they can consider publishing versioned and non-versioned copies of the records.
+### Versioned release
 
-Third parties can use the information in the releases list to fetch source data, compile and verify their own version history for a contract.
+A versioned release contains the history of all the data in the compiled release, including which fields were changed, when they were changed, and the release in which they were changed.
```
yolile: maybe add "..including the fields first value, which fields were changed.." as the versionedRelease doesn't include only the fields that changed but all of them

docs/schema/records_reference.md
```diff
@@ -28,81 +13,112 @@ The versioned release format is designed to support analysis of how data has cha
...
    :title: releases
+```
+
+Above, the first linked release has a `url` value of <https://standard.open-contracting.org/examples/releases/ocds-213czf-000-00002-01-award1.json#ocds-213czf-000-00002-01-award1>. The first part (`https://standard.open-contracting.org/examples/releases/ocds-213czf-000-00002-01-award1.json`) is the URL of the release package, and the fragment identifier (`ocds-213czf-000-00002-01-award1`) is the `id` of the release.
+
+Release `id` values are only required to be unique within the scope of a contracting process: that is, within the scope of an `ocid` value. As such, a consuming application needs to use that fragment identifier in combination with the `ocid` of the record in order to identify the matching release within the release package.
+
+#### Embedded releases
+
+An embedded release follows the [release schema](reference).
```
yolile: @jpmckinney maybe add "So instead of just linking the release, here the whole releases are embedded" or similar, to describe more what an embedded release is

docs/schema/records_reference.md
```diff
@@ -28,81 +13,112 @@ The versioned release format is designed to support analysis of how data has cha
...
-The releases that make up a contracting process can be provided in two ways:
+Each release in a record can be provided as either a linked release or an embedded release.
 
-* URLs and metadata for each release; or
-* Embedded copies of the release;
+#### Linked releases
+
+A linked release follows a simple schema:
+
+```eval_rst
```
yolile, line 60: @jpmckinney the first row of the table is missing the title/description 
image

docs/schema/records_reference.md
```diff
@@ -28,81 +13,112 @@ The versioned release format is designed to support analysis of how data has cha
...
 
-Each record package contains an array of one or more records, consisting of the following sections:
+A record **must** contain an [ocid](../identifiers/#ocid) and all [releases](#releases) about the contracting process. As such, a record functions as an index of all releases about a contracting process.
+
+A record **should** contain a [compiledRelease](#compiled-release) object, which represents the state of the contracting process at the time of the record's publication.
 
-* Releases (required)
-* Compiled Release (recommended)
-* Versioned Release (optional)
+A record **may** contain a [versionedRelease](#versioned-release) object, which aggregates all values from all releases in a single object. The versioned release is designed to make it easy to see how values change from one release to another, and will often be generated by data users, rather than by publishers.
```
yolile: "all values of all fields"?

docs/schema/records_reference.md
jpmckinney: Good point. I'll use:

> In other words, instead of linking to a release within a release package as above, that release is entirely included in the record.

docs/schema/records_reference.md
jpmckinney: Thanks. I'll fix the JSON Pointer so that the table starts at the `url` row and omits the `/0`.

docs/schema/records_reference.md
jpmckinney: Will do.

docs/schema/records_reference.md
jpmckinney: I changed the paragraph to:

> A versioned release aggregates all values of all fields from all releases. For each field, it describes the history of values from the initial value to the current value, including the `id`, `date` and `tag` of the release in which it was changed.

jpmckinney:
@yolile I've added a commit with the changes. Ready for review :)

docs/schema/records_reference.md
yolile: perfect!

yolile:
@jpmckinney all good and clear! (I cannot approve as I'm the PR's author)

PR #952, opened 2020-01-02
Add personal identifiers example

yolile:
closes #801

Issue #883, opened 2019-06-20
Add a class for natural persons


jpmckinney:
## Problem

Presently, entries in the `parties` array are only `Organization` objects:

```
    "parties": { ...
      "items": {
        "$ref": "#/definitions/Organization"
      } ...
```

`Organization` has a `description` of "A party (organization)". In other words, all parties must be organizations, at present, according to the schema. Many of the extensions to this object add organization-specific fields that don't make sense for people.

**If any natural persons belong in the `parties` array, we will need to change the schema.**

**Sidebar: Exceptional case: Sole traders:** When governments award contracts to natural persons, they are often treated administratively as if they were corporations. For example, in Canada, natural persons with significant commercial activities must register a GST (VAT) ID the same as a corporation. This being the case, it's fine to use the `Organization` object for sole traders in OCDS.

## Background

#368 anticipated allowing natural persons as parties, but it's not clear that this was realized, beyond the special case of sole traders.

## Discussion

In terms of what belongs in the `parties` array, the field is described as "Information on the parties … who are involved in the contracting process and their roles." Going over the types of individuals that have come up:

* Legal representatives of suppliers and buyers: A legal representative is not a party to the contracting process. The supplier or buyer that the person represents is the party. A legal representative is just a person who can legally bind the organization.
* Owners, authorized agents, signing officers, contact points, etc.: These are similarly not parties to the process. They are people who are members of an organization who can act on behalf of that organization, within the legal limits of the rights/privileges accorded to their role/office.

it would not make sense to present such people alongside other parties; this would likely confuse users, especially given that an unaware tool won't be able to draw the link between the many people and their related organizations.

#621 discusses a local extension that would use the `parties` array for attendees of clarification meetings, including potential bidders (organizations) and public servants (natural persons). My interpretation is that the public servants are not parties, but are instead there on behalf of the buyer.

Other examples of natural persons are members of approval or award committees, etc. This is a more ambiguous scenario. However, I think it can still be said that, even in the temporary role of a juror in a design contest, that person is working on behalf of the buyer to determine the winner.

## Proposal

For the above natural persons, we can consider a new field (`people`) on the Organization object to collect all the relevant people and their respective roles within/with respect to that organization. In that case, unaware tools would display people within the appropriate context of their related organization.

As of writing, we don't have a use case for natural persons in the `parties` array (besides sole traders). If we do, however, we'll need to change the schema.

[...]

PR #1004, opened 2020-04-15
Add People in organizations worked example

romifz:


mrshll1001 (approved):
This is great. Concise and very clear. One thing I think we might need to expect is questions on how the org-ids for individuals are created / sourced and what advice we give to publishers about doing this and what to put in their publication policies. Something for a future update, though.

romifz:
Thanks @mrshll1001! Actually @yolile wrote a guidance on [personal identifiers](https://github.com/open-contracting/standard/pull/952) that answers some of these questions, I'll add a link.

jpmckinney:
I don't think we can include this as an example in the OCDS docs. Extensions should at minimum be registered in the Extension Explorer, after they have been reviewed. That ensures that the extension meets a minimum level of quality.

In this case, some issues with the Moldova extension are that `persons` should be `people`, and that we prefer fields with clear semantics like `contactPoint` rather than a generic list of people. One issue with the fictional extension is that we wouldn't recommend putting award committees inside a party, because in some jurisdictions like Taiwan award committees are independent of the party, and should be disclosed separately. For example, EU design contests have jury members, as modelled in the [design contest](https://extensions.open-contracting.org/en/extensions/designContest/master/schema/#DesignContest.juryMembers) extension.

If we want to add a worked example, we can describe which people can be modelled already using the `contactPoint` field and the `additionalContactPoint` extension (and maybe other registered extensions like design contest, though this should only be used for actual design contests, not for other procurement procedures). Otherwise, we need to make progress on #883.

PR #1035, opened 2020-06-09
Describe the normative status of non-required fields in the schema

### LINKED ISSUE #856: Describe the normative status of non-required fields in the schema
Currently the [building blocks](http://standard.open-contracting.org/latest/en/getting_started/building_blocks/) page in the getting started section states:

> The OCDS schema sets out the fields that should be included in each section...

To avoid overfill that statement will come with a caveat of "where applicable" in 1.1.4

In 1.1.5 we should consider how the statement can be featured more prominently (e.g. in the schema reference section) and explained more fully to help users understand the normative status of non-required fields in the schema.

jpmckinney:
Also, avoid incorrect use of "compile".

Closes #856

### DIFF
--- a/docs/getting_started/building_blocks.md
+++ b/docs/getting_started/building_blocks.md
@@ -50,7 +50,7 @@ These are represented in a JSON document as follows:
 
 ### Building blocks: fields
 
-The OCDS schema sets out the fields that ought to be included in each section (where applicable), making use of simple re-usable building blocks (field structures) to represent data. 
+The OCDS schema sets out fields that can be included in each section, making use of simple re-usable building blocks (field structures) to represent data. 
 
 For example, common building blocks are provided for:
 
--- a/docs/schema/reference.md
+++ b/docs/schema/reference.md
@@ -1,8 +1,12 @@
 # Release Reference
 
-The [Release Schema](release) provides a detailed specification of the fields and data structures to use when publishing contracting data. Supplementary schemas show how to combine releases into release packages and how to compile releases into records.
+The [Release Schema](release) provides a detailed specification of the fields and data structures to use when publishing contracting data. Supplementary schemas show how to combine releases into release packages and record packages.
 
 Releases are immutable – presenting information about a particular event in the lifetime of a contracting process. Publishers must not edit a release after publication; a new release can be created by changing the release's `id` and `date`.
+
+A few fields are required, to ensure the structural integrity of OCDS data. For example, `awards.id` is required so that references from `contracts.awardID` can resolve.
+
+The number of required fields is small, because OCDS standardizes the format in which contracting information is disclosed; it does not standardize the minimum amount of contracting information to disclose. Publishers choose what to disclose. That said, publishers are encouraged to populate as many fields as they can, without using placeholder values and without repeating the same value from an information source across multiple fields.
 
 **Note: If any conflicts are found between this text, and the text within the schema, the schema takes precedence.**
 


docs/schema/reference.md
```diff
@@ -1,9 +1,13 @@
...
+The [Release Schema](release) provides a detailed specification of the fields and data structures to use when publishing contracting data. Supplementary schemas show how to combine releases into release packages and record packages.
 
 Releases are immutable – presenting information about a particular event in the lifetime of a contracting process. Publishers must not edit a release after publication; a new release can be created by changing the release's `id` and `date`.
 
+A few fields are required, to ensure the structural integrity of OCDS data. For example, `awards.id` is required so that references from `contracts.awardID` can resolve.
+
+The number of required fields is small, because OCDS standardizes the format in which contracting information is disclosed; it does not standardize the minimum amount of contracting information to disclose. Publishers choose what to disclose. That said, publishers are encouraged to populate as many fields as they can, without using placeholder values and without repeating the same value from an information source across multiple fields.
```
duncandewhurst, line 9: My understanding is that OCDS describes what information publishers should disclose and how to format it. 

The number of required fields is small, because sometimes information isn't captured in contracting systems or disclosure isn't legislated for.

docs/getting_started/building_blocks.md
```diff
@@ -50,7 +50,7 @@ These are represented in a JSON document as follows:
 
 ### Building blocks: fields
 
-The OCDS schema sets out the fields that ought to be included in each section (where applicable), making use of simple re-usable building blocks (field structures) to represent data. 
+The OCDS schema sets out fields that can be included in each section, making use of simple re-usable building blocks (field structures) to represent data.
```
duncandewhurst, line 53: Per my other comment, I would keep 'ought to'

jpmckinney: “Ought to” in combination with “where applicable” is confusing... and as weak as a “can” on its own.

docs/schema/reference.md
jpmckinney: What publishers should disclose has never been part of normative content. All we had in 1.0 and 1.1 was BIA, which was non normative and is now replaced.

jpmckinney: Conformity is a separate concern than quality. More information is unambiguously higher quality, but it is no more conformant.

duncandewhurst: OK, so we have a different understanding of the normative status of non-required fields.

My understanding is that required fields must be published, non-required fields should be published and fields in extensions may be published. This is in-line with our guidance to publishers about using the mapping process to identify data which they do not collect but which OCDS recommends for disclosure

I thought BIA was a framework to help publishers towards eventually publishing all of the information that should be published, e.g. breaking down all of the 'should' fields into something more achievable.

My original intent in #856 was to move what appeared to be a normative statement (the schema describes the fields that should be included in each section) into the normative section of the documentation.

docs/getting_started/building_blocks.md
duncandewhurst: Good point, in that case I would remove 'where applicable' and keep it as 'ought to', especially since the original word was 'should'. 'where applicable' was added to try and address the issue of overfill, but that guidance/detail is probably better located in the mapping section of the implementation guidance.

docs/schema/reference.md
jpmckinney: The existence of a field in a schema does not imply that it should be published.

Extensions were not designed as a way to set the requirement level of fields, but for the reasons described in [this document](https://docs.google.com/document/d/1zvR1PDefO6yTK28uKA6XCnxMLiC9oiEeb3uFjHuRyqI/edit) and in [this section](https://docs.google.com/document/d/1j6Ec1vV0DklKMYvIBpeoIjABXDRT0nFythGNJR2ms24/edit?ts=5a95616c#heading=h.ca0qjk52gau) of another document. As stated there:

> Technically, commiting to implement OCDS implies no new disclosure requirements.

The document goes on to discuss how publishers nonetheless perceive OCDS as setting such requirements – an ambiguity that is described as desirable, since it allows the standard to be an instrument for advocacy, etc.

We cannot introduce new rules in a patch release. Making all fields explicitly "should" would be a new rule, as the standard development discussions above suggest. Doing so would give a firm answer to what was described as an ambiguity.

Our *guidance* to publishers follows from our advocacy and broader policy goals. However, such guidance isn't to be confused with a rule.

duncandewhurst: The first document also states that the core schema covers:

> [concepts] that from either a technical or policy perspective should be collected and published. For example, when OCDS 1.0 was published, the concepts under Implementation were rarely collected or published.

That seems to be a clear statement that the existence of a field in the schema implies that it should be published.

Explicitly declaring that non-required fields should be published doesn't change the fact that committing to implement OCDS implies no new disclosure requirements. It just makes their status as recommendations (not requirements) clear.

Pre 1.1.4, when there was no distinction between normative and non-normative sections of the documentation, the status of all fields was described as "should". We then changed that to 'ought to' in 1.1.4 but failed to add the statement to the normative section of the documentation.

If ambiguity about whether non-required fields are recommended or optional is preferred, then I don't think we should add a statement that basically explicitly states that they are optional.

jpmckinney: The documents are a description of earlier authors' thinking, to assist later authors in weighing any revisions to their earlier decisions – that is why I sat down with Ben, David and Tim to document their reasoning.

In that context, the meaning of that paragraph is along the lines of: "The core schema <ins>was designed to</ins> cover<del>s</del> concepts that<ins>,</ins> from either a technical or policy perspective<ins>, its authors thought</ins> should be collected and published." (Such qualifications would have needed to be present in every sentence, and were therefore omitted.)

Similarly, the ambiguity was described as desirable as part of explaining the choices made by earlier authors. (However, in a standard, a lack of clarity about rules isn't desirable, and ambiguities should be resolved. If asked, we can't be wishy-washy.)

The intention of the documents was not to document undocumented rules of OCDS… And the majority of statements in both documents suggest a non-normative status (the statements I already cited among them).

With respect to the quoted sentence, MUST, SHOULD and MAY are all requirement levels. That is the meaning of 'requirement' in "no new disclosure requirements."

The normative review done in 1.1.4 was all about changing words and moving rules as appropriate: to decide whether a "must" is actually a "need" and vice versa, etc. Original wording was treated as indicative, but not prescriptive, during that review.

What to publish has always been, and continues to be, part of OCP's advocacy work. That is the strongest lever for increased disclosure.

Not to start a tangent, but I am also concerned whether having all fields be recommended is even a coherent statement. There are many qualifications to add for it to make sense, all of which – put together – lead to a much weaker statement, effectively the same as in this PR. For example, "recommended unless": not known at the time, not applicable to the procedure, not applicable in the jurisdiction, disallowed by law, not tracked in any source systems, too unstructured to map to the structures in OCDS, in violation of GDPR in the case of sole traders, etc. [EU legislation](https://eur-lex.europa.eu/legal-content/EN/TXT/HTML/?uri=CELEX:32014L0024), for example, has a list of reasons for withholding publication, all of which are at the discretion of the publisher:

> Certain information on the contract award or the conclusion of the framework agreement may be withheld from publication where its release would impede law enforcement or otherwise be contrary to the public interest, would harm the legitimate commercial interests of a particular economic operator, public or private, or might prejudice fair competition between economic operators.

jpmckinney: Put more shortly, norms need to be relevant to conformance. Would we consider a publication as more or less conformant based on an analysis of what fields could have been published? Is there an unambiguous, fair process for doing that analysis?

In other discussions, we made this a concern of quality (MEL2) not of recognition as a publisher (MEL1).

jpmckinney: I am also concerned that such a rule would constrain standard development, since any proposed field would thereafter have to pass the test of "is this a field that SHOULD be published?" (a design question) as opposed to simply passing the test of "is the definition and modelling of this field agreed by consensus?" (a modelling question).

The choice of monolithic versus modular schema is a question of design – not a question of rules. A modular schema was chosen to address certain design challenges (overwhelm, etc.), but we should have the freedom to reconsider the best solution to those challenges at any time, since the choice of design doesn't change any rules – unless we add this rule.

LindseyAM:
The schema reference section is already clear on the technical rules of the OCDS eg which rules must be followed.  

It would be confusing to add aspirational language into this section about how much data to publish eg what the coverage of a publication must or should be.

Instead, recommendations and guidance about how much to publish are stated and explained in the non-normative or Guidance sections of the documentation.

PR #1045, opened 2020-07-20
Update useful publication measure

duncandewhurst:
Add 'tender title' to answer the 'what' part of 'who buys what from who, when and for how much'.


jpmckinney (approved):
I approve. @LindseyAM As author of the page, do you approve?

PR #1065, opened 2020-09-11
Avoid ambiguity as of the duration of 'contracts' (wrong branch)

### LINKED ISSUE #902: Expand tender.contractPeriod description
> The period over which the contract is estimated or required to be active. If the tender does not specify explicit dates, the duration field may be used.

The EU standard procurement forms use the construction "Duration of the contract, framework agreement or dynamic purchasing system", to avoid ambiguity around which contract this period refers to (as there can be many contracts over different periods, especially in the latter cases).

ColinMaudry:
Closes #902


ColinMaudry:
Replaced with https://github.com/open-contracting/standard/pull/1114

PR #1066, opened 2020-09-11
Recommendation to put hyphen after OCID prefix (wrong branch)

### LINKED ISSUE #917: Add recommendation to put hyphen after OCID prefix
This is the practice in most implementations (and all examples), but it is not explicit. We can't make it a MUST rule, because that will be backwards-incompatible with at least Zambia.

ColinMaudry:
Closes #917

### DIFF
--- a/docs/schema/identifiers.md
+++ b/docs/schema/identifiers.md
@@ -83,9 +83,11 @@ Mexico City then registered a prefix with the OCDS helpdesk. They have been give
 
 </div>
 
-The ocid prefix itself is made up of two parts: a prefix agency identifier (currently only 'ocds' is used), and a random six-character alphanumeric string generated for each publisher of data. 
+The ocid prefix itself is made up of two parts: a prefix agency identifier (currently only 'ocds' is used), and a random six-character alphanumeric string generated for each publisher of data.
 
 The ocid is case-sensitive; in other words, the letter case of an ocid must be consistent.
+
+It is good practice to separate the ocds prefix and the internal identifier with and hyphen (`-`).
 
 ### Registered prefixes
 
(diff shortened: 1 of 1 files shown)


docs/schema/identifiers.md
```diff
@@ -83,10 +83,12 @@ Mexico City then registered a prefix with the OCDS helpdesk. They have been give
 
 </div>
 
-The ocid prefix itself is made up of two parts: a prefix agency identifier (currently only 'ocds' is used), and a random six-character alphanumeric string generated for each publisher of data. 
+The ocid prefix itself is made up of two parts: a prefix agency identifier (currently only 'ocds' is used), and a random six-character alphanumeric string generated for each publisher of data.
 
 The ocid is case-sensitive; in other words, the letter case of an ocid must be consistent.
 
+It is good practice to separate the ocds prefix and the internal identifier with and hyphen (`-`).
```
```suggestion
It is encouraged to separate the ocds prefix and the internal identifier with a hyphen (`-`).
```

We like to use the words in this table: https://tools.ietf.org/html/draft-hansen-nonkeywords-non2119-04#page-3

jpmckinney (commented):
Please also update the schema. We can't change the validation rules in 1.2, so we're using the non-normative word "encourage". However, the intention is to change the rule in 2.0, so it's better to have this in the schema as well.

ColinMaudry:
@jpmckinney The CI fails with this error:

> Error: Input required and not supplied: key

Do you know why?

ColinMaudry:
Shall I close this PR (and the others that fail) and create it from a branch that branches from `open-contracting/standard:1.2-dev`?

jpmckinney:
Aha, the secret key needed for the deploy-to-staging step is not accessible, because this PR is from a fork, instead of from just a branch of this repo. If you change your git remote origin to point to open-contracting/standard, you should be able to push your branch and create a new PR.

ColinMaudry:
Will redo this PR branching from open-contracting/standard instead of my fork.

ColinMaudry:
Replaced this PR with https://github.com/open-contracting/standard/pull/1113

PR #1086, opened 2020-10-12
String identifiers preference

### LINKED ISSUE #789: Guidance: Preference for string over integer identifiers
The schema allows both integers and strings for some `id` fields, and only strings for others. I believe this is because, during the supply-side research, it was found that some publishers used integers, and others used strings. However, this inconsistency can cause problems:

1. A publisher might have integers in their database, and therefore use integers for `id` fields, and then eventually encounter a structural error when they try to use an integer for an `id` field that only allows strings. This inconsistency could be surprising and frustrating.
1. A data user might look at data and see integers, and assume that all `id` values are integers. Their code or analysis then runs into trouble when it encounters string values. They then add extra code to either handle the inconsistency, or coerce all `id` values to strings.
1. A publisher might inconsistently use integers and strings for the same field. When making comparisons to check if two objects are the same, users will need to coerce both `id` values to string.

We can't remove integer as a possible value in 1.X, but we can strongly recommend the use of strings.

yolile:
Closes #789


duncandewhurst (changes requested):
Changes look good.

Please can you update the changelog?

yolile:
done!

docs/history/changelog.md
```diff
@@ -30,6 +30,12 @@ Per the [normative and non-normative content and changes policy](https://docs.go
...
 
+## [1.2] - YYYY-MM-DD
+
+### Schema
+
+* [#1086](https://github.com/open-contracting/standard/pull/1086) Recommend the use of string over integer for identifiers
```
```suggestion
* [#1086](https://github.com/open-contracting/standard/pull/1086) Recommend the use of strings rather than integers for identifiers.
```

jpmckinney (approved):

Issue #1009, opened 2020-05-18
Implementation guidance outstanding issues


duncandewhurst:
Outstanding issues from #986

**General:**

- [x] Add a 'publication plan' sub-heading to the design stage and link to that from the other stages, rather than linking to the template repeatedly.

**Design phase:**

- [ ] (Set your goals and priorities) Add link to [Guide to User Research](https://drive.google.com/file/d/10-4rTGmpXk4oRoyg7ccuv4Jga_6B_lpI/view) once document sharing settings updated.
- [x] Publish [Guide to OCDS Procurement Indicators](https://docs.google.com/document/d/1vSJk9-qWSTQEx9ZZc7BUhQZMHvTRcyDYVS2sl8HB__k/edit?usp=sharing) in resources section of OCP website and update link.
- [x] Publish [Spreadsheet mapping common objectives and indicators to OCDS](https://docs.google.com/spreadsheets/d/1nG7e52E1CXOXoUjz6pimW4Z7er9u3DJSs98QKdJJioE/edit#gid=110864222) in resources secction of OCP website and update link.
- [x] Re-order content in 'set your goals and priorities' section for better flow, consistent language and to link actions to supporting resources (identify stakeholders, then set goals, then determine data needs).

**Build phase:**

- [x] Publish [spreadsheet input template](https://docs.google.com/spreadsheets/d/1Py2Ki88TeerEToK8dIWcyU8rF5p6-jXhD7tx1W92m2g/edit#gid=956894726) as a resource on the OCP website and update link.
- [x] Add link to [technical case studies report](https://www.open-contracting.org/resources/technical-case-studies-ocds-implementation-insights/)
- [x] Publish remainder of series on prototyping OCDS data using spreadsheets (currently linked from the guidance, but only the first blog is published)
- [ ] Add link to packaging worked example under 'establish your publication formats and access methods', once ready.
- [x] Reorganize 'build your data pipeline' section into low tech capacity/infrastructure and high tech capacity/infrastructure.
- [x] Discuss required updates to 'Establish your publication formats and access methods' section and sub-pages, based on notes from OCDS retreat:
  * Put more emphasis in https://standard.open-contracting.org/latest/en/implementation/hosting/ ?
  * Revisit the guidance on bulk downloads and API filters
    1. We should strongly encourage the availability of bulk downloads in CSV format.
    2. We should encourage bulk downloads in JSON format, to avoid having to scrape the API to get the full dataset.
[...]

PR #1091, opened 2020-10-20
1009 various

odscrachel:
A few commits from issues in #1009 
* Emphasis on importance of bulk downloads 
* 'Re-order content in 'set your goals and priorities' section for better flow, consistent language and to link actions to supporting resources (identify stakeholders, then set goals, then determine data needs).'
* Further discussion is needed regarding 'Reorganize 'build your data pipeline' section into low tech capacity/infrastructure and high tech capacity/infrastructure.' For now, have added mention that we have tools for low and high tech environments.

### DIFF
--- a/docs/guidance/build/hosting.md
+++ b/docs/guidance/build/hosting.md
@@ -1,6 +1,6 @@
 # Data files and APIs
 
-Different users have different needs when it comes to accessing OCDS data. 
+Different users have different needs when it comes to accessing OCDS data. It also must be possible to automate the download of all the OCDS data, in some form.
 
 [Data on the Web Best Practices](https://www.w3.org/TR/dwbp/#MultipleFormats) suggests that "Data should be available in multiple data formats" in order to increases the number of different users, tools and applications that can process the data.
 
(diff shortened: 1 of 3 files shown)


docs/guidance/build/hosting.md
```diff
@@ -1,6 +1,6 @@
 # Data files and APIs
 
-Different users have different needs when it comes to accessing OCDS data. 
+Different users have different needs when it comes to accessing OCDS data. It also must be possible to automate the download of all the OCDS data, in some form.
```
jpmckinney, line 3: See this bullet from https://ocds-standard-development-handbook.readthedocs.io/en/latest/standard/contributing.html

> Never use normative words on guidance pages. Use [non-normative synonyms](https://tools.ietf.org/html/draft-hansen-nonkeywords-non2119-04#page-3) instead.

```suggestion
Different users have different needs when it comes to accessing OCDS data. It also needs to be possible to automate the download of all the OCDS data, in some form.
```

jpmckinney (approved):
Just one small fix!

PR #1094, opened 2020-10-22
1060 parties/id uniqueness

### LINKED ISSUE #1060: Clarify uniqueness of `parties/id` and document guidance on publishing data without organization ids
Niether the schema nor the [identifiers reference](https://standard.open-contracting.org/latest/en/schema/identifiers/#identifiers) are clear on the unqiueness of `parties/id` and by extension `buyer/id`, `awards/suppliers/id` etc. My understanding is that these fields are local identifiers, unique within the scope of the `parties` array / contracting process.

When a data source only contains the name of an organization, publishers often get confused about how to populate `parties/id` and try to put a unique value in there. In practice, if the only identifier for an organization is its name, it would be better just to use a fixed or sequential identifier in `parties/id` so that users are not misled into thinking it can be relied on as an organization identifier.

We could:

1. Update the description of `parties/id`, e.g. 

> The ID used for cross-referencing to this party from other sections of the release. This field **need only be unique within the scope of the contracting process, but** may be built with the following structure {identifier.scheme}-{identifier.id}(-{department-identifier}) **where an organization identifier is available.**

2. Add a section to the [Identifiers reference page](https://standard.open-contracting.org/latest/en/schema/identifiers/#identifiers) to clarify the purpose and unqiueness constraints of `parties/id` and `OrganizationReference/id`.

3. Add a page to the guidance describing how to populate `parties/id` and `OrganizationReference/id` when your data source only includes names for organizations.

duncandewhurst:
Closes #1060


duncandewhurst:
@jpmckinney should `1.2-dev` be protected? It looks like I can merge this PR without an approving review.

docs/guidance/map/organization_identifiers.md
```diff
@@ -17,3 +17,9 @@ The Government of UK uses identifiers from the UK Companies House to uniquely id
...
+
+## Party IDs
+
+Each of the parties (organizations or other participants) in the [parties section](../../../schema/reference/#parties) ought to have a [local party ID](../../../schema/identifiers/#party-ids) (`id`) used to reference it from elsewhere in the schema. For parties with an organization identifier, you can construct `id` using `{identifier.scheme}-{identifier.id}`.
+
+For parties without an organization identifier, you can populate `id` with a fixed or sequential value. For example, you can set the buyer's `id` to '1' and set each supplier's `id` sequentially from '2' onwards. Alternatively, you can set the party's `id` to its `role` and add a sequential number for roles with multiple parties, e.g. 'buyer', 'tenderer-1', 'tenderer-2' etc.
```
yolile: @duncandewhurst I don't know if it would be useful to add that that id should remain the same across all the releases of the same contracting process. Or would it be redundant/too much?

yolile (approved):

jpmckinney:
@duncandewhurst The branch is protected, but PR reviews aren't set as being required. This is so that I can both (1) enforce rules on admins like myself, but (2) not require a review when I'm just merging some changes to the build process. So far, an 'honour system' of waiting for reviews is working well enough.

docs/guidance/map/organization_identifiers.md
```diff
@@ -17,3 +17,9 @@ The Government of UK uses identifiers from the UK Companies House to uniquely id
 ```
 
 The publisher also collects two extra identifiers, which are disclosed in the `additionalIdentifiers` block. The first one is the VAT identification number for suppliers. Note that the VAT registry is not present in org-id.guide, but the publisher followed the instructions in the [org-id meta-data guide](http://docs.org-id.guide/en/latest/metadata/#assigning-a-code) to build the "GB-VAT" code used in the `scheme` field: the two-letter country prefix ("GB") plus a short abbreviation for the registry ("VAT"). The publisher checked that it does not conflict with any list code in org-id.guide.
+
+## Party IDs
+
+Each of the parties (organizations or other participants) in the [parties section](../../../schema/reference/#parties) ought to have a [local party ID](../../../schema/identifiers/#party-ids) (`id`) used to reference it from elsewhere in the schema. For parties with an organization identifier, you can construct `id` using `{identifier.scheme}-{identifier.id}`.
```
jpmckinney:
```suggestion
Each of the parties (organizations or other participants) in the [parties section](../../../schema/reference/#parties) ought to have a [local party ID](../../../schema/identifiers/#party-ids) (`id`) used to reference it from elsewhere in the data. For parties with an organization identifier, you can construct `id` using `{identifier.scheme}-{identifier.id}`.
```

docs/guidance/map/organization_identifiers.md
jpmckinney: I think it's relevant to add a sentence, so that publisher don't just re-number parties in each release. Here's what we wrote for the OCDS for EU profile, for reference: https://standard.open-contracting.org/profiles/eu/master/en/operations/#add-a-party

docs/schema/identifiers.md
```diff
@@ -156,6 +156,16 @@ See the [full worked example](../guidance/map/organization_identifiers) for more
 
 If you want to disclose identifiers for natural persons, see the [personal identifiers guidance](../guidance/map/organization_personal_identifiers).
 
+## Party IDs
+
+Each of the parties (organizations or other participants) involved in a contracting process is declared in the [parties section](../reference/#parties).
+
+Each party has a local identifier (`id`) used to reference it from elsewhere in the schema. For example, `buyer/id` references the buyer's entry in the parties section using `parties/id`.
```
```suggestion
Each party has a local identifier (`id`) used to reference it from elsewhere in the data. For example, `buyer/id` references the buyer's entry in the parties section using `parties/id`.
```

jpmckinney (commented):
Looks good, just a few tweaks ('schema' -> 'data').

docs/guidance/map/organization_identifiers.md
duncandewhurst: Good idea. In https://github.com/open-contracting/standard/pull/1094/commits/85080db6ece72d7045c49b5877a6f89a4dd54b75 I've added a normative statement on consistency in the identifiers reference (similar to that for Award and Contract IDs) and an example in the guidance

jpmckinney (approved):

PR #1106, opened 2020-11-02
add frameworks guidance

duncandewhurst:


duncandewhurst:
@jpmckinney this build is failing with an exception in sphinxcontrib-opendataservices but I don't see any recent changes in that repo. Any idea what could be causing it?

> Exception occurred:
>   File "/home/runner/work/standard/src/sphinxcontrib-opendataservices/sphinxcontrib/opendataservices.py", line 75, in run
>     pointed = resolve_pointer(json_obj, self.options['jsonpointer'])
> KeyError: 'jsonpointer'

jpmckinney:
`:jsonpointer:` is required for `jsoninclude:` even if it's just empty.

duncandewhurst:
Hmmm, seems like there's an issue with 1.1-dev which this is branched from. I'll branch from 1.1 instead.

duncandewhurst:
Replaced by https://github.com/open-contracting/standard/pull/1108

jpmckinney:
> Hmmm, seems like there's an issue with 1.1-dev which this is branched from. I'll branch from 1.1 instead.

What is the issue? 1.1-dev builds without error: https://github.com/open-contracting/standard/branches

jpmckinney:
Also, this issue was ready for review but #1108 is not? Just want to check the status.

duncandewhurst:
On 1.1-dev the [map guidance](https://standard.open-contracting.org/staging/1.1-dev/en/guidance/map/#map-your-data-to-ocds) is broken and doesn't build the mapping the hard cases pages:

image

The build in #1108 is now failing, so I wanted to resolve that before requesting a review.

jpmckinney:
Please create a PR to add the missing closing triple backticks to the eval_rst block, which is causing the issue on 1.1. Please also only branch from 1.1-dev.

duncandewhurst:
Missing backticks fixed in #1111. New PR branching from 1.1-dev in #1123

PR #1108, opened 2020-11-03
Frameworks guidance 1.1

duncandewhurst:


jpmckinney:
Looks like `related_processes` is linking to `schema/codelists`, but not with enough `../`

duncandewhurst:
Replaced by #1123

PR #1112, opened 2020-11-11
 Strengthen some normative statements in schema

### LINKED ISSUE #868: Strengthen some normative statements in schema
> The ID used for cross-referencing to this party from other sections of the release. This field **may** be built with the following structure {identifier.scheme}-{identifier.id}(-{department-identifier}).

MAY -> SHOULD

> The maximum duration of this period in days. A user interface can collect or display this data in months or years as appropriate, and then convert it into days when storing this field. This field can be used when exact dates are not known. If a startDate and endDate are set, this field, if used, **should** be equal to the difference between startDate and endDate. Otherwise, if a startDate and maxExtentDate are set, this field, if used, **should** be equal to the difference between startDate and maxExtentDate.

SHOULD -> MUST

> The goods, services, and any intangible outcomes in this contract. Note: If the items are the same as the award do not repeat.

Change second clause to either SHOULD or MUST statement.

Follow-up to #855

ColinMaudry:
Closes #868


schema/release-schema.json
```diff
@@ -685,7 +685,7 @@
         },
         "items": {
           "title": "Items contracted",
-          "description": "The goods, services, and any intangible outcomes in this contract. Note: If the items are the same as the award do not repeat.",
+          "description": "The goods, services, and any intangible outcomes in this contract. Note: you should not repeat the items if they are the same as the award.",
```
jpmckinney: Write from a neutral perspective, rather than a publisher perspective. Here, I suggest using the passive rather than the active voice:

```suggestion
          "description": "The goods, services, and any intangible outcomes in this contract. If the items contracted are identical to the items awarded, this field should be omitted.",
```

jpmckinney (commented):
I think we're missing this part of the issue description:

> The ID used for cross-referencing to this party from other sections of the release. This field **may** be built with the following structure {identifier.scheme}-{identifier.id}(-{department-identifier}).

MAY -> SHOULD

docs/history/changelog.md
```diff
@@ -36,6 +36,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
 
 * [#1086](https://github.com/open-contracting/standard/pull/1086) Recommend the use of strings rather than integers for identifiers.
 * [#1094](https://github.com/open-contracting/standard/pull/1094) Clarify the uniqueness of `Organization.id`
+* [#1112](https://github.com/open-contracting/standard/pull/1112) Strengthen some normative statements in schema
```
jpmckinney: Can you explicitly mention the three changes, so a reader of the changelog doesn't need to open the PR/issue?

ColinMaudry: Done

Issue #896, opened 2019-07-25
Clarify Contract description


jpmckinney:
# Background

`contracts`:

> Information from the contract creation phase of the procurement process.

`Contract`:

> Information regarding the signed contract between the buyer and supplier(s).

The eProcurement Ontology (ePO) has a class for Contract:

> A voluntary, deliberate, and legally binding agreement between two or more competent parties.

The ePO has a class for Purchase Contract, which is [synonymous](https://github.com/eprocurementontology/eprocurementontology/issues/170) with: specific contract, call-off contract, draw-down contract (Ireland) and contract under a framework agreement. This concept also applies to "dynamic purchasing systems", which UNCITRAL considers synonymous with "open framework agreement".

# Discussion

None of these definitions clearly indicate which contracts between buyer and suppliers belong in the `contracts` array, which can include:

* A contract establishing the relationship, like the set-up of a framework agreement
* A contract within the relationship, like a call-off contract under a framework agreement
* Purchase orders ([the acceptance of a purchase order forms a legal contract](https://en.wikipedia.org/wiki/Purchase_order))
* Catalog purchases

## Framework agreements

The most important difference between the set-up contract and call-off contracts is that no goods or services are delivered directly under the set-up contract. The framework agreement has a maximum value, but no value in the same sense as a call-off contract. It might specify items to be delivered by call-off contracts, but these are not delivered directly under the set-up contract and might not have specific quantities.

The [present guidance](http://standard.open-contracting.org/latest/en/implementation/related_processes/) puts call-off contracts in the `contracts` array, and leaves the set-up contract in the `awards` array. This avoids confusion of the differences above, but it isn't perfect, as the set-up contract is indeed a contract itself. Some facts about the set-up contract might be better understood in the context of a contract than in the context of an award decision.

TED and eForms can inform this discussion.

## Purchase orders

In a contract with definite quantities of items, purchase orders are more of an implementation detail, describing the specific timing, quantities, deliveries, etc. of the items in the contract.

In this first case, purchase orders seem appropriate under `contracts.implementation`.

In a contract with indefinite quantities, and especially in a contract with multiple suppliers (e.g 'simultaneous supply' in Paraguay #888), a user will have to look at the purchase orders to determine the actual quantities delivered by which suppliers, and to determine the real value of the contract. This is closer, in form, to a framework agreement with call-off contracts.

In this second case, it seems, at first, more appropriate to follow the same model as framework agreements.

[...]

Issue #827, opened 2019-03-22
Check schema field and codelist code descriptions against OCDS, ePO, UNCITRAL glossaries


jpmckinney:
Same as #826 but for OCDS 1.2 due to versioning policy.

* The EU [e-Procurement Ontology](https://eprocurementontology.github.io) working group has a list of [glossary terms](https://github.com/eprocurementontology/eprocurementontology/wiki/eProcurement-Glossary) that link to [this spreadsheet](https://docs.google.com/spreadsheets/d/1zw9aR8GDIDUiTDtSznMxDlZQEAGb8uNzib9KBZLf5yE/edit#gid=0). I haven't checked if the ontology itself re-uses the glossary definitions verbatim.
* [UNCITRAL glossary](http://www.uncitral.org/pdf/english/texts/procurem/ml-procurement-2011/Glossary-e.pdf)
* [EBRD glossary](https://www.ppi-ebrd-uncitral.com/index.php/en/procurement-glossary)
* [eForms](https://ec.europa.eu/growth/single-market/public-procurement/digital/eforms_en)
* [OCDS glossary](https://docs.google.com/spreadsheets/d/1WGH9_mHYuF4JbK2tdyeckqsmj8v4HrRqDOEbKQ7CI4A/edit#gid=0)

jpmckinney:
Copying comment from https://github.com/open-contracting/standard/issues/380#issuecomment-519951704:

UNCITRAL has a term matching the semantics of 'direct':

> **Single-source procurement**: A method of procurement of last resort which main distinct feature is the absence of competition since the invitation to present a quotation or proposal is addressed only to one supplier or contractor.

jpmckinney:
http://www.mapsinitiative.org/methodology/MAPS-glossary.pdf

ColinMaudry:
We could work toward having a taxonomy, connecting concepts according to various properties (equivalent to, similar to, related, broader, narrower), using the [SKOS ontology](https://www.w3.org/TR/skos-reference/) (W3C recommendation) and publishing it using Skosmos([website](http://skosmos.org/), [Github](https://github.com/NatLibFi/Skosmos), [example of taxonomy](http://finto.fi/koko/en/page/p54285)).

ColinMaudry:
Working document: https://docs.google.com/document/d/1lXRWIOjHjhK5KvVrfuC1MBAMl1JvoV9zBCR7Lt5UcUM/edit

ColinMaudry:
[...]

PR #1114, opened 2020-11-11
Avoid ambiguity as of the duration of 'contracts'

### LINKED ISSUE #902: Expand tender.contractPeriod description

ColinMaudry:
Closes #902


ColinMaudry:
@jpmckinney Do you know what's wrong here?

I ran `python util/make_dereferenced_release_schema.py` and merged the 1.2-dev branch in this one.

jpmckinney:
I think the search backend was under high load due to the many PRs, so I'll just re-run the checks.

schema/release-schema.json
```diff
@@ -389,7 +389,7 @@
           "$ref": "#/definitions/Period"
         },
         "contractPeriod": {
-          "description": "The period over which the contract is estimated or required to be active. If the tender does not specify explicit dates, the duration field may be used.",
+          "description": "The period over which the contract, the framework agreement or the dynamic purchasing system is estimated or required to be active. If the tender does not specify explicit dates, the duration field may be used.",
```
```suggestion
          "description": "The period over which the contract, framework agreement or dynamic purchasing system is estimated or required to be active. If the tender does not specify explicit dates, the duration field may be used.",
```

jpmckinney:
Closing in favor of the approach described in https://github.com/open-contracting/standard/issues/827#issuecomment-728036400, whereby all mentions of "contract" would defer to #896, rather than repeating the long statement everywhere.

Issue #1110, opened 2020-11-05
Items: Add more detailed fields


yolile:
from discussion with @micayael from DNCP Paraguay

In #751 a dynamic "attributes" array was added to the item object as a local extension. This allowed the DNCP to disclose specific information about their items, such as product packaging, brand, and for some kind of products, as computers their specific characteristics as RAM Memory, etc.

Although they are including all the possible attributes of the items here, they consider that there are some of them that can be included in the item object itself, as those are always in all the items no matter what the item is. 

Specifically, they consider that the product packaging and unit of measure of the packaging are key fields to include in order to be able to compare and analyses items' unit prices. 

Another key field to include is the product packaging and unit of measure with which the tenderers can present their bids, this can be more than one. (for example, the tenderer only sells pencils in boxes of 100 units therefore he can only make his offer by boxes and not by unit)

The final use case is being able to compare and analyze items' unit prices in the most accurate way

jpmckinney:
Sounds good to me! Related to #1050. Could you share draft JSON Schema or example JSON data?

yolile:
For including the packaging and its unit (maybe the same can be applied to medicines form/dosage? from #1050):
(I'm not sure about the "packaging" name, but I couldn't find any existing schemas describing the package/form of products eg http://www.heppnetz.de/ontologies/goodrelations/v1#SomeItems , https://schema.org/TypeAndQuantityNode,  https://schema.org/Product, https://schema.org/Demand)

### Schema

```json
{
    "definitions": {
        "Item": {
            "properties": {
                "packaging": {
                    "title": "Packaging",
                    "description": "A description of the package in which the supplies are provided (e.g. box, bag) and its relation with the unit of measure.",
                    "type": "object",
                    "properties": {
                        "name": {
                            "title": "Name",
                            "description": "Name of the package.",
                            "type": [
                                "string",
                                "null"
                            ]
                        },
                        "unitsPerPackage": {
                            "title": "Unit",
                            "description": "A description of the unit in which the supplies package is provided and its quantity per package.",
                            "type": "object",
                            "properties": {
                                "scheme": {
                                    "title": "Scheme",
                                    "description": "The list from which identifiers for units of measure are taken, using the open [unitClassificationScheme](https://standard.open-contracting.org/1.1/en/schema/codelists/#unit-classification-scheme) codelist. 'UNCEFACT' is recommended.",
                                    "type": [
                                        "string",
                                        "null"
                                    ],
[...]

Issue #1107, opened 2020-11-02
Move unit to definitions


duncandewhurst:
`unit` is currently defined as a property of `Item`.

The [metrics extension](https://extensions.open-contracting.org/en/extensions/metrics/master/) repeats the definition and [Moldova's eOCDS-eAuction](https://github.com/eOCDS-Extensions/eOCDS-eAuction) extension will do the same.

It would be better to have a definition for `unit` which can be referenced by extensions.

**Proposal**

1. Move the definition of `unit` to the definitions section of the schema.
2. Update the metrics extension to reference the definition, rather than repeat it.

cc @PaulBoroday

jpmckinney:
Agreed. This is a simple refactor of the schema.

jpmckinney:
@yolile for the metrics extension, we can rename master to 1.1 (and update the registry), and then add a 1.2 branch that references the Unit definition.

yolile:
Closing in favor of https://github.com/open-contracting/ocds-extensions/issues/179 as the part related to the schema was done in #1240

Issue #858, opened 2019-04-02
Add ISO639-1 language codelist


jpmckinney:
The language fields allow BCP47 language tags, but we can at least have an open codelist containing the ISO639-1 language codes (which are the recommended values). We would need a script like the one for the currency codelist.

http://www.loc.gov/standards/iso639-2/php/code_list.php

duncandewhurst:
+1

I came here via some publisher data which used 'english' in the language tag and wasn't flagged by the validator

akuckartz:
Linked Data can be used. See http://id.loc.gov/vocabulary/iso639-1.html

duncandewhurst:
It seems like the TSV file would be the easiest to integrate: http://id.loc.gov/vocabulary/iso639-1.tsv

Issue #857, opened 2019-04-02
Add IANA Media Types codelist


jpmckinney:
`Document.format` refers to the codelist, but the schema doesn't specify it. We would need a script like the one for the currency codelist.

http://www.iana.org/assignments/media-types/

jpmckinney:
This would have to start as an open codelist in 1.2.0, then switch to a closed codelist in 2.0.0.

duncandewhurst:
I've written a simple script, similar to `fetch_currency_codelist.py`, to fetch the media types from the CSV files on https://www.iana.org/assignments/media-types/media-types.

There's a couple of issues to address before making a PR:

**Missing values**

`Document/format` specifies the values in the IANA list's 'Template' column, which consist of the type and subtype (e.g. application/pdf), but some media types do not have a value in the template column. For example:

![image](https://user-images.githubusercontent.com/19265814/121975113-fb6e4200-cdd4-11eb-8dc6-0074eebb66c0.png)

I searched the various RFCs linked from the reference and notes sections of https://www.iana.org/assignments/media-types/media-types.xml, but I couldn't find an explanation for the missing template types.

However, the missing type/subtype codes (e.g. image/gif, image/jpeg) seem to be in common usage (e.g. the [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types)), so it seems like they might simply be omissions from the IANA page. In which case, should I update the script to generate necessary values by combining the type and the subtype?

**Deprecated/Obsolete types**

Deprecated and obsolete types are noted in the 'Name' column. For example:

![image](https://user-images.githubusercontent.com/19265814/121975276-59028e80-cdd5-11eb-8047-c36801f8299c.png)

With one exception (vnd.geo+json), it should be possible to extract the notes into a separate column by splitting the Name column on ' - '.

Should I update the script to do that?

jpmckinney:
Yes, please update the script in those ways.

FYI, all the scripts will be later consolidated into a single `manage.py` interface: https://github.com/open-contracting/standard_profile_template/issues/40

PR #1125, opened 2020-11-24
Add 'weight' to Item/unit

yolile:
ref #1110


schema/dereferenced-release-schema.json
```diff
@@ -2173,6 +2173,29 @@
                       "string",
                       "null"
                     ]
+                  },
+                  "weight": {
+                    "title": "Weight",
+                    "description": "The weight of the item unit.",
```
duncandewhurst: Unit occurs in a couple of places other than items (see https://github.com/open-contracting/standard/issues/1107).

```suggestion
                    "description": "The weight of one unit.",
```

schema/dereferenced-release-schema.json
```diff
@@ -2173,6 +2173,29 @@
...
+                    "type": "object",
+                    "properties": {
+                      "name": {
+                        "title": "Name",
+                        "description": "The name of the weight, using UN/CEFACT Common Codes for Units of Measurement.",
+                        "type": [
+                          "string",
+                          "null"
+                        ]
+                      },
```
duncandewhurst: This differs from the proposal in https://github.com/open-contracting/standard/issues/1110#issuecomment-730064497
```suggestion
                      "unit": {
                        "title": "Unit",
                        "description": "The code for the unit in which the weight is specified, using UN/CEFACT Common Codes for Units of Measurement.",
                        "type": [
                          "string",
                          "null"
                        ]
                      },
```

schema/dereferenced-release-schema.json
```diff
@@ -2173,6 +2173,29 @@
...
+                      "value": {
```
duncandewhurst: I think this should be either `amount` or `quantity`, since `value` fields in OCDS otherwise refer to the `Value` object.

duncandewhurst (changes requested):
@yolile I mistakenly left my review comments on the dereferenced schema!

schema/dereferenced-release-schema.json
yolile: Hmm, you are right, following the same logic, shouldn't we change `unit` to another name too?
cc @jpmckinney

jpmckinney: I think it's okay to use `unit` again, because there is no better term (unless anyone has a better idea).

A unit of weight (or length, etc.) is basic and standardized, such that it can be expressed as a single code string. (We could of course add fields to reference the standards body, fields from [QUDT](http://www.qudt.org/), etc., but such details can be looked up in other datasets based on the code only.)

A unit of goods is neither basic nor standardized (at least, not in all cases), requiring multiple data elements, so it is expressed as an object.

It is the same term `unit`, but in different contexts, so I think it's fine for the value to differ. We *could* disambiguate by adding prefixes like `weightUnit`, but that goes against our [style guide](https://ocds-standard-development-handbook.readthedocs.io/en/latest/meta/schema_style_guide.html#field-and-code-names).

As for `value`, I agree that `quantity` is better, which is commonly used with units of measure. `value` has a tendency toward overuse anyway, so if we can reduce confusion, that is better.

jpmckinney: Or, we could use `unitID` or `unitCode`, since it's identifying the unit, for lookup in other datasets. We don't do this for `currency`, `language` #858 or `format` #857, but we are considering it for `countryCode` #524.

jpmckinney: Actually, we probably want `unit` to be an object with `scheme` and `id`, so that people can choose between UN/CEFACT, QUDT, etc. #1127 This can be a reference to a `SimpleUnit` definition or similar.

Then, as part of #1107, we could add a `ComplexUnit` definition for item/unit.

schema/release-schema.json
```diff
@@ -1434,6 +1434,26 @@
...
+              "type": "object",
+              "properties": {
+                "unit": {
+                  "title": "Unit",
+                  "description": "The code and scheme for the unit in which the weight is specified.",
+                  "$ref": "#/definitions/SimpleUnit"
+                },
+                "quantity": {
+                  "title": "Quantity",
+                  "description": "The number of units",
```
duncandewhurst:
```suggestion
                  "description": "The number of units.",
```

schema/release-schema.json
```diff
@@ -2135,6 +2155,41 @@
         }
       }
     },
+    "SimpleUnit": {
+      "title": "Simple Unit",
+      "description": "A simple unit consists of: an identifier for the list (scheme) from which the unit is taken, and an identifier for the category from that list being applied.",
```
duncandewhurst:
```suggestion
      "description": "A simple unit of measure, consisting of: an identifier for the list (scheme) from which the unit is taken, and an identifier for the unit from that list.",
```

schema/release-schema.json
```diff
@@ -2135,6 +2155,41 @@
...
+          "openCodelist": true
+        },
+        "id": {
+          "title": "ID",
+          "description": "The unit code taken from the scheme. Although an integer is allowed, it is recommended to use a string.",
+          "type": [
+            "string",
+            "integer",
+            "null"
+          ],
```
duncandewhurst: Is there a reason to use a different description from `item/unit/id`?

Since this is a new object, we can restrict the type to string only:

```suggestion
          "description": "The identifier from the codelist referenced in the `scheme` field. Check the [unitClassificationScheme](https://standard.open-contracting.org/{{version}}/{{lang}}/schema/codelists/#unit-classification-scheme) codelist for details of how to find and use identifiers from the scheme in use.",
          "type": [
            "string",
            "null"
          ],
```

schema/release-schema.json
```diff
@@ -2135,6 +2155,41 @@
...
+          "versionId": true
+        }
+      },
+      "patternProperties": {
+        "^(description_(((([A-Za-z]{2,3}(-([A-Za-z]{3}(-[A-Za-z]{3}){0,2}))?)|[A-Za-z]{4}|[A-Za-z]{5,8})(-([A-Za-z]{4}))?(-([A-Za-z]{2}|[0-9]{3}))?(-([A-Za-z0-9]{5,8}|[0-9][A-Za-z0-9]{3}))*(-([0-9A-WY-Za-wy-z](-[A-Za-z0-9]{2,8})+))*(-(x(-[A-Za-z0-9]{1,8})+))?)|(x(-[A-Za-z0-9]{1,8})+)))$": {
```
duncandewhurst: Are `patternProperties` necessary? I don't see a `description` field which can be translated.

duncandewhurst (changes requested):

schema/release-schema.json
yolile: You are right!

duncandewhurst (approved):

yolile:
@jpmckinney FYI this [check](https://github.com/open-contracting/standard/pull/1125/checks?check_run_id=1650015911#step:8:14) is failing

jpmckinney:
> @jpmckinney FYI this [check](https://github.com/open-contracting/standard/pull/1125/checks?check_run_id=1650015911#step:8:14) is failing

Check now passes.

jpmckinney:
NB: Reverted in https://github.com/open-contracting/standard/pull/1481. See https://github.com/open-contracting/standard/issues/1343

Issue #1156, opened 2021-01-06
Review procuringEntity description


jpmckinney:
@JachymHercher wrote in #825 

> ProcuringEntity covers three codes from eForms' organisation role: CPB wholesaler, CPB intermediary and a procurement service provider. They could either be split as they are in eForms, or they could be dealt with by extending the "fullly accurate" approach mentioned above (e.g. by adding codes such as "party receiving the items", "party distributing the items", "party signing the contract for its own use", "party signing the contract for the use by another party", "party using the items").

Re: "fully accurate", this was described as:

> We can group the "needs" / "use cases" into three levels:
> 
> - Basic: distinguishing the "buyer side", supplier, subcontractor, unsuccessful tenderers, review bodies, etc. 
> - Medium: within the "buyer side", distinguishing "buyers", "procuring entities", "central purchasing bodies", "contract implementors" (see https://github.com/open-contracting/standard/issues/548#issuecomment-727912060), etc. 
> - Fully accurate: knowing who exactly does what, i.e. organisation receiving tenders, org. processing tenders, org. signing the contract, org. whose budget is used to pay for the contract, org. sending the payment, etc.
>
> The first two are based primarily on "who they are" (often based on legal definitions), the third is rather "what they do" (often based on real world actions).  Judging from the EU, I think most systems work on the first two levels, because that's the way the legislation is written, people are used to it, and requirements it puts on manual data input. However, I think systems sometimes "dip" into the third level, but less for "analytical" purposes and more for "practical" purposes such as "who do I actually send my tender to", "who do I actually send my questions to", etc. 
> 
> In eForms, we try to cover the first two levels by [organisation role](https://op.europa.eu/en/web/eu-vocabularies/at-concept-scheme/-/resource/authority/organisation-role/?target=Browse&uri=http://publications.europa.eu/resource/authority/organisation-role), while giving the possibility to go into the third level with [organisation subrole](https://op.europa.eu/en/web/eu-vocabularies/at-concept-scheme/-/resource/authority/organisation-subrole/?target=Browse&uri=http://publications.europa.eu/resource/authority/organisation-subrole). 

I guess as a starting point: Is "procuring entity" a useful concept on its own? If so, then this issue is about adding subroles (see below quote). If not, then this issue is about replacing it with new options. Viz. subroles:

> OCDS does not have an equivalent of eForms' organisation subrole. Since legal definitions differ but real world actions are univeral, I think this approach might be valuable for standardization.

[...]

PR #1163, opened 2021-01-12
Improving the description of procuringEntity

### LINKED ISSUE #571: Edit tender.procuringEntity description
Since the definition of procuringEntity is "The entity managing the procurement", I would expect that every procurement process should have at least one such entity. I think there is currently no such requirement for an OCDS contracting process. Shouldn't there be?

I'm also wondering whether saying "procurement process" instead of "procurement" in the description ("The entity managing the procurement process") wouldn't be a more consistent use of terminology?

----

Editor's summary: Change the definition of procuringEntity to "The entity managing the procurement process"

ColinMaudry:
#571


ColinMaudry:
Unless we want to improve the definition further, it's ready for review

ColinMaudry:
Oops I still need to add the changelog entry.

jpmckinney:
Sorry, I had accidentally opened #1156, forgetting that #571 was open. We'll need to continue discussion in #571 before completing this PR.

jpmckinney (approved):

PR #1174, opened 2021-01-17
1.2 dev - issue 895

### LINKED ISSUE #895: Clarify award description
# Background

`awards` in OCDS:

> Information from the award phase of the contracting process. There can be more than one award per contracting process e.g. because the contract is split among different providers, or because it is a standing offer.

`Award` in OCDS:

> An award for the given procurement. There can be more than one award per contracting process e.g. because the contract is split among different providers, or because it is a standing offer.

[UNCITRAL glossary](http://www.uncitral.org/pdf/english/texts/procurem/ml-procurement-2011/Glossary-e.pdf) defines "Award of a procurement contract or framework agreement":

> A final stage of the procurement proceedings regulated by the Model Law, resulting in the conclusion and entry into force of a procurement contract or framework agreement between the procuring entity and selected supplier(s) or contractor(s).

and "Public notice of the award":

> Announcement to the public in general through publication in the media specified in the legislation of the enacting State to whom the procurement contract or the framework agreement was awarded and the price of the procurement contract.

The eProcurement Ontology has a class for Award Decision:

> Resolution of the buyer as to the result of the procurement procedure.

and for Contract Award Notice:

> An announcement of the award or non-award of a contract by a procuring entity.

This was also an "additional note" in #790

# Discussion

The schema and documentation are not clear what, precisely, is intended by 'award'. This therefore leaves its precise semantics up to interpretation by publishers, which leads to different publishers choosing different semantics, which reduces the value of the standard in terms of interoperability and comparability.

In OCDS, the `Award` object is intended to communicate a direct relationship between items, suppliers, and values. It should be possible to know, at the award stage, in OCDS data, which items will later be supplied by which suppliers, and what the value of those contracts will be. If all suppliers and all items are put into one object, then there are no direct relationships, and it is not possible to know the breakdown intended to be communicated by the Award object. This is reflected within the description of `suppliers`:

> If different suppliers have been awarded different items or values, these should be split into separate award blocks.

For the sake of clarity, the `Award` object is not intended to match the 'award notice', because some jurisdictions announce all the relationships above at once. It is also not intended to match the 'award decision', because some jurisdictions create all the relationships in one decision. Another reason for confusion is that some jurisdictions have no term matching the `Award` object.

(See related discussion in #249)

# Proposal

A new definition can better describe the intention in the discussion, and align (where possible) with the definitions of the UNCITRAL glossary or eProcurement Ontology.

JachymHercher:
Closes https://github.com/open-contracting/standard/issues/895 (changed definition for award and awards)

### DIFF
--- a/schema/release-schema.json
+++ b/schema/release-schema.json
@@ -91,7 +91,7 @@
     },
     "awards": {
       "title": "Awards",
-      "description": "Information from the award phase of the contracting process. There can be more than one award per contracting process e.g. because the contract is split among different providers, or because it is a standing offer.",
+      "description": "Information from the award phase of the contracting process.",
       "type": "array",
       "items": {
         "$ref": "#/definitions/Award"
--- a/schema/release-schema.json
+++ b/schema/release-schema.json
@@ -485,7 +485,7 @@
     },
     "Award": {
       "title": "Award",
-      "description": "An award for the given procurement. There can be more than one award per contracting process e.g. because the contract is split among different providers, or because it is a standing offer.",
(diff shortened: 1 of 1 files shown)


JachymHercher:
Replaced by #1175

Issue #1160, opened 2021-01-06
Update all status codelists


jpmckinney:
The 'unsuccessful' code's description is:

> This award could not be successfully made. If items or supplier details are included within the award section, then these narrow the scope of the unsuccessful award (i.e. the award of noted items, or an award to the noted supplier, was unsuccessful, but there can be other successful awards for different items listed in the tender, or to different suppliers).

This seems to imply that if the process is unsuccessful, then there would be an award object without suppliers and `status` set to 'unsuccessful'.

But, https://standard.open-contracting.org/latest/en/guidance/map/unsuccessful_tender/ doesn't mention awards. This seems correct for, e.g., a procedure where all tenderers were excluded because of exclusion grounds or selection criteria.

The OCDS for EU profile uses awards in [F03](https://standard.open-contracting.org/profiles/eu/master/en/forms/F03/#/AWARD_CONTRACT/NO_AWARDED_CONTRACT). Since there might be many CANs for the same procedure, and it's unknown whether a later CAN will be successful, this approach makes sense. But, it means all awards could be unsuccessful.

JachymHercher:
More generally: why isn't there just a single contractingProcess.status field, i.e. what are the use cases for having one in each block? Also, shouldn't there be a planning.status, since there could be an OCDS planning process that would not have `Tender` (cf https://github.com/open-contracting/standard/issues/866#issuecomment-751683275). 

eForms have a [Not Awarded Reason (BT-144)](https://op.europa.eu/en/web/eu-vocabularies/at-concept-scheme/-/resource/authority/non-award-justification) field that lists various reasons why a process could be unsuccessful. I could imagine classifying the different fields into whether the "unsuccess" happened at the tender, award or contract stage. However, if that is the approach, we should explain it in quite some detail, probably both in the general https://standard.open-contracting.org/latest/en/guidance/map/unsuccessful_tender/ (the name of which shouldn't be just about the tender though) and https://standard.open-contracting.org/latest/en/schema/codelists/#closed-codelists.

ColinMaudry:
@JachymHercher The idea behind `tender.status` is have a single field to refer to in order to know how the contracting process went/is going.

The problem is that it doesn't carry enough granularity and one must analyze the data further (awards? contracts? suppliers?) in order to infer accurate insights.

The proposal of Jachym is interesting:

[...]

Issue #891, opened 2019-07-11
Merge Lots extension


jpmckinney:
## Discussion

In retrospect, it would have been better to have lots in OCDS (not in an extension), and to always have at least one lot in every process. For example, in the EU, for eForms and the e-Procurement Ontology, the proposal is that, even if the procedure is not divided into lots, there will still be a single, virtual lot.

The advantage is that it makes analysis identical whether the procedure is divided into multiple lots or not. Right now, analysts need to analyze fields on `tender` and then also fields on `lots` (if present). The analyst doesn't know what to do if there is conflicting data between `lots` and `tender` (e.g. do `lots` take priority? how to reconcile the two?).

The current work to improve OCDS to cover EU requirements is leading to the duplication of dozens of fields on `tender` and `lots`, to satisfy cases where procedures are divided into multiple lots or not. It would be much simpler for users if instead there were just one unit of analysis (the lot) instead of two (the lot and tender objects).

## Proposal

For OCDS 1.2, the proposal is to deprecate the fields on `tender` that ought to be on `lots`, and to merge the lots extension into the release schema. Then, in OCDS 2.0, those fields would be removed on `tender`.

ColinMaudry:
> For OCDS 1.2, the proposal is to deprecate the fields on tender that ought to be on lots, and to merge the lots extension into the release schema

The fields that would remain in `tender` would be those that systematically apply to all lots.

jpmckinney:
Yes, like the procuring entity, procurement method, etc.

jpmckinney:
A couple relevant consecutive comments in this issue: https://github.com/open-contracting/standard/issues/888#issuecomment-516972841 In particular:

> [In OCDS, 'Lot' might be better understood as] 'ItemContainer', … described as "A group of one or more items, where each group is bid on, awarded, and supplied separately. If a contracting process has no groups, it is interpreted as having a single group for all items."

Expressed that way, it is easier to explain why OCDS would have a lot even if the original procedure doesn't explicitly refer to any lot.

jpmckinney:
We need to also think through how to integrate extensions into the docs without overwhelming users.

ColinMaudry:
> We need to also think through how to integrate extensions into the docs without overwhelming users.

By the docs I assum you mean the core docs.

@jpmckinney What would overwhelm them? The quantity of the docs? Or the complexity (regarding this issue, I think it brings simplicity)?

jpmckinney:
[...]

Issue #790, opened 2018-12-14
Discussion: One contract related to multiple awards


jpmckinney:
A process in Kyrgyz Republic might make independent award decisions for each lot, then group the lots by supplier and sign a single contract for each group of lots.

Award decisions are made at a different time than contracts are signed, and we want award decisions to be disclosed as early as possible. At the time of the award decision, we don't know how the lots will be grouped into contracts. (I assume that, in some cases, a buyer might sign multiple contracts with the same supplier instead of always grouping its lots.) Therefore, the publisher can't disclose the awards such that they can avoid (in this case) a one-to-many mapping from contract to awards.

I don't know an elegant solution besides having e.g. `awardIDs` on `Contract`.

Additional note: We should clarify in the documentation that the "Award" means the "Award decision".

Related: #249

JachymHercher:
In eForms, a lot can receive one or more bids and a contract can be linked to one or more winning bids, which I think is equivalent to the situation described above. This seems to be a common use case for example in the medical sector, where a procedure can contain hundreds of lots for particular pharmaceuticals / chemical substances and a few pharma companies submit their bids for all lots. Once the bids are evaluated, a contract is signed with each bidder, not per lot, as that would be very cumbersome.

yolile:
Could this be modeled as: create an award for each supplier, but without declaring `award/items`. Then, for each supplier, create a contract that relates to its award and disclose the items in `contract/items`?

jpmckinney:
To narrow the issue:

* a lot can receive one or more bids
  * Covered by [bids.details.relatedLots](https://extensions.open-contracting.org/en/extensions/lots/master/schema/#bid)
* a contract can be linked to one or more winning bids
  * Presently covered for a single bid by [awards.relatedBid](relatedBid) (to which a contract is related via `contracts.awardID`). There is an issue to [make awards.relatedBid an array](https://github.com/open-contracting/ocds-extensions/issues/125).
* a contract is signed with each bidder, not per lot
  * Covered by [awards.relatedLots](https://extensions.open-contracting.org/en/extensions/lots/master/schema/#award), to link multiple lots to one award

In the issue description, the scenario (which we might want to validate) is that apparently the Kyrgyz Republic *does* do the cumbersome step of awarding per lot, and only later signs contracts per bidder.

[...]

PR #1175, opened 2021-01-20
Update definition of award

### LINKED ISSUE #895: Clarify award description

JachymHercher:
closes #895


jpmckinney:
@JachymHercher We didn't discuss this one. What is keeping it in draft?

JachymHercher:
Two things:
- Added https://github.com/open-contracting/standard/pull/1175/commits/2a2d451e1b6ca3177cef87aba18497f3a60fbdc9 to change the guidance.
- Still waiting for the resolution of https://github.com/open-contracting/standard/issues/1160, which might add a sentence to the definition.

jpmckinney:
> Still waiting for the resolution of #1160, which might add a sentence to the definition.

I've lost track of what needs to be resolved. Can you summarize?

JachymHercher:
I've made a summary in https://github.com/open-contracting/standard/issues/1160#issuecomment-792348486 and I think there probably won't be any implications for the description of award, assuming we explain it in the guidance and other places. I'll add the changelog and submit the PR.

docs/history/changelog.md
```diff
@@ -14,6 +14,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
 * [#955](https://github.com/open-contracting/standard/pull/955) Split the Guidance section into sub-sections. Add a new History section, and move the Changelog, Credits, and History and Development pages to it.
 * [#958](https://github.com/open-contracting/standard/pull/958) Governance page: Change responsibility for prioritization review to OCP's Head of Data Products & Services. Change "technical team" to "standard development team".
 * Guidance section:
+  * [#1175](https://github.com/open-contracting/standard/pull/1175) Update description of `Award`
```
```suggestion
  * [#1175](https://github.com/open-contracting/standard/pull/1175) Use OCDS' definition of "award" instead of UNCITRAL's definition.
```

docs/history/changelog.md
```diff
@@ -65,7 +66,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
 * [#1238](https://github.com/open-contracting/standard/pull/1238) Open the `releaseTag` codelist.
 
 ### Schema
-
+* [#1175](https://github.com/open-contracting/standard/pull/1175) Update description of `Award`
```
```suggestion

* [#1175](https://github.com/open-contracting/standard/pull/1175) Update the description of `Award`.
```

schema/release-schema.json
```diff
@@ -492,7 +492,7 @@
     },
     "Award": {
       "title": "Award",
-      "description": "An award for the given procurement. There can be more than one award per contracting process e.g. because the contract is split among different providers, or because it is a standing offer.",
+      "description": "Decision by the buyer or procuring entity on the supplier with whom it intends to conclude a contract, including the items the supplier should supply and their price. Typically, this decision leads to a contract, but not always (e.g. the award is appealed at court or the supplier refuses to sign the contract). Depending on the jurisdiction, a single decision may concern a single supplier, item and value; or batches of suppliers and/or items and/or values. Similarly, sometimes the award is published as soon as it is made and sometimes only together with a contract (including only being implicitly covered by the contract). As far as possible, the award should be published at the most granular level (i.e. a given supplier will deliver a given item at a given value) and as soon as it is made.",
```
jpmckinney: @JachymHercher Sorry, I hadn't caught this earlier. But what does "(including only being implicitly covered by the contract)" mean?

If a lot has many items, does the last sentence suggest that each item should be published separately? Should the "i.e." be re-phrased?

We need to avoid the "should" in "the items the supplier should supply" which is descriptive (not normative). Maybe rephrase as "the items to be supplied".

JachymHercher: > @JachymHercher Sorry, I hadn't caught this earlier. But what does "(including only being implicitly covered by the contract)" mean?

I assumed (mainly on the basis of "Source systems can contain data on awards, on contracts, or on both." in https://standard.open-contracting.org/latest/en/guidance/map/awards_contracts_buyers_suppliers/) that there are jurisdictions that don't recognize  "award" as a separate stage in contracting, but just think about contracts. The phrase "Only being implicitly covered by the contract" was supposed to describe this scenario. 
1. I'm not sure this scenario exists? If not, we can remove the sentence.
2. If yes, I agree I should have worded this more clearly. On second thought though, maybe an extra clarification isn't necessary anyway and "only together with a contract" is clear enough? 

> If a lot has many items, does the last sentence suggest that each item should be published separately? Should the "i.e." be re-phrased?
3. It does imply that. Isn't it correct? "As far as possible", don't we want users to separately publish the value for each item / member of the consortium, which then requires an award for each item / member? (If we removed / weakened this significantly, do we say somewhere that e.g. publishing a single award for all lots is a bad practice (outside a framework agreement, see point 5) ?) 

> We need to avoid the "should" in "the items the supplier should supply" which is descriptive (not normative). Maybe rephrase as "the items to be supplied".

4. Why shouldn't this be normative? (Probably linked to the previous point :). )

5. Linked to this discussion, so I'll add it here: in https://github.com/open-contracting/standard/issues/896#issuecomment-755067515 we agreed "That we should recommend/require a single Award that informs about all suppliers in the FA." I'd propose adding it to the definition of award, possibly phrased as an exception to the "as granular as possible".

jpmckinney: 1. Given that we define an award as "Decision by the buyer or procuring entity on the supplier with whom it intends to conclude a contract", I think it is already clear that such a decision must have been made whenever there exists a signed contract, regardless of whether a formal award step preceded the contract's signature.
2. Yes, I don't think the extra clarification is needed, per (1).
3. A contract is related to a single award via `contracts.awardID`. As such, the award cannot be more granular than the eventual contract. The contract, in OCDS, should match as much as possible the real-world contract, so that we have the same number of contracts in OCDS as in the real world, with the same values, etc. The award, on the other hand, is a more abstract concept, and as such can be made more or less granular than the contract. Ideally, it is as granular (but not more granular) than the contract. For example, a lot might have 10 items, and the intention of the buyer is to sign a single contract with the lot's winner, not 10 contracts with the same winner for each item. In that case, there should be one award for that lot (if the procedure is successful, without failed contract signatures, etc.)
4. "the items the supplier should supply" describes a norm established by the real-world contract. It is not a norm established by OCDS. OCDS has no normative power to oblige suppliers to do anything. As such, we need to use a non-normative word instead.
5. Sounds good. Let's work on the wording in #895, since PR comments are less traceable.

JachymHercher: Points 1,2,4,5 are clear, thank you. Concerning point 3, I'd like to make sure I understand it correctly: 

a. In reality, a single contract can stem from one or more (in case of multiple lots) decisions on which tenderer should get the contract. 

In OCDS 1.X, a single `Contract` follows a single `Award`. Using the lot extension, a single `Award` can refer to multiple lots, so a single contract can cover multiple lots. That's good. However, since some fields are only included in `Award`, this is not flexible enough to always describe reality precisely.  Our (reasonable) approach is that when we have to choose between `Contract` or `Award` being closer to reality, we choose `Contract`. For example, when multiple lots (with different values and dates of award) are covered by one contract, then instead of having multiple `Contract`s (each linked to a single award for each lot; each having a value and a date), we stick to one `Contract` and live with not being able to publish information about values of individual lots nor their individual dates of award. Instead, we publish the total value of the award for all lots and *some* date. Do I understand this correctly? If yes, is it documented somewhere, if not, should we add it, probably including guidance on which date to include (e.g. "use the mode / mean date")? 

For OCDS 2.X, do we already have an issue somewhere suggesting Contract:Award being 1:N? (I'm pretty sure there is and I'm pretty sure we alrady discussed this at some point, sorry.)

b. Shouldn't the description of `Award` work better with plurals? In particular, e.g. since an award can concern multiple lots, shouldn't we replace "Decision" by "Decision(s)"?

jpmckinney: a) Yes, that is correct. In practice, do buyers award multiple lots to a single supplier on different dates? I had assumed they would decide this at once. Not having anywhere to state the awarded value of the lot is indeed an oversight. I've created an issue: https://github.com/open-contracting/ocds-extensions/issues/162 Alternately, not anticipating that lots can be awarded separately and making awards-contracts a 1-n instead of a m-n relationship was an oversight. I added a comment https://github.com/open-contracting/standard/issues/891#issuecomment-832277089

I don't think this is documented anywhere. In any case, it is not in the schema, where it should be. So, let's add it.

For 2.x, there is this issue, which we can re-open: https://github.com/open-contracting/standard/issues/790

b) I guess it depends on our metaphysical perspective on decisions. Is a ham and cheese baguette three decisions, or one? I think it's okay to use the singular for the first word. I worry using the plural is confusing.

JachymHercher: a) I've double checked with @jskuhrovec and it seems quite common. For example, an award for one of the lots may be delayed compared to the other ones because the company is answering clarification questions from the buyer about its offer, but they want to move forward with the other lots.

Ah, https://github.com/open-contracting/standard/issues/790, right. Indeed. 

b) Funny example :-). I'm wondering whether it isn't subjective/metaphysical just because of a lack of context, while with context, it would be clear cut. In our cases, I think the decision is "choose X amongst multiple options". For sandwiches, this means that if the menu has "ham and cheese baguette", "philadelphia and salmon baguette", "whatever baguette", you are making one choice. If, by contrast, the menu has 1) "bread", "baguette", "bun"; 2) "protein A/B/C" 3) "salad A/B/C" and you want to chose one of each, then it's three decisions. Doesn't chosing a winning bid from multiple bids for the given lot clearly fall within the latter category?

jpmckinney: b) It's tricky. While standing at the counter, it seems more correct to describe the situation as three decisions, but when a colleague later asks "what did you decide for lunch?", it seems more correct to describe the answer as one decision. The later number is invariable, whether or not additional decisions were introduced at the counter, e.g. "we're out of baguette, will you take a bun?", or a friend says "I thought you were trying to cut down on meat." If the person shared all their micro-decisions in their answer, it might be more correct to describe the answer as multiple decisions, but this would have exceeded the quantity of information that their colleague wanted or expected. ([Maxim of Quantity](https://www.sas.upenn.edu/~haroldfs/dravling/grice.html))

Now, from the perspective of describing the awards in a contracting process, are we (1) at the counter, (2) summarizing our activity for the reader, or (3) over-sharing our activity for the reader?

I think we are at (2). In the case of automated decision making with real-time OCDS publication, we might move to (1), but I consider this case not worth considering. We might be at (3) if we view the disclosure as over-sharing. 

In any case, I think the choice of singular or plural might be best resolved after we decide how to update the rest of the definition to accommodate the award of lots at different times (unless we don't need to?).

JachymHercher: b) I've tried hard to wrap my head around b), but I'm probably in too deep in my line of thought here - I'd never want to summarize information about different lots (even though anything can be in principle summarized into one big decision, sure, but...).  Maybe it's their independence that is important for me? They can have different subject matter, tenderers, criteria, evaluation committees, pretty much anything... I wouldn't want to summarize that. 

But indeed, best resolved after we resolve the differing information per lot / per award under one contract (I guess mainly in https://github.com/open-contracting/standard/issues/891. The changes might end up being none / minimal though.

JachymHercher: I have implemented points 1 - 4, see https://github.com/open-contracting/standard/issues/895#issuecomment-888259348. 

Concerning point 5, see https://github.com/open-contracting/standard/issues/896#issuecomment-888254209. 

The only remaining question in this thread is a detail - "decision" or "decision(s)". Philosophically, I'm very much on the  "decision(s)" side. Whether in practice that's the case of "too much precision causes confusion" I leave to you :).

jpmckinney:
Since the comments might be collapsed, note a long thread in https://github.com/open-contracting/standard/pull/1175#discussion_r616320247

schema/release-schema.json
```diff
@@ -546,7 +546,7 @@
     },
     "Award": {
       "title": "Award",
-      "description": "An award for the given procurement. There can be more than one award per contracting process e.g. because the contract is split among different providers, or because it is a standing offer.",
+      "description": "Decision by the buyer or the procuring entity on the supplier with whom it intends to conclude a contract, including the items to be supplied and their price. Typically, this decision leads to one or more contracts, but not always (e.g. the award is appealed at court or the supplier refuses to sign the contract). Depending on the jurisdiction, a single decision may concern a single supplier, item and value; or batches of suppliers and/or items and/or values. Similarly, sometimes the award is published as soon as it is made and sometimes only together with a contract.",
```
jpmckinney: Using value instead of price, as it corresponds to the field name.

```suggestion
      "description": "Decision by the buyer or the procuring entity on the supplier with whom it intends to conclude a contract, including the items to be supplied and their value. Typically, this decision leads to one or more contracts, but not always (for example, the award is appealed at court or the supplier refuses to sign the contract). Depending on the jurisdiction, a single decision may concern a single supplier, item and value; or batches of suppliers and/or items and/or values. Similarly, sometimes the award is published as soon as it is made and sometimes only together with a contract.",
```

jpmckinney:
In a8eae81bf9f35c5abd167917dc4086b0ecfe9c7d I had to remove the last sentence with the link the GitHub, because while that issue does happen to discuss many-to-one awards-to-contracts, it is not the primary subject of the issue, and so readers are likely to be confused (especially if we close it per the primary subject).

JachymHercher:
Makes sense. Are we good to merge?

PR #1187, opened 2021-01-31
Add 'contractImplementationManager' in partyRole (548)

### LINKED ISSUE #548: partyRole: Add code for Contract administration entity
The code list for party roles doesn't include any "Contract administration entity"

http://standard.open-contracting.org/latest/en/schema/codelists/#party-role

A contract administration entity might be present in more large contracts and in PPPs. Would you consider adding it to the open code list as one of the defaults?

JachymHercher:
closes #548


jpmckinney:
Looks like this PR branched from the #1182 branch instead of 1.2-dev. I'll review #1182 first.

jpmckinney (approved):

PR #1192, opened 2021-02-01
Initial update to use MyST Parser

choldgraf:
This is a first-pass PR to get some low-hanging fruit out of the way. It does the following things:

### DIFF
--- a/common-requirements.in
+++ b/common-requirements.in
@@ -4,8 +4,8 @@ Sphinx<3
 ocds-babel>=0.2.1
 # See https://ocds-babel.readthedocs.io/en/latest/api/translate.html#install-requirements-for-markdown-translation
 -e git+https://github.com/readthedocs/commonmark.py.git@dafae75015cc342f3fddb499674bab97ac4a6a96#egg=commonmark
--e git+https://github.com/jpmckinney/recommonmark.git@hotfix#egg=recommonmark
 -e git+https://github.com/open-contracting/standard_theme.git@open_contracting#egg=standard_theme
+myst-parser~=0.13.3
 
 # Make
 sphinx-intl<1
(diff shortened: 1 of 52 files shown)


ColinMaudry:
What are the benefits of switching to MyST?

choldgraf:
I'll let @jpmckinney weigh in with his thoughts on this repo specifically, but in general the main benefit of MyST is that it "natively" speaks Sphinx. It is a superset of commonmark that includes CM syntax + a few extra tokens that are designed to work for Sphinx. So you can define roles, directives, etc directly in markdown without using `eval_rst` everywhere. See [the myst docs](https://myst-parser.readthedocs.io/en/latest/) for some features/information/etc.

jpmckinney:
Hi @ColinMaudry, as background, OCP is working with @choldgraf to update the standard's theme, to improve navigation issues, etc. This includes switching to MyST, which is better functionally and is better maintained than recommonmark / Commonmark. You can see some of the issues at https://github.com/open-contracting/standard_profile_template/issues

jpmckinney:
@choldgraf I just realized that it'd actually be better to use 1.1-dev as our base branch. Version 1.2 is in development and might be a ways off, but we'd like to make these improvements available for users of the current version 1.1.

I committed a change to 1.1-dev that should fix the issue you were having earlier (re: conflicting requirements) #1194. If you `pip install pip-tools` and then run `pip-sync`, then it should work.

As for the commits you already made, I think you can rebase them onto a new branch from 1.1-dev. I was reading the last part of [this page](https://www.devroom.io/2010/06/10/cherry-picking-specific-commits-from-another-branch/), but ran out of time today. Alternately, if there are conflicts that are hard to resolve, perhaps you can cherry-pick some of the commits and re-do others?

choldgraf:
hey @jpmckinney hmm OK I'll give a shot at rebasing and will cross my fingers that it isn't too disruptive haha

PR #1197, opened 2021-02-04
Updating requirements for myst-parser

choldgraf:
This is a first-pass PR to get some low-hanging fruit out of the way. It does the following things:

supercedes https://github.com/open-contracting/standard/pull/1192

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,8 @@ jobs:
       run: |
         if [ $GITHUB_REPOSITORY = 'open-contracting/standard_profile_template' ]; then python schema/build-profile.py; make extract; fi
         make
-        python util/add_translation_notes.py
+      # NOTE: un-comment when this script is fixed for the new MyST docs.
+      # python util/add_translation_notes.py
     - run: pytest
     - uses: shimataro/ssh-key-action@v2
       with:
--- a/common-requirements.in
+++ b/common-requirements.in
@@ -7,6 +7,7 @@ ocds-babel>=0.2.1
 -e git+https://github.com/readthedocs/commonmark.py.git@dafae75015cc342f3fddb499674bab97ac4a6a96#egg=commonmark
 -e git+https://github.com/open-contracting/standard_theme.git@open_contracting#egg=standard_theme
 myst-parser~=0.13.3
+mdit-py-plugins~=0.2.5
 
 # Make
 sphinx-intl<1
--- a/common-requirements.in
+++ b/common-requirements.in
@@ -4,8 +4,8 @@ Sphinx<3
 ocds-babel>=0.2.1
 # See https://ocds-babel.readthedocs.io/en/latest/api/translate.html#install-requirements-for-markdown-translation
 -e git+https://github.com/readthedocs/commonmark.py.git@dafae75015cc342f3fddb499674bab97ac4a6a96#egg=commonmark
--e git+https://github.com/jpmckinney/recommonmark.git@hotfix#egg=recommonmark
 -e git+https://github.com/open-contracting/standard_theme.git@open_contracting#egg=standard_theme
+myst-parser~=0.13.3
 
 # Make
 sphinx-intl<1
--- a/common-requirements.in
+++ b/common-requirements.in
@@ -1,4 +1,5 @@
-Sphinx<3
+Sphinx
+-e git+https://github.com/sphinx-doc/sphinx.git@3207da1fa81a5d3171d3bbb15768e1c6f6b39b58#egg=sphinx
 
 # Sphinx
 ocds-babel>=0.2.1
--- a/common-requirements.txt
+++ b/common-requirements.txt
@@ -6,6 +6,11 @@
 #
 -e git+https://github.com/readthedocs/commonmark.py.git@dafae75015cc342f3fddb499674bab97ac4a6a96#egg=commonmark
     # via -r common-requirements.in
+-e git+https://github.com/sphinx-doc/sphinx.git@3207da1fa81a5d3171d3bbb15768e1c6f6b39b58#egg=sphinx
+    # via
+    #   -r common-requirements.in
+    #   myst-parser
+    #   sphinx-intl
 -e git+https://github.com/open-contracting/standard_theme.git@open_contracting#egg=standard_theme
     # via -r common-requirements.in
 alabaster==0.7.12
@@ -111,11 +116,6 @@ snowballstemmer==2.0.0
     # via sphinx
 sphinx-intl==0.9.12
     # via -r common-requirements.in
-sphinx==2.4.4
-    # via
-    #   -r common-requirements.in
-    #   myst-parser
-    #   sphinx-intl
 sphinxcontrib-applehelp==1.0.1
     # via sphinx
 sphinxcontrib-devhelp==1.0.1
--- a/common-requirements.txt
+++ b/common-requirements.txt
@@ -63,8 +63,9 @@ markdown-it-py==0.6.1
     #   myst-parser
 markupsafe==1.1.1
     # via jinja2
-mdit-py-plugins==0.2.4
+mdit-py-plugins==0.2.5
     # via
+    #   -r common-requirements.in
     #   markdown-it-py
     #   myst-parser
 more-itertools==8.1.0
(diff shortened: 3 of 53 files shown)


choldgraf:
I think that the cherry-pick worked OK! Will start trying to figure out errors.

BTW, @jpmckinney - I'm trying to locally run the command that the CI build is failing on, but it is giving me a different command locally. I wonder if there's an easy "gotcha" that I am not noticing:

```
$ sphinx-build -q -b dirhtml docs build/es -D language="es"

Exception occurred:
  File "/home/choldgraf/anaconda/envs/ocds/lib/python3.6/gettext.py", line 514, in translation
    raise OSError(ENOENT, 'No translation file found for domain', domain)
FileNotFoundError: [Errno 2] No translation file found for domain: 'schema'
```

I will dig into the docs a bit closer next time to see if there's more in there about building the translations. I don't have a ton of familiarity with translation workflows in Sphinx

jpmckinney:
There are dependencies between commands (eg I think this error is because  the `schema` message catalog hasn’t been compiled; it’s used by PyBabel to translate JSON files - see the `translate` method call in `conf.py`) but if you run `make`, it will run all commands as needed.

There are many Make targets - if you just want to build one language for example.

choldgraf:
Gotcha, thanks. This is probably the most formidable Makefile I've ever used 😅 I am very impressed and a little fearful!

chrisjsewell:
I would suggest turning on fail on errors when building the documentation at e.g.: https://github.com/open-contracting/standard/blob/323dee4273dbf682a2c7f491bef498087edbedee/include/common.mk#L99

`sphinx-build -nW --keep-going -q -b dirhtml $(DOCS_DIR) $(BUILD_DIR)/en`

currently there are a bunch of warnings in the documentation build (even before this PR): https://github.com/open-contracting/standard/runs/1810749509

jpmckinney:
I'm making some updates to one of our Sphinx extensions to not warn about things that are actually fine.

However, some of the warnings I don't know how to quiet. For example:

```
/home/runner/work/standard/standard/docs/schema/codelists.md:18:<translated>:1: WARNING: None:any reference target not found: ../../release-schema.json
```

When we build the documentation, we write some JSON and CSV files to the build directory that Sphinx doesn't know about, so it complains about "target not found".

I figure there's some way to tell Sphinx about those files. I've seen `env.note_dependency` around - not sure if that's the trick I'm missing.

choldgraf:
@jpmckinney would it be acceptable to use the `{download}` role for this? I think that this would suppress these warnings. E.g.:

```
Here is my {download}`schema <../../release-schema.json>`!
```

This is used to link directly to assets that are not "part of the documentation" but are meant to be downloaded

jpmckinney:
Aha! That sounds like the solution. I didn't know about that role (for my later reading: https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#referencing-downloadable-files)

jpmckinney:
@choldgraf Sphinx merged my PR (https://github.com/sphinx-doc/sphinx/pull/8853) for heading level issue (https://github.com/executablebooks/MyST-Parser/issues/302), so we can pin to a commit in Sphinx until the next 3.x release is made.

choldgraf:
Wow, nice! Will add that to the next commit. Right now I am trying to figure out this weird bug that we seem to be triggering in the markdown-it-py library when doing translations.

choldgraf:
looking at your PR there, I wonder if that actually fixes our bug? Will investigate

jpmckinney:
> looking at your PR there, I wonder if that actually fixes our bug? Will investigate

The PR makes the heading level go from 0 to 1 instead of 0 to 2, so it should fix the immediate problem.

jpmckinney:
> Wow, nice! Will add that to the next commit. Right now I am trying to figure out this weird bug that we seem to be triggering in the markdown-it-py library when doing translations.

I think the following error is fixed if you upgrade mdit-py-plugins to 0.2.5 (released for this fix https://github.com/executablebooks/mdit-py-plugins/pull/11)

```
  File "/opt/hostedtoolcache/Python/3.6.12/x64/lib/python3.6/site-packages/mdit_py_plugins/front_matter/index.py", line 51, in frontMatter
    if marker_str[(pos - start) % marker_len] != state.src[pos]:
IndexError: string index out of range
```

choldgraf:
@jpmckinney ahh I just spent way too long trying to figure out what the heck was going on with that bug haha, I didn't realize that was the package to upgrade, I knew it had been addressed somewhere.

FWIW this commit is what I finally used as a workaround, and I have no idea why it works 😅 [`dcd9b4b` (#1197)](https://github.com/open-contracting/standard/pull/1197/commits/dcd9b4b50f825fdef64ee0c4f23e5a5d89366eb6)

OK I reverted that change and pushed another commit that updates mdit-py-plugins.

choldgraf:
@jpmckinney any thoughts on this one? Latest error seems to be some custom docs code in this repo?

https://github.com/open-contracting/standard/pull/1197/checks?check_run_id=1884827925#step:6:64

jpmckinney:
> @jpmckinney any thoughts on this one? Latest error seems to be some custom docs code in this repo?
> 
> https://github.com/open-contracting/standard/pull/1197/checks?check_run_id=1884827925#step:6:64

Oh, yeah, I'll have to edit that script for the new theme (should be easy - just changing a selector). For now, you can comment out `python util/add_translation_notes.py` in `ci.yml`. What it does is, if any translations are missing, it adds an admonition (as HTML) to the page to say, "hey, this page has been updated in English, but we're showing you the most recent Spanish version for now" more or less.

It's possible some of the tests in the `tests/` directory will start failing next, if their selectors don't match the theme.

I'll also need to customize the theme later, because we replace the search page's JavaScript to send a query to Elasticsearch. (The deployment process creates an Elasticsearch index.)

choldgraf:
@jpmckinney sounds good - I don't have any more cycles for this one tonight, so I'll just comment that line out and see what breaks next but then gonna log off :-)

Are you using ElasticSearch just to speed things up? Nifty...

jpmckinney:
> Are you using ElasticSearch just to speed things up? Nifty...

The quality of Sphinx's search results are not great, so we use Elasticsearch, which supports quoted phrases, etc. and we're able to tune the results better.

choldgraf:
@jpmckinney you ever write up your workflow there in a blog post or something? That is pretty cool, a lot of folks ask for improved search in Jupyter Book and I've been trying to figure out ways that we could recommend something, or improve the tech itself

Curious if you ever played around with using something like [lunr](https://lunrjs.com/)

jpmckinney:
Interesting - I'll have a look at Lunr.

I haven't written it up, but https://ocds-index.readthedocs.io/en/latest/ is the Python package that does the work (at the Python-level, it's possible to configure it for different themes).

ocds-index's CLI gets called by this script, which is run by GitHub Actions once the tests pass: https://github.com/open-contracting/deploy/blob/master/deploy-docs.sh#L8-L9

Then, the theme queries Elasticsearch: https://github.com/open-contracting/standard_theme/blob/open_contracting/standard_theme/static/js/search.js#L20-L53

We use https://readonlyrest.com to expose Elasticsearch to the web safely. (You'll notice the JavaScript in the theme uses Basic Authentication nonetheless, just so that random bots don't hit Elasticsearch.)

choldgraf:
Very cool, thanks for sharing!

Note: I'm not sure what the latest failure is. Though tests seem to pass 🙂 

https://github.com/open-contracting/standard/pull/1197/checks?check_run_id=1884900587#step:8:1

jpmckinney:
Oh, it's because this branch is from your fork, rather than in the repo itself, and an ssh-related GitHub Action isn't letting the build access the repo's secrets for that reason.

You said you'd be logging off for the evening 😛 but if you change your origin and push a new branch to the repo directly, then it should build!

choldgraf:
@jpmckinney - hmmm - I don't believe that I have permission to push directly to the standard repo:

```
$ git push upstream myst-parser-switch
remote: Permission to open-contracting/standard.git denied to choldgraf.
fatal: unable to access 'https://github.com/open-contracting/standard/': The requested URL returned error: 403
```

jpmckinney:
Oops, I thought I had already invited you. I've sent an invite now.

PR #1200, opened 2021-02-04
Deprecate all the codes from submissionMethods codelist

### LINKED ISSUE #322: submissionMethod: Deprecate codes
Updates to the submission method codelist are proposed for version 1.1 of OCDS.
## The issue

Submission method is an open codelist which currently includes:
- electronicAuction
- electronicSubmission 
- written
- inPerson

This appears to mix up the mode of submission (online, post, in person), with the way the process is run (auction or otherwise)
## Proposal

This codelist should be updated based upon expert input and consultation
## Engagement

We are inviting suggestions for what the submissionMethod codelist should contain.
- What codes are required by key use cases? (e.g. Red Flag Analysis?)
- Are there existing codelists we can defer to here? 
## History

Comments from user:

> The inclusion of 'electronicAuction' in the [submissionMethod](http://standard.open-contracting.org/latest/en/schema/codelists/#submission-method) codelist appears odd - as use of an auction is a separate issue from submitting electronically.

ColinMaudry:
Closes #322


docs/history/changelog.md
```diff
@@ -35,6 +35,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
 
 ### Codelists
 
+* [#1200](https://github.com/open-contracting/standard/pull/1200) Deprecate all the codes of the `submissionMethod` codelist.
```
jpmckinney: Add an entry about deprecating the field as well.

ColinMaudry: Done, I have also summarized why the codes are deprecated.

PR #1204, opened 2021-02-06
Add field and code description directives

jpmckinney:


jpmckinney:
Need to fix directive to work with translated codelists.

Issue #903, opened 2019-08-01
Add more synonyms to various field descriptions


jpmckinney:
## Background

In at least the EU, design contests use a different set of terms:

* contract -> prize
* bidder -> participant
* supplier -> winner
* tenders (bids) -> projects
* (award period) -> ’date of jury decision’

For other procedures:

* award -> result of procedure

## Proposal

The alternative terms can be added in field descriptions to make it clear that the fields are appropriate for design contests as well as other procedures.

jpmckinney:
## Background

More broadly, OCDS uses `buyer` and `suppliers`, but these terms are inappropriate for many contracting processes, like design contests (above), public private partnerships, government asset sales, concessions, etc.

In the OCDS for PPPs profile, to address this, `buyer` and `suppliers` are deleted from the schema in favor of `publicAuthority` and `preferredBidders` (though 'privateParty' is also used).

In the EU, to address this, the more generic terms 'contracting authority' or 'contracting entity' (the distinction is irrelevant from OCDS' perspective) are used instead of 'buyer'. At the UI level, the EU changes its term on forms from 'contractor' to 'concessionaire' ('supplier' in OCDS), but the TED data uses CONTRACTOR for both.

## Proposal

Rather than multiplying the fields used to identify the two parties, I think it's better to (1) accept we chose less-than-ideal terms and then (2) change the titles/definitions to broaden their scope though ideally in limited ways, e.g. "If this contracting process is to award a concession, then the `buyer` is X. If this contracting process it to sell government assets, then the `buyer` is Y, etc."

We can of course deprecate the old term and add a new term.

From a data use perspective, you would typically narrow the processes to those you care about first (concessions, asset sales, procurements, etc.), then look at the buyers and suppliers (whose semantics are then narrowly scoped). A broader analysis might still want to know e.g. "which government agencies are doing some sort of business with external parties" in which case it's much easier to look for the answer in a pair of fields instead of a growing number of fields, and in which case the broader semantics are acceptable (even desirable).

ColinMaudry:
Looking at eForms annex spreadsheet, I'm not sure contract -> prize.

BG-44 mentions ranks, and BT-41 says the winners of a contest aren't necessarily awarded all contracts, implying prizes and contracts are not bound.

jpmckinney:
In a design contest, my understanding was that participants were at most awarded a prize – but that, following the contest, they might be awarded a contract. Does it work differently?

ColinMaudry:
[...]

PR #1123, opened 2020-11-22
add frameworks guidance

duncandewhurst:
Replaces #1108 and #1106 

closes #784

PR #1208, opened 2021-02-07
Clarify contracts description (896)

### LINKED ISSUE #896: Clarify Contract description

JachymHercher:
#896 

together with #1216, addresses https://github.com/open-contracting/standard/issues/903


jpmckinney:
@JachymHercher Please click "Ready for review" unless this PR is still in draft.

JachymHercher:
I left it as a draft intentionally, because I haven't yet drafted the guidance section. I'm waiting for https://github.com/open-contracting/standard/pull/1123 with that (and perhaps I should wait also for getting started), see https://github.com/open-contracting/standard/issues/896#issuecomment-774772868.

jpmckinney:
@JachymHercher #1123 is merged, and you've now seen the Primer, so I think this issue is now unblocked.

JachymHercher:
Ready for review, see https://github.com/open-contracting/standard/issues/896#issuecomment-888948114

jpmckinney:
Notes on the merge:

* Somehow, docs/guidance/map/awards_contracts_buyers_suppliers.md reappeared – possibly from a bad merge in #1356
* The changelog was getting quite messy, so I took the opportunity to put it in order

I also added a commit for a broken link, that I assume also occurred in #1356.

jpmckinney (approved):

Issue #1190, opened 2021-02-01
Convert documentation to MyST Markdown and new theme


choldgraf:
This is a meta-issue to track conversion of documentation syntax to use MyST Markdown. It may grow or shrink as more focused issues pop up during initial conversion so I'll track them here.

- [x] Initial conversion attempt via MyST parser (https://github.com/open-contracting/standard/pull/1192)
- [x] Convert over eval-rst blocks into MyST markdown (~~#1220, #1221,~~ #1247)
- [x] Check that we are using references to headers / labels rather than hard-coding headers in URLs #1253
- [ ] Convert the theme to either Sphinx Book Theme or the PyData theme: https://github.com/open-contracting/standard/pull/1254
- [ ] Modifications to theme as-needed for the OCDS docs
- [ ] Content modifications that take advantage of theme/myst features (e.g., sphinx-panels)
- [ ] Updates to the contributing content documentation to highlight any new functionality

After this is mostly done

_Note - Chris is not sure if the following are in scope for him or not...he suspects somebody else will be more efficient in doing it than he is_

- [ ] Create sub-theme to replace search page like https://github.com/open-contracting/standard_theme/blob/open_contracting/standard_theme/search.html and https://github.com/open-contracting/standard_theme/blob/open_contracting/standard_theme/static/js/search.js
- [ ] Update ocds-index https://github.com/open-contracting/ocds-index/issues/6

choldgraf:
Also note - I'm going to prefix all of these issues with `[MyST]` to help differentiate them from others, since I cannot add tags etc myself.

choldgraf:
@jpmckinney do you prefer to have one mega-PR that does all of the changes at once, or is it OK to break this into several smaller PRs that get merged over time?

jpmckinney:
I've changed your role to Maintain, so you should be able to create/add labels. An issue prefix is also fine.

I have no preference in terms of PR size. For larger PRs, I sometimes read one commit diff at a time when reviewing.

jpmckinney:
Follow-up for this issue:

* [x] Update selector(s) in util/add_translation_notes.py

Small cleanup that can be done in later PRs:

* [x] Build with `sphinx-build -nW --keep-going -q -b dirhtml $(DOCS_DIR) $(BUILD_DIR)/en` once warnings are cleared
* [x] Change some eval-rst to block-level directives https://myst-parser.readthedocs.io/en/latest/using/intro.html#block-level-directives-with-myst-markdown
[...]

PR #1213, opened 2021-02-12
Updating requirements for myst-parser

choldgraf:
Switching the myst parser branch to `origin` so that the CI can pass fully.

Supercedes #1192 

----

Copying issue description from #1192/#1197:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,8 @@ jobs:
       run: |
         if [ $GITHUB_REPOSITORY = 'open-contracting/standard_profile_template' ]; then python schema/build-profile.py; make extract; fi
         make
-        python util/add_translation_notes.py
+      # NOTE: un-comment when this script is fixed for the new MyST docs.
+      # python util/add_translation_notes.py
     - run: pytest
     - uses: shimataro/ssh-key-action@v2
       with:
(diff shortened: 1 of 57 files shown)


jpmckinney:
Aha, I think I need to give you Maintain privileges.

jpmckinney:
Also supersedes #1197

jpmckinney:
*Comment moved to https://github.com/open-contracting/standard/issues/1190#issuecomment-778335978*

choldgraf:
@jpmckinney shall we add those to-do items to the meta issue here? https://github.com/open-contracting/standard/issues/1190

Or do you want them addressed before merging this PR?  (I think my preference is to merge-and-iterate, so long as there are no bugs or errors being introduced, so that we don't overly complicate each PR).

The other thing we should do is a bit of triage on those issues - which are in-scope for what I'm doing, which are more important than others ,etc.

jpmckinney:
@choldgraf Indeed - I've moved the comment to #1190; some of the tasks are notes to myself to cover (e.g. search). I'll review this PR with an aim to getting it merged without including further tasks.

jpmckinney:
🎉

choldgraf:
Woohoo! Step 1 🙂

Issue #1180, opened 2021-01-23
partyRole codelist: add wholesaleBuyer


JachymHercher:
When discussing `procuringEntity` in https://github.com/open-contracting/standard/issues/571, I noted that several roles in contracting processes that can be marked in the EU's eForms and that seem useful, but that cannot currently be expressed in OCDS. 

### Introductory analysis 
(from https://github.com/open-contracting/standard/issues/571#issuecomment-758726651 and https://github.com/open-contracting/standard/issues/571#issuecomment-762291881 and https://github.com/open-contracting/standard/issues/571#issuecomment-763136799)

> Concerning 'procuringEntity', a good starting point is to review the definitions (OCDS and Art. 2 of Directive 2014/24/EU): 
>    - 'procuringEntity': "The entity managing the contracting process." 
>    - (17) ‘procurement service provider’ (PSP) means a public or private body which offers ancillary purchasing activities on the market;
>    - (16) ‘central purchasing body’ (CPB) means a contracting authority providing centralised purchasing activities and, possibly, ancillary purchasing activities; 
>    -  (15) ‘ancillary purchasing activities’ means activities consisting in the provision of support to purchasing activities, in particular in the following forms: 
>          - (a) technical infrastructure enabling contracting authorities to award public contracts or to conclude framework agreements for works, supplies or services;
>        -   (b) advice on the conduct or design of public procurement procedures;
>        -   (c) preparation and management of procurement procedures on behalf and for the account of the contracting authority concerned;
>    - (14) ‘centralised purchasing activities’ means activities conducted on a permanent basis, in one of the following forms:
>        -  (a) the acquisition of supplies and/or services intended for contracting authorities,
>        -  (b) the award of public contracts or the conclusion of framework agreements for works, supplies or services intended for contracting authorities;

> There might be some legal distinction between "on behalf and for the account of" (in 15) ) and "intended for" (in 14) ), but my non-legal reading would be that (14)(b) is a type (15)(c), and thus all (14)(b) CPBs are also PSPs (but not vice versa). (14)(a) CPBs stand on their own, they are simply a 'buyer' (that passes on its purchases later on). "Managing the contracting process" (in the definition of 'procuringEntity') is part of (15)(c), which implies that a (14)(b) CPB is always a 'procuringEntity', while a PSP is a 'procuringEntity' under (15)(c) but not under (15)(a) and (15)(b). 

[...]

Issue #1157, opened 2021-01-06
Review additions to open codelists by extensions


jpmckinney:
After review, the new codes can be merged into the standard itself, including:

* [x] ocds_enquiry_extension: [partyRole.csv](https://github.com/open-contracting-extensions/ocds_enquiry_extension/blob/master/codelists/%2BpartyRole.csv)
* [x] ocds_legalBasis_extension: [itemClassificationScheme.csv](https://github.com/open-contracting-extensions/ocds_legalBasis_extension/blob/master/codelists/%2BitemClassificationScheme.csv)
  * #1210
  * https://github.com/open-contracting-extensions/ocds_legalBasis_extension/pull/6
* [x] ocds_organizationClassification_extension: [itemClassificationScheme.csv](https://github.com/open-contracting-extensions/ocds_organizationClassification_extension/blob/master/codelists/%2BitemClassificationScheme.csv) ('COFOG' only)
  * #1218
  * https://github.com/open-contracting-extensions/ocds_organizationClassification_extension/pull/3
* [x] ocds_tariffs_extension: [documentType.csv](https://github.com/open-contracting-extensions/ocds_tariffs_extension/blob/master/codelists/%2BdocumentType.csv)
  * https://github.com/open-contracting/standard/pull/1219
  * https://github.com/open-contracting-extensions/ocds_tariffs_extension/pull/30
* [x] [ocds_eu_extension](https://github.com/open-contracting-extensions/ocds_eu_extension/tree/master/codelists)
  * #1225
  * https://github.com/open-contracting-extensions/ocds_eu_extension/pull/10
* [x] [ocds_ppp_extension](https://github.com/open-contracting-extensions/ocds_ppp_extension/tree/master/codelists)
  * #1243

Included for completeness, but not to be merged:

* ocds_procurementMethodRationaleClassifications_extension: [itemClassificationScheme.csv](https://github.com/open-contracting-extensions/ocds_procurementMethodRationaleClassifications_extension/blob/master/codelists/%2BitemClassificationScheme.csv) (TED-specific)
* ocds_tenderClassification_extension: [itemClassificationScheme.csv](https://github.com/open-contracting-extensions/ocds_tenderClassification_extension/blob/master/codelists/%2BitemClassificationScheme.csv) (TED-specific)

ColinMaudry:
@jpmckinney  When they are merged into the standard, the codelists are also removed from the extensions, right?

jpmckinney:
[...]

PR #1225, opened 2021-02-23
Add the codes from EU extension

ColinMaudry:
#1157

documentTypes:

- [x] economicSelectionCriteria
- [x] technicalSelectionCriteria

milestoneType:

- [x] securityClearanceDeadline

partyRole:

- [x] mediationBody
- [x] processContactPoint
- [x] reviewContactPoint
- [x] selectedParticipant
- [x] informationService

Edit: But not centralPurchasingBody , discussed below https://github.com/open-contracting/standard/pull/1225#discussion_r585206371


docs/history/changelog.md
```diff
@@ -42,16 +42,26 @@ Per the [normative and non-normative content and changes policy](https://docs.go
...
   * [#1163](https://github.com/open-contracting/standard/pull/1163) Improve the description of 'procuringEntity'
   * [#1182](https://github.com/open-contracting/standard/pull/1182) Update the description of 'buyer'
   * [#1182](https://github.com/open-contracting/standard/pull/1182) Deprecate 'funder'
   * [#1187](https://github.com/open-contracting/standard/pull/1187) Add 'contractImplementationManager'
+  * [#1225](https://github.com/open-contracting/standard/pull/1225) Add codes from the EU extension:
+    * 'informationService'
+    * 'mediationBody'
+    * 'processContactPoint'
+    * 'reviewContactPoint'
+    * 'selectedParticipant'
```
jpmckinney: It also adds centralPurchasingBody

docs/history/changelog.md
```diff
@@ -42,16 +42,26 @@ Per the [normative and non-normative content and changes policy](https://docs.go
...
 * [#1201](https://github.com/open-contracting/standard/pull/1201) Add 'terminatedEarly' and 'terminatedSuccessfully' codes to the `contractStatus` codelist, to distinguish between successful completion and early termination of the contract.
 * [#1209](https://github.com/open-contracting/standard/pull/1209) Replace "electronic goods" with "digital goods" in the description of 'goods' from the `procurementCategory` codelist, to align with the description in the World Trade Organization's Agreement on Government Procurement, and to avoid confusion between electronic goods like computers and digital goods like software.
 * [#1210](https://github.com/open-contracting/standard/pull/1210) Add the 'CELEX' and 'LEXML' codes from the legal basis extension to the `itemClassificationScheme` codelist.
 * [#1218](https://github.com/open-contracting/standard/pull/1218) Add the 'COFOG' code from the organization classifiation extension to the `itemClassificationScheme` codelist.
+* [#1225](https://github.com/open-contracting/standard/pull/1225) Add 'securityClearanceDate' code from the EU extension to the `milestoneType`codelist
```
```suggestion
* [#1225](https://github.com/open-contracting/standard/pull/1225) Add 'securityClearanceDeadline' code from the EU extension to the `milestoneType` codelist.
```

schema/codelists/partyRole.csv
```diff
@@ -10,3 +10,9 @@ payee,Payee,A party in receipt of a payment from a transaction.,
 reviewBody,Review body,A party responsible for the review of this procurement process. This party often has a role in any challenges made to the contract award.,
 interestedParty,Interested party,"A party that has expressed an interest in the contracting process: for example, by purchasing tender documents or submitting clarification questions.",
 contractImplementationManager,Contract implementation manager,The organization responsible for managing the implementation of the contract on behalf of the buyer. (This may be different from the procuring entity that manages the contracting process.),
+mediationBody,Mediation body,The body responsible for mediation procedures.,
+centralPurchasingBody,Central purchasing body,"The procuring entity providing centralized purchasing activities and, possibly, ancillary purchasing activities.",
```
jpmckinney: @JachymHercher This was added to the EU extension for I.2 "The contract is awarded by a central purchasing body" from the current EU forms. With respect to #1180, does it make sense to use this code? Does it match any of the 4 codes proposed in #1180?

schema/codelists/partyRole.csv
```diff
@@ -10,3 +10,9 @@ payee,Payee,A party in receipt of a payment from a transaction.,
...
+processContactPoint,Process contact point,A contact point dedicated to this contracting process.,
+reviewContactPoint,Review contact point,The service from which information about the review procedure can be obtained.,
+selectedParticipant,Selected participant,A party that has already been selected to participate in the design contest.,
```
jpmckinney: This was added to the EU extension for IV.1.7 "Names of participants already selected (in the case of a restricted contest)" from the current EU forms.

@ColinMaudry Should this instead be modelled as a Bid with a status of 'invited'? https://extensions.open-contracting.org/en/extensions/bids/master/codelists/#bidStatus.csv

schema/codelists/partyRole.csv
JachymHercher: 'centralPurchasingBody' covers both 'wholesaler' and 'contract intermediary' from https://github.com/open-contracting/standard/issues/1180#issue-792561618. In eForms, the codes only exist in their split form (see [cpb-acq, cpb-awa](https://op.europa.eu/en/web/eu-vocabularies/concept-scheme/-/resource?uri=http://publications.europa.eu/resource/authority/organisation-role)), but in the TED 2.0.9 forms they only exist as one code. To be able to map to both, we would either need to include all three codes, or decide that for TED 2.0.9, a CPB should be marked by including both the 'wholesaler' and 'contract intermediary'. The latter option avoids clutter in OCDS, but might be less obvious to users (not sure how many there are for TED 2.0.9 though).

schema/codelists/partyRole.csv
ColinMaudry: Good idea, but I see several issues:

- That would be a bid with only `status` and `tenderers`, no `date` and no `value`.
- At the time of publication of this design contest notice, we don't know whether the selected/invited participant actually submitted a bid. They may be selected/invited but not participate. @JachymHercher Could you please confirm this point?
- Only the "selected participants" would appear in bids, as if they were the only bidders (F13 Results of design contest only discloses statistics about participants, not details)

jpmckinney: Yes, there are two issues: (1) how to model incomplete data, and (2) how to model invitations to bid.

(1) From this perspective, it's no problem for a bid to lack a `date` and `value`, and for `bids` to only contain invited applicants. A more complete dataset would add/update such values.
(2) However, maybe there should be no such status as 'invited', because an invitation to bid is not the same thing as a bid, and therefore they should be modelled separately.

We should move this to an issue to discuss.

jpmckinney: Actually, there is already an issue: https://github.com/open-contracting/ocds-extensions/issues/141

schema/codelists/partyRole.csv
jpmckinney: What was the use case for splitting CPB into two codes? What need does it satisfy? Is it just to be maximally accurate from the publisher's perspective, or does it have analytical value?

jpmckinney: Noting that we can also leave/restore 'centralPurchasingBody' in the EU extension until we resolve #1180.

jpmckinney:
Noting that we should restore 'selectedParticipant' to the EU extension until #1245 is resolved.

schema/codelists/partyRole.csv
ColinMaudry: For now 'selectedParticipant' remains in the EU extension.

schema/codelists/partyRole.csv
ColinMaudry: I have removed 'centralPurchasingBody' from the PR.

JachymHercher: > What was the use case for splitting CPB into two codes? What need does it satisfy? Is it just to be maximally accurate from the publisher's perspective, or does it have analytical value?

Primarily policymaking: wholesale and intermediary are two very different ways of doing/organizing central purchasing and being able to distinguish them is the first step to knowing how they are used and, ultimately, which one is better in which contexts (e.g. for which types of purchases). Speculatively, wholesale/intermediary might also turn out to be a "shorthand" way of communicating certain information to the market such as single/multiple place of delivery, single/multiple payments (and the related delays in payments), etc. These can always be specified in more detail elsewhere of course.

schema/codelists/partyRole.csv
ColinMaudry: Done in https://github.com/open-contracting-extensions/ocds_eu_extension/pull/11

schema/codelists/milestoneType.csv
```diff
@@ -7,3 +7,4 @@ delivery,Delivery milestones,"For delivery milestones, such as the date when a g
 reporting,Reporting milestones,"For reporting milestones, such as when key reports are to be provided."
 financing,Financing milestones,"For events related to the funding of the contracting process. In public-private partnerships, such events might include, for example, planned payments and equity transfers."
 payment,Payment milestones,"For payment milestones, such as the date when a payment is expected to be made to a supplier. If the `status` of a contract's payment milestone is 'met', then the contract ought to have an entry for the payment under its `implementation.transactions`."
+securityClearanceDeadline,Security clearance deadline,The tenderers may obtain their security clearance until this date.
```
jpmckinney:
```suggestion
securityClearanceDeadline,Security clearance deadline,The tenderers can obtain their security clearance until this date.
```

jpmckinney (approved):

PR #1231, opened 2021-02-28
Update description of supplier

### LINKED ISSUE #1149: Supplier - definition ("awarded or contracted")
In https://github.com/open-contracting/standard/issues/903#issuecomment-728178107, I questioned whether the definition of 'Supplier' ("An entity awarded or contracted to provide goods, works or services.") requires both "awarded" and "contracted".

> As long as the definition includes "awarded", if a buyer awards a contract to supplier A, then supplier B goes to court, then a new award goes to supplier B, you have two suppliers (as long as the original award isn't cancelled, but just doesn't lead to a contract). I'm not sure this is ideal.

> On the other hand, since OCDS only has 'suppliers' in Award, perhaps "awarded" needs to stay as the centrepiece. However, it might be clearer if "contracted" stems out of "awarded", something like "An entity to which the buyer or procuring entity has decided to award a contract".

To clarify, the point of the first paragrah was that if we mention "awarded", then non-intuitive things can sometimes happen. The point of the second paragraph was that maybe having "awarded" is unavoidable, because in OCDS, suppliers are referenced in `Award`, not `Contract`. However, in that case, let's still avoid having two verbs at the same level ("awarded or contracted") and instead have only one verb ("awarded"). Since "awarded" as the only verb might be unclear (e.g. if the definition was "An awarded entity" or "An entity listed in an award"), we should also mention a contract (but not as a verb of its own, but as an object of the main "awarded" verb) and for readability purposes also the subject that awards the contract. Thus the proposed: "An entity to which the buyer or procuring entity has decided to award a contract".

However, thinking about it again and looking at the definition of "Award" in https://github.com/open-contracting/standard/issues/895#issuecomment-720684175 (including https://github.com/open-contracting/standard/issues/895#issuecomment-747835840), I would actually propose to avoid both verbs - "awarding" and "contracting". Instead, let's use a definition reusing the terms from the definition of an award: "An entity with which a buyer or a procuring entity wishes to conclude a contract."

JachymHercher:
Accidentally created from 1.1 instead of 1.2, closed and replaced by #1232


JachymHercher:
Accidentally created from 1.1 instead of 1.2, closed and replaced by https://github.com/open-contracting/standard/pull/1232

PR #1115, opened 2020-11-11
Add guidance on when having multiple suppliers per award

yolile:
closes #912

PR #1232, opened 2021-02-28
Update supplier description (1149)

### LINKED ISSUE #1149: Supplier - definition ("awarded or contracted")

JachymHercher:
closes #1149


JachymHercher:
Test probaby failed because the codelist does not correctly work with the "deprecated" column. Not sure why though - I thought that was merged into 1.2 through https://github.com/open-contracting/standard/pull/1182? Should be easy to fix when merging.

schema/codelists/partyRole.csv
```diff
@@ -1,7 +1,7 @@
 Code,Title,Description,Deprecated
 buyer,Buyer,"The organization aiming to conclude a contract with a supplier or to use the goods, works or services resulting from the contract.",
 procuringEntity,Procuring entity,"The organization managing the contracting process. If an organization is both a buyer and a procuring entity (as can be the case in simple contracting processes), its roles should include 'buyer', but not 'procuringEntity'.",
-supplier,Supplier,"An entity awarded or contracted to provide goods, works or services.",
+supplier,Supplier,"An entity with which a buyer or a procuring entity decided to conclude a contract.",
```
jpmckinney: It's a bit silly, but the tests require quotation marks to be omitted unless the cell contains a comma.

```suggestion
supplier,Supplier,An entity with which a buyer or a procuring entity decided to conclude a contract.,
```

jpmckinney (commented):
What about the `awards/suppliers` field? Right now it starts with the circular "The suppliers awarded this award".

JachymHercher:
I overlooked it, sorry. The same definition added, just in plural. 

I removed the explanation that was there, as it partially repeats the definition of award ("Depending on the jurisdiction, a single decision may concern a single supplier, item and value; or batches of suppliers and/or items and/or values. [...] As far as possible, the award should be published at the most granular level (i.e. a given supplier will deliver a given item at a given value)" and I think we should rather explain usage in the guidance mentioned in the last paragraph of https://github.com/open-contracting/standard/issues/896#issuecomment-774772868.

docs/history/changelog.md
```diff
@@ -65,6 +66,8 @@ Per the [normative and non-normative content and changes policy](https://docs.go
 * [#1238](https://github.com/open-contracting/standard/pull/1238) Open the `releaseTag` codelist.
 
 ### Schema
+
+* [#1232](https://github.com/open-contracting/standard/pull/1232) Update the description of `award.suppliers`.
```
```suggestion
* [#1232](https://github.com/open-contracting/standard/pull/1232) Update the description of `awards.suppliers`.
```

jpmckinney (approved):

jpmckinney:
FYI, @yolile This PR essentially overrides #1115. Not sure how important that is.

yolile:
> FYI, @yolile This PR essentially overrides #1115. Not sure how important that is.

The import part of #1115 was to clarify when to use multiple suppliers. We don't have that here in the supplier description anymore, but that guidance will be in the Award definition, as I see in #1175, so  I guess we are good.

PR #1240, opened 2021-03-08
release-schema: Move Unit to definitions

yolile:
refs #1107


jpmckinney:
Why is the description for the definition:

> A unit of measure for an item and its unit-price.

But for the field:

> A description of the unit in which the supplies, services or works are provided (e.g. hours, kilograms) and the unit-price.

Any reason not to re-use the field description?

yolile:
@jpmckinney I wasn't sure if they need to be different or not. Should we keep them the same then?

jpmckinney:
I think OK to make identical.

yolile:
@jpmckinney done

jpmckinney (approved):

PR #1241, opened 2021-03-08
record_package_schema: fix record definition letter case

### LINKED ISSUE #641: Fix letter case of schema definitions
Change the case of the `record` definition in `record-package-schema.json` to `Record`.

(Noting that, among properties, `former_value` doesn't follow the lowerCamelCase pattern, but it's deprecated anyhow.)

yolile:
closes #641

### DIFF
--- a/docs/history/changelog.md
+++ b/docs/history/changelog.md
@@ -75,6 +75,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
   * `Contract.items`: "If the items contracted are identical to the items awarded, this field **should** be omitted." (rephrased)
 * Record package schema:
   * [#1241](https://github.com/open-contracting/standard/pull/1241) In the definition section, rename record to `Record` and add `LinkedRelease`.
+  * [#1241](https://github.com/open-contracting/standard/pull/1241) Align descriptions of `Record` `releases`, linked releases and embedded releases.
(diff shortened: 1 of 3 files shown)


yolile:
hmm now the test is failing due to:

>   schema/record-package-schema.json includes "object" in "items/type" at /definitions/Record/properties/releases/oneOf/0/items

This is because linked releases. Should I move linked releases to record package schema definition as well?

jpmckinney:
In `records_reference` there's `:pointer: /definitions/record/properties/releases/oneOf/0/items` which needs to be updated. 

Otherwise, for the lint error, we can either copy the exception that was there for `record`, or we can moved linked release to a definition to `$ref`erence.

yolile:
> or we can moved linked release to a definition to $reference.

I will move the linked release object to a definition in the record_package_schema

jpmckinney:
@yolile I added a commit, because we try to avoid referencing the JSON type in field descriptions. While fixing that for `LinkedRelease` I noticed a few related descriptions I could tidy.

yolile:
@jpmckinney looks good! should we include the descriptions updates in the changelog as well?

jpmckinney:
Good catch! I added a commit.

yolile:
@jpmckinney perfect, in that case, do you want to approve this or should I merge it directly?

jpmckinney (approved):

PR #1242, opened 2021-03-09
Release id scoped by the ocid AND the version

### LINKED ISSUE #849: Upgrade docs: Interaction with release immutability and dates
A publisher ought to be allowed to upgrade their data between OCDS versions. If they do so, we would presumably want the release `id` to remain the same; however, this would mean that there is an exception to a release's immutability.

Furthermore, per #848, they would need to either keep the release `date` values the same, or ensure that the releases have the same chronological order after an upgrade.

For dates, I propose:

* `date` fields remain the same. We might need to clarify the description of the `date` field to make it clear that upgrading a release doesn't entail its re-publication. It's currently: "The date this information was first released, or published."

For immutability, we can either:

1. Change the uniqueness of release `id` to be scoped by the `ocid` *and* the version, such that there can be two releases with the same `id` as long as they have different versions.
    * The version isn't presently declared on individual releases, but we might explore using `$schema` to indicate this in #426.
    * The merge routine would have to specify that, if two releases have the same `id`, then they should be treated as duplicates and one should be ignored (presumably the one whose version matches the fewest other releases).
    * Similar to the uniqueness of the record in #847, it's very likely that publishers might have archives of older data using older versions of OCDS. If there were a major upgrade of OCDS (2.0), publishers might also want to offer OCDS data in multiple versions.
    * Essentially, this option says that the version of OCDS doesn't matter with respect to the uniqueness of a release; each data file is just a different serialization of the abstract release.
1. Allow releases to be mutated *only if* they are being upgraded.
    * This option would preserve the idea of there only being one data file per release. However, as mentioned above, this might not be realistic, or desirable in case of a major upgrade.

I prefer something like option 1.

ColinMaudry:
Closes #849


ColinMaudry:
I still need to update the documentation of the merge routine:

> The merge routine would have to specify that, if two releases have the same id, then they should be treated as duplicates and one should be ignored (presumably the one whose version matches the fewest other releases). (https://github.com/open-contracting/standard/issues/849#issue-427113362)

ColinMaudry:
The merge routine documentation doesn't mention packages and OCDS version. It's only about merging a number of releases into a combined or versioned release. We assume the releases come from a single or multiple release packages (with a potentially different OCDS version), but it looks like this aspect is out of the scope of this chapter.  

One change we could  do is change this sentence (first section):

> Merging is the process of combining individual releases **with the same OCDS version** into a compiled or versioned release, described in more detail below. At a high level:

ColinMaudry:
Remaining:

- add a note in  **Merge routine** section 'same ocid value and same OCDS version'
- Merging is the process of combining individual releases with the same OCDS version into a compiled or versioned release, described in more detail below. At a high level:

jpmckinney:
@ColinMaudry I assume this should no longer be marked draft?

docs/history/changelog.md
```diff
@@ -75,6 +75,11 @@ Per the [normative and non-normative content and changes policy](https://docs.go
   * `Organization.id`: "This field need only be unique within the scope of the contracting process, but **should** be built with the following structure {identifier.scheme}-{identifier.id}(-{department-identifier}) where an organization identifier is available" ("may" replaced with "should")
   * `Period.durationInDays`:  "If a startDate and endDate are set, this field, if used, **must** be equal to the difference between startDate and endDate. Otherwise, if a startDate and maxExtentDate are set, this field, if used, **must** be equal to the difference between startDate and maxExtentDate." ("should" replaced with "must")
   * `Contract.items`: "If the items contracted are identical to the items awarded, this field **should** be omitted." (rephrased)
+* [#1125](https://github.com/open-contracting/standard/pull/1125) Add `weight` to `Item.unit`
+* [#1115](https://github.com/open-contracting/standard/pull/1115) Clarify when to have multiple suppliers per award
+* [#1067](https://github.com/open-contracting/standard/pull/1067) Clarify the description of `Publisher.name` to indicate that it is the organization or department responsible for publishing the OCDS version of the data.
```
jpmckinney: Appears elsewhere in changelog.

```suggestion
```

docs/history/changelog.md
```diff
@@ -75,6 +75,11 @@ Per the [normative and non-normative content and changes policy](https://docs.go
...
+* [#1242](https://github.com/open-contracting/standard/pull/1242) Expand the scope of release `id` to ocid and OCDS version (was ocid only), allowing the publication of datasets for different versions of OCDS.
```
```suggestion
* [#1242](https://github.com/open-contracting/standard/pull/1242) Narrow the uniqueness scope of a release's `id` to its `ocid` and OCDS version (was `ocid` only), to allow the publication of the same release for different versions of OCDS.
```

docs/history/changelog.md
```diff
@@ -75,6 +75,11 @@ Per the [normative and non-normative content and changes policy](https://docs.go
...
+* [#1242](https://github.com/open-contracting/standard/pull/1242) Add OCDS version on top of ocid as a merging facet.
```
```suggestion
* [#1242](https://github.com/open-contracting/standard/pull/1242) Merging: Clarify that the releases to merge must use the same version of OCDS.
```

Issue #370, opened 2016-09-15
Award: Consortium suppliers


timgdavies:
## Corsortium suppliers

This is under consideration for the 1.1 upgrade of OCDS. It builds on changes to organisation handling proposed in #368 
## The issue

Suppliers may be consortia. 

In some countries, a consortium is a legally registered organization. In others, they are simply a collection of firms.

In some countries, the Special Purpose Vehicle (SPV) for a consortium is only created after the contract is signed. 

Users may want to know who the consortium members are and their share in the consortium. 

Note: The model we adopt for this may also be possible to adapt to presenting shareholder and beneficial ownership information within the standard. 
## The proposal

Adding `consortium` and `consortiumMember` to the new organization role codelist (see Slide 17).

Introducing an extension to the organization building block to provide an `ownership` field with an array of objects containingN:
- id - an identifier for this record of shareholding
- shareholder
  - id and name of shareholder in the consortium;
- share - share held
- period - the period this shareholding applied for

The exact details of this extension should be developed over the coming months  through wider consultation, with particular reference to the Global Beneficial Ownership Register work. 
### Worked example

A consortium (The Consortium) made up of three shareholders, (Mr A, AnyCorp and Any State) is awarded a contract. The abbreviated example below shows how this coulde be modelled with the consortium, and each of the parties to the consortium, listed in the `entities` array. 

The share that each consortium member holds, and the period of that holding, is given in the `ownership` array of the consortium. 

``` json
{"releases":[{
       "ocid": "ocds-213czf-1.1-upgrade",
       "id": "ocds-213czf-1.1-upgrade-consortium",
       "tag":["award"],
       "entities": [
          {
                     "id":"XI-GUID-569949c5-2df4-4d9f-81ce-bb34dc134f7e",
                     "roles": ["tenderer","supplier","consortium"],
                     "identifier": {
                         "scheme": "XI-GUID",
                         "id": "569949c5-2df4-4d9f-81ce-bb34dc134f7e",
                         "legalName": "The Consortium",
                     },
                     "name": "The Consoritum",
                     "ownership":[
                        {
                            "id":"1",
                            "shareholder":{
                              "id":"XI-GUID-61a491d2-fc8b-4a91-b6bc-7bede34e466d",
                              "name":"Mr A"
                            },
                            "share":0.1,
                            "period": {
                              "startDate":"2016-01-01T00:00:00Z"
                            }
                        },
                        {
                            "id":"2",
                            "shareholder":{
[...]

PR #1243, opened 2021-03-09
Add the milestoneType and partyRole codes from the PPP extension 

### LINKED ISSUE #1157: Review additions to open codelists by extensions

ColinMaudry:
Original lists:

- [milestoneType](https://github.com/open-contracting-extensions/ocds_ppp_extension/blob/master/codelists/%2BmilestoneType.csv)
- [partyRole](https://github.com/open-contracting-extensions/ocds_ppp_extension/blob/master/codelists/%2BpartyRole.csv)

#1157

Merge https://github.com/open-contracting-extensions/ocds_ppp_extension/pull/14 after merging this PR.


ColinMaudry:
I have discarded the content of the Source column when moving the codes to the core codelists. Is it OK?

jpmckinney:
Source for [milestoneType](https://github.com/open-contracting-extensions/ocds_ppp_extension/blob/b5466c7d2b493da05bafd1758f23d6cf8d238fb0/codelists/%2BmilestoneType.csv) only refers to GitHub issues. Source for [partyRole](https://github.com/open-contracting-extensions/ocds_ppp_extension/blob/b5466c7d2b493da05bafd1758f23d6cf8d238fb0/codelists/%2BpartyRole.csv) refers twice to https://ppp-certification.com/ppp-certification-guide/glossary. I think it's OK to remove.

schema/codelists/partyRole.csv
```diff
@@ -10,3 +10,12 @@ payee,Payee,A party in receipt of a payment from a transaction.,
 reviewBody,Review body,A party responsible for the review of this procurement process. This party often has a role in any challenges made to the contract award.,
 interestedParty,Interested party,"A party that has expressed an interest in the contracting process: for example, by purchasing tender documents or submitting clarification questions.",
 contractImplementationManager,Contract implementation manager,The organization responsible for managing the implementation of the contract on behalf of the buyer. (This may be different from the procuring entity that manages the contracting process.),
+privateParty,Private party,"The counter party of the procuring authority in the PPP contract. A private entity which has been granted the contract to construct and operate a government asset, and which is usually created under the form of a Special Purpose Vehicle or SPV.",
+leadBank,Lead bank,A lead bank is a bank that oversees the arrangement of a loan syndication,
+lender,Lender,"The APMG PPP Certification Program defines lenders as ""Institutions that provide lending to the project: mainly banks and institutional investors through project bonds""",
```
jpmckinney: Maybe we do want a source column, so that we can just use the straight definition, and reference the source by URL in the Source column only.

schema/codelists/partyRole.csv
```diff
@@ -10,3 +10,12 @@ payee,Payee,A party in receipt of a payment from a transaction.,
 reviewBody,Review body,A party responsible for the review of this procurement process. This party often has a role in any challenges made to the contract award.,
 interestedParty,Interested party,"A party that has expressed an interest in the contracting process: for example, by purchasing tender documents or submitting clarification questions.",
 contractImplementationManager,Contract implementation manager,The organization responsible for managing the implementation of the contract on behalf of the buyer. (This may be different from the procuring entity that manages the contracting process.),
+privateParty,Private party,"The counter party of the procuring authority in the PPP contract. A private entity which has been granted the contract to construct and operate a government asset, and which is usually created under the form of a Special Purpose Vehicle or SPV.",
```
jpmckinney: Expand "PPP" to "public-private partnership". Change "or SPV" to "(SPV)".

jpmckinney:
@duncandewhurst Do all the code titles/descriptions make sense in OCDS core? Also, these seem to not be used / referenced in OCDS for PPPs: consortiaMember, grantor, socialWitness, otherWitness, notary.

We had some long discussions about consortia (#370) and added this worked example https://standard.open-contracting.org/1.1/en/guidance/map/consortia/ Do we need a 'consortiaMember' code?

cc @yolile

duncandewhurst:
`milestoneType.csv`

* 'procurement' - fine to add
* 'contractClose' - 'Contract close' [seems to be a PPP-specific term](https://www.google.com/search?q=%22contract+close%22+-out+-outs+procurement&oq=%22contract+close%22+-out+-outs+procurement). In OCDS core the current title might be confused with ['contract close-out'](https://buyandsell.gc.ca/for-government/buying-for-the-government-of-canada/manage-the-contract/contract-close-out) which is about the end of the contract, rather than its signature. We could address that by changing the title to 'Commercial and financial close' and improving the description based on the definitions for commercial close and financial close from the [PPP Knowledge Lab glossary](https://pppknowledgelab.org/glossary/) and the [PPP Certification Guide glossary](https://ppp-certification.com/ppp-certification-guide/glossary). We should at least update the description to mention PPPs.

`partyRoles.csv`

* 'leadBank', 'lender' and 'equityInvestor' - I am not sure how applicable these terms are in non-PPP procurement contexts, so we should perhaps append something like ", this role typically relates to financing a public-private partnership project."
* 'consortiaMember' is sufficiently covered by the worked example and the new guidance on disclosing bidders in OCDS for PPPs added in https://github.com/open-contracting-extensions/public-private-partnerships/pull/245. As an aside, it might be good to expand the worked example to cover disclosing consortia bidders using the bids extension.
* 'grantor', 'socialWitness', 'otherWitness' and 'notary' were all roles identified for disclosure in Red Compartida (the pilot project for OCDS for PPPs), although they are not part of the World Bank PPP framework itself. They are all applicable to non-PPP procurement contexts, so I think these are fine to add with the possible exception of 'notary', based on https://github.com/open-contracting/standard/issues/883

schema/codelists/partyRole.csv
ColinMaudry: Sourcing is always a good idea, but we must think of how it will look once displayed in a [700px-wide table](https://standard.open-contracting.org/latest/en/schema/codelists/#party-role) (which is not wide). That would look good with a [Source](https://ppp-certification.com/ppp-certification-guide/glossary) link in the source column. Or just the domain name: [ppp-certification.com](https://ppp-certification.com/ppp-certification-guide/glossary).

ColinMaudry: I add the Source column in partyRole.csv, and @jpmckinney takes care of making the table look good :)

jpmckinney: Once you add it back, you can check how it looks here: https://standard.open-contracting.org/staging/1157_addPPPCodes/en/schema/codelists/#party-role

schema/codelists/partyRole.csv
ColinMaudry: Done

ColinMaudry:
To do:

- [x] Change the title and improve the decription of milestoneType.contractClose
- [x] Mention PPP in the description of partyRole.leadBank
- [ ] ~~Create an issue in ppp extension to "expand the worked example to cover disclosing consortia bidders using the bids extension" https://github.com/open-contracting/standard/issues/1262~~
- [x] Decide whether we keep partyRole.notary based on #883

ColinMaudry:
For 'notary', according to #883, it would be considered a party if it doesn't act on behalf of another party. Does it?

Another question: why would we want to know who the notary is? What problem does it solve? I don't know, as I don't know what role notaries play in contracting processes. @JachymHercher, any opinion on this?

ColinMaudry:
@duncandewhurst Here is my take to an improved description of 'contractClose':

> The point at which all the significant commercial issues between the procuring entity (buyer?) and the supplier have been agreed, the contract has been signed, the pre-conditions attached to the financing have been met or waived, and the project funding is available.

duncandewhurst:
> @duncandewhurst Here is my take to an improved description of 'contractClose':

Thanks @ColinMaudry! I don't think the description should be about a point in time since there might be many milestones related to the commercial and financial close. For example, there could be separate milestones for commercial close and for financial close, and there could also be more detailed milestones for individual financing conditions.

Following the pattern of the other codes in `milestoneType.csv`, I would go with something like:

> **Commercial and financial close**
> For events relating to agreeing commercial issues between the parties to a public-private partnership contract, executing contract documentation, and meeting any pre-conditions relating to the project's financing.

Regarding:

>  * [ ]  Create an issue in ppp extension to "expand the worked example to cover disclosing consortia bidders using the bids extension"

I wasn't clear, sorry. I meant the [consortia worked example](https://standard.open-contracting.org/1.1/en/guidance/map/consortia/) in the main OCDS docs since consortia bidders are already covered in the PPPs example.

> For 'notary', according to #883, it would be considered a party if it doesn't act on behalf of another party. Does it?
> 
> Another question: why would we want to know who the notary is? What problem does it solve? I don't know, as I don't know what role notaries play in contracting processes. @JachymHercher, any opinion on this?

I'm not sure if the notary acts on behalf of a particular party. My understanding of the role of a notary in general is that they are present for the signing of a contract in order to validate that the people signing the contract are the same people named on the contract. However, in the specific case of Mexico, the [OECD](http://www.oecd.org/gov/ethics/Corruption-Public-Procurement-Brochure.pdf) mentions that:

> Once the public contracts are signed, they are posted on the Commission’s webpage the same day, and a public notary certifies that the official version of the contract that has been signed is the same than (sic) the one that was posted online.

I don't know of any use cases which depend on knowing who the notary is, but there was a desire to disclose the notary for Red Compartida.

I'm happy to omit 'notary' until we have an opportunity to clarify its semantics and use cases.

ColinMaudry:
I'm not sure what would be the cons of adding 'notary', and I assume that Red Compartida still needs it, so I suggest we also add it in core.

schema/codelists/partyRole.csv
```diff
@@ -1,16 +1,25 @@
...
+lender,Lender,"The APMG PPP Certification Program defines lenders as ""Institutions that provide lending to the project: mainly banks and institutional investors through project bonds""",,
+equityInvestor,Equity investor,"The APMG PPP Certification Program defines equity investors as ""Investors who finance a portion of the project’s capital expenditures (Capex), typically as share capital or subordinated debt""",,
+consortiaMember,Consortia member,One of a group of companies that has joined together to bid for a project.,,
+grantor,Grantor,"The party which grants a concession, a license or some other right.",https://ppp-certification.com/ppp-certification-guide/glossary,
+socialWitness,Social witness,A social witness participates as an external observer in a public procurements process. They have a formal role as an observer.,,
+otherWitness,Other witness,A witness attends one or more stages of the procurement process to observe proceedings.,,
+notary,Notary,"A person authorized to perform certain legal formalities, especially to certify documents and proceedings.",,
```
jpmckinney: 'notary' has very specific semantics in Red Compartida https://github.com/open-contracting/standard/pull/1243#issuecomment-808970349, which was a one-time disclosure in 2017 that has not been/will not be updated.

This is an open codelist, and a code that only one publisher uses is the perfect case for a local code. We therefore omit it until there is clear evidence for standardization.

```suggestion
```

schema/codelists/partyRole.csv
```diff
@@ -1,16 +1,25 @@
...
+leadBank,Lead bank,A lead bank is a bank that oversees the arrangement of a loan syndication. This role typically relates to financing a public-private partnership project.,,
+lender,Lender,"The APMG PPP Certification Program defines lenders as ""Institutions that provide lending to the project: mainly banks and institutional investors through project bonds""",,
+equityInvestor,Equity investor,"The APMG PPP Certification Program defines equity investors as ""Investors who finance a portion of the project’s capital expenditures (Capex), typically as share capital or subordinated debt""",,
+consortiaMember,Consortia member,One of a group of companies that has joined together to bid for a project.,,
+grantor,Grantor,"The party which grants a concession, a license or some other right.",https://ppp-certification.com/ppp-certification-guide/glossary,
+socialWitness,Social witness,A social witness participates as an external observer in a public procurements process. They have a formal role as an observer.,,
```
jpmckinney: I can only find "social witness" in use in Mexico. A concept needs to be standardized across jurisdictions. As such, this ought to be omitted.

```suggestion
```

schema/codelists/partyRole.csv
```diff
@@ -1,16 +1,25 @@
...
+otherWitness,Other witness,A witness attends one or more stages of the procurement process to observe proceedings.,,
```
jpmckinney: These semantics are too loose, and this code does not make sense in the absence of 'socialWitness', so it ought to be omitted.

```suggestion
```

schema/codelists/partyRole.csv
```diff
@@ -1,16 +1,25 @@
...
+leadBank,Lead bank,A lead bank is a bank that oversees the arrangement of a loan syndication. This role typically relates to financing a public-private partnership project.,,
+lender,Lender,"The APMG PPP Certification Program defines lenders as ""Institutions that provide lending to the project: mainly banks and institutional investors through project bonds""",,
+equityInvestor,Equity investor,"The APMG PPP Certification Program defines equity investors as ""Investors who finance a portion of the project’s capital expenditures (Capex), typically as share capital or subordinated debt""",,
+consortiaMember,Consortia member,One of a group of companies that has joined together to bid for a project.,,
+grantor,Grantor,"The party which grants a concession, a license or some other right.",https://ppp-certification.com/ppp-certification-guide/glossary,
```
jpmckinney: The definition at the link ends with "Please see procuring authority." which is defined as:

> The unit/body/department within a government that is tendering and contracting the project. The public counterpart in the PPP contract. This is usually the same unit or body that promotes the project (the public promoter). For example, the ministry or department of transportation, the ministry of finance, and so on.  **It also includes “public party”, “public partner”, “public authority”, and “grantor.”**

As such, this is not different from the buyer or procuring entity. It is simply a sub-role within the APMG glossary.

```suggestion
```

schema/codelists/partyRole.csv
```diff
@@ -1,16 +1,25 @@
...
+privateParty,Private party,"The counter party of the procuring authority in the public-private partnership (PPP) contract. A private entity which has been granted the contract to construct and operate a government asset, and which is usually created under the form of a Special Purpose Vehicle (SPV).",https://ppp-certification.com/ppp-certification-guide/glossary,
+leadBank,Lead bank,A lead bank is a bank that oversees the arrangement of a loan syndication. This role typically relates to financing a public-private partnership project.,,
+lender,Lender,"The APMG PPP Certification Program defines lenders as ""Institutions that provide lending to the project: mainly banks and institutional investors through project bonds""",,
+equityInvestor,Equity investor,"The APMG PPP Certification Program defines equity investors as ""Investors who finance a portion of the project’s capital expenditures (Capex), typically as share capital or subordinated debt""",,
+consortiaMember,Consortia member,One of a group of companies that has joined together to bid for a project.,,
```
jpmckinney: @duncandewhurst had written:

> 'consortiaMember' is sufficiently covered by the worked example

So, we remove it.

```suggestion
```

docs/history/changelog.md
```diff
@@ -50,6 +50,16 @@ Per the [normative and non-normative content and changes policy](https://docs.go
...
+  * [#1243](https://github.com/open-contracting/standard/pull/1243) Add the codes from the PPP extension:
+    * 'privateParty'
+    * 'leadBank'
+    * 'lender'
+    * 'equityInvestor'
+    * 'consortiaMember'
+    * 'grantor'
+    * 'socialWitness'
+    * 'otherWitness'
+    * 'notary'
```
```suggestion
    * 'privateParty'
    * 'leadBank'
    * 'lender'
    * 'equityInvestor'
```

schema/codelists/partyRole.csv
```diff
@@ -1,16 +1,25 @@
...
+contractImplementationManager,Contract implementation manager,The organization responsible for managing the implementation of the contract on behalf of the buyer. (This may be different from the procuring entity that manages the contracting process.),,
+mediationBody,Mediation body,The body responsible for mediation procedures.,,
+processContactPoint,Process contact point,A contact point dedicated to this contracting process.,,
+reviewContactPoint,Review contact point,The service from which information about the review procedure can be obtained.,,
+informationService,Information service,An organization that can provide information regarding specific aspects of the contracting process.,,
+privateParty,Private party,"The counter party of the procuring authority in the public-private partnership (PPP) contract. A private entity which has been granted the contract to construct and operate a government asset, and which is usually created under the form of a Special Purpose Vehicle (SPV).",https://ppp-certification.com/ppp-certification-guide/glossary,
```
jpmckinney:
```suggestion
privateParty,Private party,"The counter party of the procuring authority in the public-private partnership (PPP) contract. A private entity which has been granted the contract to construct and operate a government asset, and which is usually created under the form of a special purpose vehicle (SPV).",https://ppp-certification.com/ppp-certification-guide/glossary,
```

schema/codelists/partyRole.csv
```diff
@@ -1,16 +1,25 @@
...
+leadBank,Lead bank,A lead bank is a bank that oversees the arrangement of a loan syndication. This role typically relates to financing a public-private partnership project.,,
+lender,Lender,"The APMG PPP Certification Program defines lenders as ""Institutions that provide lending to the project: mainly banks and institutional investors through project bonds""",,
+equityInvestor,Equity investor,"The APMG PPP Certification Program defines equity investors as ""Investors who finance a portion of the project’s capital expenditures (Capex), typically as share capital or subordinated debt""",,
```
jpmckinney: These codes (unlike the others) are explicitly mentioned in the PPPs guidance, and have distinct semantics in the APMG glossary.

A more general role would be 'financier' or similar. However, my understanding is that there are use cases to distinguish financiers by the category of financing they provide. Some institutions offer both categories, in which case I guess they are assigned both codes.

(In this weren't an important use case, then the [finance](https://extensions.open-contracting.org/en/extensions/finance/master/schema/) extension does have a `financeCategory` field to indicate whether the financing is via debt, grant, equity investment, etc. and a `financingParty` field to refer to the financier.)

The APMG glossary has two other roles/sub-roles: co-lenders and secondary investors. I think co-lender might be implied when there is a 'leadBank'. Secondary investing occurs after financial close, and might not be tracked in the lifecycle of a contracting process.

We typically use `Source` to acknowledge the source instead of putting this in the `Description`. These two descriptions also need to be edited to not be project-based, and to append @duncandewhurst's suggestion.

Also, besides APMG, I can't find any content that describes an equity investor as financing via subordinated debt. The SEC specifically says subordinated debt is not equity: https://www.sec.gov/interps/account/sabcodet4.htm

```suggestion
lender,Lender,"An institution that finances the contracting process through debt financing, like traditional debt or project bonds. This role typically relates to financing a public-private partnership project.",https://ppp-certification.com/ppp-certification-guide/glossary,
equityInvestor,Equity investor,"An investor that finances the contracting process through equity financing: that is, capital in exchange for a share of ownership. This role typically relates to financing a public-private partnership project.",https://ppp-certification.com/ppp-certification-guide/glossary,
```

schema/codelists/partyRole.csv
```diff
@@ -1,16 +1,25 @@
...
+mediationBody,Mediation body,The body responsible for mediation procedures.,,
+processContactPoint,Process contact point,A contact point dedicated to this contracting process.,,
+reviewContactPoint,Review contact point,The service from which information about the review procedure can be obtained.,,
+informationService,Information service,An organization that can provide information regarding specific aspects of the contracting process.,,
+privateParty,Private party,"The counter party of the procuring authority in the public-private partnership (PPP) contract. A private entity which has been granted the contract to construct and operate a government asset, and which is usually created under the form of a Special Purpose Vehicle (SPV).",https://ppp-certification.com/ppp-certification-guide/glossary,
+leadBank,Lead bank,A lead bank is a bank that oversees the arrangement of a loan syndication. This role typically relates to financing a public-private partnership project.,,
```
jpmckinney:
```suggestion
leadBank,Lead bank,A lead bank is a bank that oversees the arrangement of a loan syndication. This role typically relates to financing a public-private partnership project.,https://www.investopedia.com/terms/l/lead-bank.asp,
```

schema/codelists/milestoneType.csv
```diff
@@ -7,4 +7,6 @@ delivery,Delivery milestones,"For delivery milestones, such as the date when a g
 reporting,Reporting milestones,"For reporting milestones, such as when key reports are to be provided."
 financing,Financing milestones,"For events related to the funding of the contracting process. In public-private partnerships, such events might include, for example, planned payments and equity transfers."
 payment,Payment milestones,"For payment milestones, such as the date when a payment is expected to be made to a supplier. If the `status` of a contract's payment milestone is 'met', then the contract ought to have an entry for the payment under its `implementation.transactions`."
+procurement,Procurement,"Events taking place during the procurement that are not captured by other codes, such as additional enquiry periods or draft tender submissions in a competitive dialogue process"
```
jpmckinney: It could be codes (like in this list) or fields (like tender period, etc.).

```suggestion
procurement,Procurement,"For events taking place during the procurement that are not captured elsewhere, such as additional enquiry periods or draft tender submissions in a competitive dialogue process."
```

jpmckinney:
@ColinMaudry I fixed the CI.

schema/codelists/partyRole.csv
jpmckinney: @ColinMaudry This is the last suggestion to resolve. By the way, you can use the "Add suggestion to batch" button to commit multiple suggestions at once.

jpmckinney (approved):

yolile:
> 'consortiaMember' is sufficiently covered by the worked example and the new guidance on disclosing bidders in OCDS for PPPs added in https://github.com/open-contracting-extensions/public-private-partnerships/pull/245. As an aside, it might be good to expand the worked example to cover disclosing consortia bidders using the bids extension.

Is it? We can also have multiple suppliers in framework agreements, and if the publisher doesn't use the bids extension the user would need to check for the `relatedProcesses` array to discard this as a framework agreement and assume it is a consortium. Isn't it clearer to add the party role instead? cc @Camilamila

jpmckinney:
@yolile Can you clarify your example? I'm not clear why 'consortiaMember' is relevant to determining whether a contracting process (or lot) involves a framework agreement.

yolile:
The use case to identify if the process was awarded to a consortia. 
If the user checks if len(awards/suppliers) > 1 only, then they might be considering framework agreement suppliers, so we need an additional condition to be sure.
To achieve that, we could either discard that the process is a framework agreement or add more conditions to ensure this is a consortium (without relying on the presence of bids)

jpmckinney:
We already have several ways to identify FAs (techniques extension, relatedProcesses array, procurementMethodDetails on a case-by-case basis). If a publisher hasn't implemented any of these available mappings, I'm not sure that they would take the extra step to add a 'consortiaMember' mapping 🤔 Do we have a specific publisher in mind?

yolile:
> We already have several ways to identify FAs

Aha, but if later we find another scenario where an award could have more than one supplier then the user would need to discard that case as well. 

> Do we have a specific publisher in mind?

Camila can provide more details, but this is for some updates to the red flags guide, so we are looking for a standardized approach for measuring this.

jpmckinney:
It would be helpful to know what the red flag is, @Camilamila.

Camilamila:
I am working on a red flag that flags a specific market if the prevalence of consortia is high, so we need to identify if the awarded supplier is a consortia (I dont think we have a current publisher disclosing this).  
Screen Shot 2024-10-21 at 11 52 28

jpmckinney:
Curious what the academic source for that flag is?

Camilamila:
"Prevalence of consortia" in [Toolkit for detecting collusive bidding in public procurement](https://www.govtransparency.eu/wp-content/uploads/2015/11/GTI_WP2014_2_Toth_et_al_150413.pdf), and [Assessing the potential for detecting collusion in Swedish public procurement](https://www.govtransparency.eu/wp-content/uploads/2021/08/Fazekas-Toth_SE_PPcartel_detection_20161115.pdf), “Two or more companies submit a joint bid, although they should be able to participate individually” in [Fraud in Public Procurement: A collection of Red Flags and Best Practices](https://ec.europa.eu/sfc/system/files/documents/sfc-files/fraud-public-procurement-final-20122017-ares20176254403.pdf).

jpmckinney:
My thinking here is:

How we model this affects a core question ("who bought what **from whom** …"). In general, we want the fields related to a core question to be (1) stable over time and (2) frequent among publications.

A new party role will not be frequent – not now (because it doesn't exist yet) and not likely in the future, either (because most publishers are unlikely to upgrade their publication with a new party role).

So, we *could* "solve" this with a party role, but, unless the role sees major adoption, then it will not solve anything for the user.

So what can we do? We can look for alternatives that match existing data and/or use existing fields. The existing alternative is that multiple suppliers means consortium, except in the case where the lot/procedure involves an FA. This can be calculated today, and increasingly in the future, as we have been good about promoting good FA modelling (and publishers seem interested in getting in right).

The only problem I know of with this alternative is:

> Aha, but if later we find another scenario where an award could have more than one supplier then the user would need to discard that case as well.

But, OCDS has been around for a decade and we have not come across a multi-supplier-but-not-consortium-and-not-FA scenario, so I think this is low likelihood (and it if occurs, it will not create many false positives for the red flag).

What do you think?

yolile:
Makes sense. For the record, the only additional (and existing) issue is when partners incorrectly map multiple suppliers to the same award (seen at least in Chile and Colombia), which could lead to false positives as well. But in the guide, we assume data is well-mapped anyway

Issue #1214, opened 2021-02-12
Tag: Open the codelist


jpmckinney:
Following from #792 and #1184, we will open the `tag` codelist.

In terms of guidance, we can recommend that a code from the codelist be used wherever possible, but that additional codes can be used based on the publisher's use cases. The additional codes can closely map the notice types (in a notice-based or paper-based system) or the event types (in an electronic records-based system).

ColinMaudry:
@jpmckinney On top of modifying the schema, does it make sense to reflect this change in the documentation?

We could update:

- https://standard.open-contracting.org/latest/en/getting_started/releases_and_records/
- https://standard.open-contracting.org/latest/en/guidance/build/easy_releases/
- https://standard.open-contracting.org/latest/en/schema/reference/

jpmckinney:
What changes would you make to those pages? Getting started is being rewritten (#1239) and guidance on tags is probably too low-level for the primer/getting started section.

ColinMaudry:
In the sections that mention release tags, II would advertise the fact that one can use their own codes to stick to their internal process.

jpmckinney:
> In the sections that mention release tags, II would advertise the fact that one can use their own codes to stick to their internal process.

I don't think tagging is fundamental to the worked example at https://standard.open-contracting.org/latest/en/guidance/build/easy_releases/, so it's not necessary to explain this there.

The new Primer (replacing Getting Started) doesn't discuss tags at all. (The Primer focuses on the concepts needed to understand the rest of the docs. Tags are a more narrow topic that can be addressed later.)

I'll review the PR to see where to include this in the reference section.

ColinMaudry:
With #1244 merged, the only guidance is in reference/codelists.md  and it mentions the fact that one can use their own tags:  https://github.com/open-contracting/standard/pull/1244/files#diff-d6958c79ed9293ec0b969c4db1ebbeeeeb6542bc6d2124e25b01769896e5b7fcR25

I can't think of another section in reference were we could mention this possibility.

PR #1244, opened 2021-03-10
Expanded guidance on additional tags

ColinMaudry:
I had forgotten to add guidance on this bit:

> In terms of guidance, we can recommend that a code from the codelist be used wherever possible, but that additional codes can be used based on the publisher's use cases. The additional codes can closely map the notice types (in a notice-based or paper-based system) or the event types (in an electronic records-based system). https://github.com/open-contracting/standard/issues/1214#issue-807607295

 #1214


schema/release-schema.json
```diff
@@ -27,7 +27,7 @@
     },
     "tag": {
       "title": "Release Tag",
-      "description": "One or more values from the open [releaseTag](https://standard.open-contracting.org/{{version}}/{{lang}}/schema/codelists/#release-tag) codelist. Tags can be used to filter releases and to understand the kind of information that releases might contain.",
+      "description": "One or more values from the open [releaseTag](https://standard.open-contracting.org/{{version}}/{{lang}}/schema/codelists/#release-tag) codelist. Tags can be used to filter releases and to understand the kind of information that releases might contain. It is encouraged to use tags from the releaseTags codelist whenever possible, but additional ones can be used based on specific use cases such as mapping with a particular notice, form or event.",
```
jpmckinney: I think we're trying to reduce publisher-specific instructions in the schema, so that it's useful to both publishers and users.

We can instead put publisher instructions at https://standard.open-contracting.org/latest/en/schema/codelists/#release-tag Note that the general policy about using existing codes is already expressed at the top of that page.

For the schema, to assist users in interpreting additional codes, we can do:

```suggestion
      "description": "One or more values from the open [releaseTag](https://standard.open-contracting.org/{{version}}/{{lang}}/schema/codelists/#release-tag) codelist. Tags can be used to filter releases and to understand the kind of information that releases might contain. Codes outside the releaseTag codelist might indicate, for example, the notice or form to which the release corresponds, or the event that triggered the publication of the release.",
```

ColinMaudry:
To do:

- [x] Put publisher instructions at https://standard.open-contracting.org/latest/en/schema/codelists/#release-tag Note that the general policy about using existing codes is already expressed at the top of that page.

schema/release-schema.json
```diff
@@ -27,7 +27,7 @@
     },
     "tag": {
       "title": "Release Tag",
-      "description": "One or more values from the open [releaseTag](https://standard.open-contracting.org/{{version}}/{{lang}}/schema/codelists/#release-tag) codelist. Tags can be used to filter releases and to understand the kind of information that releases might contain.",
+      "description": "One or more values from the open [releaseTag](https://standard.open-contracting.org/{{version}}/{{lang}}/schema/codelists/#release-tag) codelist. Tags can be used to filter releases and to understand the kind of information that releases might contain. Codes outside the releaseTag codelist might indicate, for example, the notice or form to which the release corresponds, or the event that triggered the publication of the release.",
```
jpmckinney: Clarify some of the existing content.

```suggestion
      "description": "One or more values from the open [releaseTag](https://standard.open-contracting.org/{{version}}/{{lang}}/schema/codelists/#release-tag) codelist. A tag labels the release: for example, as corresponding to a particular stage of the contracting process. Codes outside the releaseTag codelist might indicate, for example, the notice or form to which the release corresponds, or the event that triggered the publication of the release.",
```

docs/schema/codelists.md
```diff
@@ -23,7 +23,7 @@ Codes are case-sensitive, and are generally provided as English language camelCa
 
 ### Release Tag
 
-A contracting process can result in a number of releases of information over time. These must be tagged to indicate the stage of the contracting process they relate to.
+A contracting process can result in a number of releases of information over time. These must be tagged to indicate the stage of the contracting process they relate to. Additional codes can be used based on the publisher's use cases, such as the mapping of notice or event types.
```
jpmckinney: Make the new text a separate paragraph, to separate the MUST from the MAY.

Also, refer to user needs, not publisher's use cases.

```suggestion
A contracting process can result in a number of releases of information over time. These must be tagged to indicate the stage of the contracting process they relate to.

Additional codes may be used to label releases, based on user needs: for example, to indicate the notice or form to which a release corresponds.
```

jpmckinney:
@ColinMaudry Will this close #1214?

ColinMaudry:
> @ColinMaudry Will this close #1214?

In the issue you wrote:

> > In the sections that mention release tags, II would advertise the fact that one can use their own codes to stick to their internal process.

> I'll review the PR to see where to include this in the reference section.

Editing the definition in the schema and the text above the table in codelist is probably enough.

ColinMaudry:
@jpmckinney The CI fails at translating codelists to French: https://github.com/open-contracting/standard/pull/1244/checks?check_run_id=2464856812#step:6:32

jpmckinney:
@ColinMaudry I've fixed that issue now.

jpmckinney (approved):

PR #1221, opened 2021-02-17
Updating eval-rst blocks to myst directives

choldgraf:
There are `{eval-rst}` blocks scattered throughout the documentation that should now be convertible to pure MyST markdown directives. This is a PR to take a first pass at this.

PR #1247, opened 2021-03-19
build: Convert eval-rst directives

jpmckinney:
https://gist.github.com/jpmckinney/88c846651ab5411d1ed3a6fe222ce08f

### DIFF
--- a/docs/getting_started/building_blocks.md
+++ b/docs/getting_started/building_blocks.md
@@ -26,26 +26,31 @@ An OCDS document is made up of a number of sections. These are:
 
 These are represented in a JSON document as follows:
 
-```{eval-rst}
-.. code-block:: json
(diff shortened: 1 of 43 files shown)


jpmckinney:
@choldgraf Based on #1221 I wrote the above script and compared the output. After running it (and many fixes), the output was the same (except for end-of-line whitespace, additional directives converted, etc.), plus one `.. markdown::` block where I had to manually convert RST syntax to Markdown. (I'll use the script on some of our smaller docs sites.)

choldgraf:
Whoah nicely done 👍

jpmckinney:
Now we can do the fun part of updating the theme 😄

choldgraf:
haha indeed - much more enjoyable than programmatically removing eval-rst blocks.

PR #1255, opened 2021-03-29
 Recommend a length limit for tender.title 

### LINKED ISSUE #1104: tender.title: Recommend a length limit
We found when trying to use UK OCDS data that some of the titles were so long (mostly because they were listing information such as organizations in the title) that when we tried to use this for charts and exporting it, it essentially broke the system, and we had to create our own limits and cut them off at a certain point. Signifying how long a Tender Title (and other similar fields) could be would help make the data more usable.

nativaldezt:
Closes #1104

### DIFF
--- a/schema/dereferenced-release-schema.json
+++ b/schema/dereferenced-release-schema.json
@@ -1357,7 +1357,7 @@
         },
         "title": {
           "title": "Tender title",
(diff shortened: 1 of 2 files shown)


jpmckinney:
We should change the wording to:

> The title should be less than ### characters in length.

However, we need to decide an appropriate length. I'll discuss in #1104.

PR #1256, opened 2021-03-29
Recommend tax-free amounts for value.amount

### LINKED ISSUE #817: Update Value.amount description to suggest/recommend tax-free amounts
Following discussion from https://github.com/open-contracting/standard/issues/383#issuecomment-468534963 and following comments.

nativaldezt:
ref #817


jpmckinney:
This doesn't close #817. #817 requires adding `amountNet` and `amountGross` fields. Please see discussion in #817.

Once those fields are created, the guidance for `amount` is likely to be longer.

PR #1257, opened 2021-03-29
Remove 'default' for language field.

### LINKED ISSUE #1005: Remove "default": "en" for language field
This was in the schema since the initial draft (before `locale` was renamed to `language`). However, there is nothing in the standard that:

* describes English as the default language
* instructs users to take default values from the schema

I don't think the standard should use the "default" property (this is the only occurrence), as it leads to ambiguity when interpreting the data (did the publisher intentionally leave it unset with the intention of the default value being used, does the publisher not know the language and therefore cannot set it, did the publisher forget to map to this field, etc.).

https://github.com/open-contracting/standard/commit/a00ad931267de7c687f2870a495fb56ea83aa140#diff-0eaeaf6a8f844dc3725590c958f7c127R69

nativaldezt:
Closes #1005

### DIFF
--- a/docs/history/changelog.md
+++ b/docs/history/changelog.md
@@ -82,7 +82,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
   * [#1241](https://github.com/open-contracting/standard/pull/1241) Align descriptions of `Record` `releases`, linked releases and embedded releases.
 * Improve the clarity of field descriptions in the release package schema and record package schema:
   * [#1067](https://github.com/open-contracting/standard/pull/1067) `Publisher.name`, to indicate that it is the organization or department responsible for publishing the OCDS version of the data.
-* [#1257](https://github.com/open-contracting/standard/pull/1257) Remove default on `language`.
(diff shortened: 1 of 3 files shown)


jpmckinney:
Please update the changelog as documented at https://ocds-standard-development-handbook.readthedocs.io/en/latest/standard/contributing.html#proposing-changes

docs/history/changelog.md
```diff
@@ -82,6 +82,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
   * [#1241](https://github.com/open-contracting/standard/pull/1241) Align descriptions of `Record` `releases`, linked releases and embedded releases.
 * Improve the clarity of field descriptions in the release package schema and record package schema:
   * [#1067](https://github.com/open-contracting/standard/pull/1067) `Publisher.name`, to indicate that it is the organization or department responsible for publishing the OCDS version of the data.
+* [#1257](https://github.com/open-contracting/standard/pull/1257) Remove default on `language`.
```
```suggestion
* [#1257](https://github.com/open-contracting/standard/pull/1257) Remove default value of `"en"` for `language` field.
```

PR #1269, opened 2021-04-12
related_processes: Fix examples

jpmckinney:
I noticed a few things after merging #1106. I had also added a few commits before merging (I optimistically merged after checking all but the last JSON file).

- Fix open_multiple_award.json
  - Correct ocid
  - Correct tender.id
  - Remove unreferenced party
  - Add language
- Remove reference to buyer and procuring entity in last example
- Remove expand tender for file without tender section
- Expand parties.roles and relatedProcesses.relationship
- Use consistent order in JSON files


yolile:
Thank you! Looks good, I realized now that we hadn't updated the changelog either! Updated now

yolile (approved):

PR #1278, opened 2021-04-16
map: Move linked standards to map

### LINKED ISSUE #1275: Move "Linked Standards" section to Map
Currently, the "Linked Standard" section is under [Guidance->Map->Extension](https://standard.open-contracting.org/latest/en/guidance/map/extensions/#linked-standards), however, it could have its own section under Map, and we should add an admonition on the extensions page to reference this new page

yolile:
closes #1275

### DIFF
--- a/docs/guidance/map.md
+++ b/docs/guidance/map.md
@@ -126,20 +126,15 @@ map/extensions
 
 ## Linked standards
 
-OCDS data often refers to other datasets, like company registers, government budgets and infrastructure projects. These datasets contain information that is managed outside the lifecycle of a contracting process.
+Not all information that is related to a contracting process belongs in OCDS. For example, a company's annual filings and incorporation status are typically managed in a company registry, outside the lifecycle of a contracting process.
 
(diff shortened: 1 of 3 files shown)


jpmckinney:
@yolile Is this ready for review?

jpmckinney:
Hmm, this is a lot of content for the Map page itself (especially with more linked standards like BODS coming). Should we instead have a new "linked standards" page? Then, on the Map page, we can have under a Linked standards heading:

Not all information that is related to a contracting process belongs in OCDS. For example, a company's annual filings and incorporation status are typically managed in a company registry, outside the lifecycle of a contracting process.

For guidance on integrating your OCDS data with these related datasets, refer to [Linked standards]().

yolile:
> Should we instead have a new "linked standards" page?

That was my first approach but I wasn't sure if it was the best, I will change it to a new page then

yolile:
done @jpmckinney

jpmckinney (approved):

PR #1294, opened 2021-05-08
i18n: Remove Italian translation, as less than 25% translated

jpmckinney:
Per the [translation and localization policy](https://docs.google.com/document/d/1GLwWTpgDDkKmMr1hZE4D75LI8VQtFiD0VA7TF_FvY_Q/view):

> Community translations will only be linked from the documentation when the schema is fully translated, and at least 50% of other documentation pages are translated. Community translations may be removed if they appear to be abandoned, misleading, inaccurate or detrimental to the standard, by the OCDS technical team at any time.


jpmckinney:
Noting that this also involved: Removing from Transifex project. Removing from deploy repository.

PR #1296, opened 2021-05-14
Deprecate eligibilityCriteria, add exclusionGrounds

### LINKED ISSUE #901: Deprecate eligibilityCriteria and add exclusionGrounds
# Background

In OCDS, selection criteria (for example, the criteria used to qualify bidders in a selective procedure in the sense of the GPA) are often mapped to `eligibilityCriteria`.

However, selection criteria are not the same as eligibility criteria. For example, the [UNDP](https://www.undp.org/content/undp/en/home/procurement/business/qualifications.html) has qualified but ineligible bidders, e.g. due to suspension or debarment:

> … all vendors must be qualified, as well as eligible.

> Eligible vendors are qualified vendors that have not been temporarily suspended or debarred by UNDP or another UN agency, fund or programme.

`eligibilityCriteria` was added in #139, and is described as:

> A description of any eligibility criteria for potential suppliers.

At the time, `awardCriteria` was incorrectly termed `selectionCriteria`, which was fixed in #162; the confusion might have started in #140. 

# Proposal

An effective solution depends on how frequently `eligibilityCriteria` is used to disclose selection criteria (a very common concept in procurement) versus eligibility criteria (rare).

If the field is mainly used for selection criteria, then adding a new `selectionCriteria` field would create a situation in which there are two terms used to disclose the same concept.

If the field is rarely used, or only used for eligibility criteria, then we can add a new `selectionCriteria` field, and clearly explain the difference between the two.

We should also avoid the terms 'eligible' and 'eligibility' in all related guidance and extensions, unless that is actually what we mean.

ColinMaudry:
Closes  #901

### DIFF
--- a/schema/dereferenced-release-schema.json
+++ b/schema/dereferenced-release-schema.json
@@ -3090,6 +3090,18 @@
           "type": [
             "string",
             "null"
+          ],
+          "deprecated": {
+            "description": "This field is deprecated in favor of exclusionGrounds, which uses more standard terminology.",
+            "deprecatedVersion": 1.2
+          }
+        },
+        "exclusionGrounds": {
+          "title": "Exclusion grounds",
+          "description": "The criteria regarding the situation of a tenderer that can lead to its exclusion from the process. Typical exclusion grounds include bankruptcy, presence on a blacklist or failure to pay taxes.",
+          "type": [
+            "string",
+            "null"
           ]
         },
         "selectionCriteria": {
@@ -11524,6 +11536,18 @@
         "eligibilityCriteria": {
           "title": "Eligibility criteria",
           "description": "A description of any eligibility criteria for potential suppliers.",
+          "type": [
+            "string",
+            "null"
+          ],
+          "deprecated": {
+            "description": "This field is deprecated in favor of exclusionGrounds, which uses more standard terminology.",
+            "deprecatedVersion": 1.2
+          }
+        },
+        "exclusionGrounds": {
+          "title": "Exclusion grounds",
+          "description": "The criteria regarding the situation of a tenderer that can lead to its exclusion from the process. Typical exclusion grounds include bankruptcy, presence on a blacklist or failure to pay taxes.",
           "type": [
             "string",
             "null"
--- a/schema/dereferenced-release-schema.json
+++ b/schema/dereferenced-release-schema.json
@@ -3093,7 +3093,7 @@
           ],
           "deprecated": {
             "description": "This field is deprecated in favor of exclusionGrounds, which uses more standard terminology.",
-            "deprecatedVersion": 1.2
+            "deprecatedVersion": "1.2"
           }
         },
         "exclusionGrounds": {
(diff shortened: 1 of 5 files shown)


ColinMaudry:
Changelog still need to be updated.

schema/metaschema/json-schema-draft-4.json
```diff
@@ -1,149 +1,222 @@
 {
```
jpmckinney, line 1: Why is this file re-indented? I think the tests ignore it. It is an identical copy of the JSON Schema draft 4 file.

schema/release-schema.json
```diff
@@ -371,6 +371,18 @@
...
+            "null"
+          ],
+          "deprecated": {
+            "description": "This field is deprecated in favor of exclusionGrounds, which uses more standard terminology.",
+            "deprecatedVersion": "1.2"
+          }
+        },
+        "exclusionGrounds": {
+          "title": "Exclusion grounds",
+          "description": "The criteria regarding the situation of a tenderer that can lead to its exclusion from the process. Typical exclusion grounds include bankruptcy, presence on a blacklist or failure to pay taxes.",
```
jpmckinney: In #901: 

> In https://www.procurementjourney.scot/route-3/develop-documents/exclusion-selection-and-award-criteria/exclusion-criteria, bankruptcy is in the "may ask" column, thus less typical than "criminal convictions". I'll update my proposal.

I agree that criminal convictions is a more familiar example.

```suggestion
          "description": "The criteria regarding the situation of a tenderer that can lead to its exclusion from the contracting process. For example: criminal convictions, bankruptcy, presence on a blacklist or failure to pay taxes.",
```

schema/dereferenced-release-schema.json
```diff
@@ -3087,6 +3087,18 @@
         "eligibilityCriteria": {
           "title": "Eligibility criteria",
           "description": "A description of any eligibility criteria for potential suppliers.",
+          "type": [
+            "string",
+            "null"
+          ],
+          "deprecated": {
+            "description": "This field is deprecated in favor of exclusionGrounds, which uses more standard terminology.",
```
jpmckinney:
```suggestion
            "description": "This field had unclear semantics and used ambiguous terms, and is deprecated in favor of exclusionGrounds.",
```

ColinMaudry: New issue in the style guide for deprecated.description fields: https://github.com/open-contracting/standard-development-handbook/issues/241

Thanks!

schema/metaschema/json-schema-draft-4.json
ColinMaudry: Weird, I haven't opened it... I have recenlty switched from Atom to Sublime, but still I don't know what could cause this.

ColinMaudry: I guess it wasn't well indented and I did a `ocdskit indent -r .` that re-indented it.

jpmckinney:
@ColinMaudry I had accidentally commented on the dereferenced schema, so the suggested change that was committed in 3b16852 ended up being reversed when the util script was run.

ColinMaudry:
> @ColinMaudry I had accidentally commented on the dereferenced schema, so the suggested change that was committed in [3b16852](https://github.com/open-contracting/standard/commit/3b16852a60605e24b6dc7596d59dedcb73a3427b) ended up being reversed when the util script was run.

Done

jpmckinney (approved):

Issue #1084, opened 2020-10-06
Deprecate remaining package metadata and add bulk data format


jpmckinney:
There are issues with packages as discussed in open-contracting/infrastructure#89, #605, and CRM-4282 (all relevant comments reflected here), and the current packaging formats confer very few benefits.

## Benefits

The benefits of the current packaging format are:

1. A standardized way to publish multiple releases/records as a single file
1. Easy access to metadata:
    * `publisher`
    * `version` and `extensions`
    * `license` and `publicationPolicy`

A package also sets `uri` and `publishedDate`, but this is metadata about the package itself, not about the releases/records it contains.

### Metadata

* `publisher` should be moved to the release-level (see #325)
* `version` and `extensions` can be handled using `describedby` at the release-level (see #426)

Regarding `license` and `publicationPolicy`, paraphrasing open-contracting/infrastructure#89:

* License and publication policy metadata are important, but it isn't critical that they be distributed as data; that said, they can be expressed in the machine-readable description of the OCDS dataset in a data registry, using [DCAT](https://www.w3.org/TR/vocab-dcat-2/) for example (DCAT has a property for license, and a property for publication policy can be added as an extension, which [DCAT-US](https://resources.data.gov/resources/dcat-us/) does with other properties).
* Most open data (CSVs, etc.) have no means of declaring their license or publication policy, but this poses no major problem to reuse – these are instead declared on the HTML pages that serve or link to the data. Users generally only need to refer to these once, so it's not a challenge to data workflows.

See similar comments in https://github.com/open-contracting/standard/issues/325#issuecomment-445892448

As such, all metadata provided by the package can be omitted or moved to the release-level, without major issue.

### Format

We still want a standardized way to publish multiple releases/records as a single file. A minimal package in the current format with all metadata removed would be:

```json
{
  "releases": [
    // big list of releases
  ]
}
```

The problem with this format is that naive applications will load the entire file into memory. Because bulk download OCDS files can be very large (GBs), doing so exhausts memory on much consumer hardware. Iterative JSON parsers like `ijson` can be used to index to the `releases` array and yield one release at a time (as is done in OCDS Kit, for example); however, relatively few users are aware of such libraries, and many common data analysis tools don't use them (Pandas, for example). Indeed, no OCDS software written by ODS uses iterative parsing, leading to memory being exhausted in critical tools like the Data Review Tool on medium-to-large datasets; retrofitting these tools to parse iteratively is not trivial.

[...]

PR #997, opened 2020-03-26
Change history and System Architectures

romifz:

Issue #606, opened 2017-10-23
Guidance around extending subschema (building blocks)


jpmckinney:
If an extension extends the `Period` building block with a new field, because the publisher wants to add a detail or nuance to `awardPeriod`, that extension will also effectively make that new field available to all other Period objects (`tenderPeriod`, `enquiryPeriod`, `contractPeriod`, `award/contractPeriod`, `contract/period`). However, that new field may be meaningless or incoherent in those other contexts.

We should author guidance about when to extend building blocks, to avoid this scenario. In general, building blocks should only be extended if those changes make sense for all fields/objects using the building block.

Alternately, if we want to allow people to extend the `awardPeriod` object without affecting other Period objects, then we may need to abandon the idea of reusable building blocks at least as currently implemented (or severely diminish their value and create new challenges, if e.g. a publisher changes the `$ref` of `awardPeriod` to point to a new, encapsulated Period object), but that may be 2.0.

Follow-up from https://github.com/open-contracting/standard/issues/593#issuecomment-338720823

timgdavies:
I've been thinking about this a bit too, both in terms of how we maintain the schema itself, and how we manage extensions.

I can see two options that might allow us (in a 2.0 most likely, though perhaps in backwards compatible ways) to increase the flexibility of extensions. (Not suggesting these strongly at all right now - just exploring). 

**1. Factoring out each building block to it's own file, and then compiling release-schema.json from these**

Instead of a release-schema.json containing all it's definitions, you would have: 

* release-schema-template.json
* building_blocks/Period.json
* building_blocks/Party.json
* and so-on

and a build process that merges the building_blocks into release-schema-template.json **and** resolves all references.

The official OCDS schema would not then include any references. 

The main reason for considering this would be easier change management of building blocks (diffs of the full schema can be tricky to read and follow at times)

However, it would also provide the possibility of extensions patching individual building blocks **or** the compiled (and ref-resolved) release-schema.json so that an extension can choose what it is extending.

**2. Updating extensions mechanism to support applying extensions before or after refs are resolved**

This is a simpler version of the above, although it places the burden of doing the reference resolving, and creating a 'validation schema' on the validation tooling, rather than on the core OCDS creation process. 

We could likely use folder structure, or some extensions.json meta-data to decide what should be applied before or after references are resolved. 

I suspect this approach may be easier to have backwards compatible.

jpmckinney:
[...]

PR #1302, opened 2021-05-31
Primer 0.1

### LINKED ISSUE #614: Reword use cases page
The 'use cases' on this page are more like 'goals'. http://standard.open-contracting.org/latest/en/getting_started/use_cases/ My understanding of nomenclature is:

* goal: 'Value for money'
* outcome: Related to the goal of value for money, my desired outcome is 'More bidders per process', as my hypothesis is that more competition leads to lower cost to government.
* indicator: I design a methodology for periodically measuring progress against that outcome, e.g. number of bidders per process at time X compared to time Y, scoping to, e.g., procurement of medication. The same indicator may be relevant to multiple outcomes and goals.
* use case: A specific instance of using data to achieve an outcome in pursuit of a goal, which may involve the measurement of an indicator. It may be possible to generalize a use case beyond the specific context in which it was recorded.
* requirement: The specific needs to satisfy the use case, which may be needs for data, information, publication patterns, stakeholder engagement, etc.
* user need: Same as 'requirement'.
* user focus: An attention to users, which in the context of this discussion may simply mean attention to the cases where users can use OCDS to achieve outcomes in pursuit of goals.

See internal comment about this page: https://crm.open-contracting.org/issues/2217#note-21

### LINKED ISSUE #1239: Draft Primer to replace Getting Started section of documentation
As discussed, we've been working on a replacement to the current format for the Getting Started section of the documentation. To do that, I've drafted a series of pages that together form a "Primer." The primer will be organized around answering four main questions, with each having their own section within the primer, plus a landing page for the Primer to provide general orientation for users to the purpose of the primer.

What is OCDS?
How does OCDS work?
What does OCDS data look like? 
How can I implement OCDS?

Within each page, there will be a standard structure and order of elements: objectives, video inset, detail on key concepts, and further resources (OCDS documentation pages and other OCP materials).

EricReese15:
Fully updated primer based on user and Helpdesk feedback 

closes #1239, closes #614


jpmckinney:
We need to decide what (if anything) to retain from these pages (if retained, need to add to Guidance):

* https://standard.open-contracting.org/1.1/en/getting_started/building_blocks/
  * This is covered in the reference section (schema/reference) and some is covered on primer/how, though maybe some examples could be retained.
* https://standard.open-contracting.org/1.1/en/getting_started/releases_and_records/
  * This is largely covered by primer/releases_and_records, though maybe some examples could be retained. I think "easy releases" and other pages from #997 might cover the rest, but I'm not sure. The animation, I think, is not easily understandable outside helpdesk analysts who are familiar with the "block" visualization.

@yolile What do you think?

I'm not sure if/where to re-integrate the OCDS example:

````md
```{jsoninclude} ../examples/contract.json
:jsonpointer: /releases
:expand: releases, tender, awards, contracts, period, value, items, tag, parties, documents
```
````

jpmckinney:
Preview should be available here if build succeeds: https://standard.open-contracting.org/staging/primer_0.1/en/primer/

yolile:
For https://standard.open-contracting.org/1.1/en/getting_started/building_blocks/:
- We can move https://standard.open-contracting.org/1.1/en/getting_started/building_blocks/#codelists worked example to the map section, where the other worked examples are
- https://standard.open-contracting.org/1.1/en/getting_started/building_blocks/#extensions This is already covered by the specific Extensions page
- https://standard.open-contracting.org/1.1/en/getting_started/building_blocks/#building-blocks-fields could be preserved in https://standard.open-contracting.org/staging/primer_0.1/en/primer/how/#how-does-the-ocds-work after _"When mapping your data to the OCDS our sing OCDS data, you will encounter a number of common data fields and structures, which are described by the schema.. "_ section

For https://standard.open-contracting.org/1.1/en/getting_started/releases_and_records/:
- We have a complete video with examples now, so I'm ok with removing the animation. 
- https://standard.open-contracting.org/staging/primer_0.1/en/primer/releases_and_records/ doesn't mention packages and their metadata at all is that on purpose?
- I think that the releases and records examples that are part of https://standard.open-contracting.org/1.1/en/getting_started/releases_and_records/ are useful and not covered anywhere else, so we might want to keep them

jpmckinney:
Thanks! I’ll move those to new pages. Can you review the rest of the PR in the meantime?

yolile:
> Can you review the ready of the PR in the meantime?

Sure, I will!

jpmckinney:
> For https://standard.open-contracting.org/1.1/en/getting_started/building_blocks/:
> 
> * We can move https://standard.open-contracting.org/1.1/en/getting_started/building_blocks/#codelists worked example to the map section, where the other worked examples are

I don't think this example does much more than the example in https://www.open-contracting.org/resources/ocds-1-1-codelist-mapping-template-guidance/, which is what we want people to use for mapping codelists. What do you think?

> * https://standard.open-contracting.org/1.1/en/getting_started/building_blocks/#building-blocks-fields could be preserved in https://standard.open-contracting.org/staging/primer_0.1/en/primer/how/#how-does-the-ocds-work after _"When mapping your data to the OCDS our sing OCDS data, you will encounter a number of common data fields and structures, which are described by the schema.. "_ section

When preparing the Primer, Eric and I were not sure that the metaphor of "building blocks" was actually very helpful.

* We might not preserve common definitions for items, etc., because extending them can have unintended consequences #606, in which case we will not have "building blocks" in the same way.
* It's not clear that it is especially relevant for an implementer to learn that some objects follow the same schema in different parts of the data. It seems okay to learn this in the schema reference.
* I created #1347 about changing the terminology.

So, I'm okay with not having this section. What do you think?

I'll create an issue to add example data to the schema reference, since that is the main benefit of this section. #1348

> For https://standard.open-contracting.org/1.1/en/getting_started/releases_and_records/:
> 
> * https://standard.open-contracting.org/staging/primer_0.1/en/primer/releases_and_records/ doesn't mention packages and their metadata at all is that on purpose?

The old page says very little about packages (only mentioning the `uri` metadata). Since packages are going to become optional #1084, it is not important to mention them in the Primer.

> * I think that the releases and records examples that are part of https://standard.open-contracting.org/1.1/en/getting_started/releases_and_records/ are useful and not covered anywhere else, so we might want to keep them

Do you mean the two Example and Hint boxes? Isn't the Example covered by: https://standard.open-contracting.org/1.1/en/guidance/build/change_history/

For the Hint, it is a useful demonstration of the value of compiled releases. However, we know that publishers are unlikely to correctly implement the merging behavior (it's quite easy to have bugs), and we know that very few publishers offer compiled releases anyway (unless they do "easy releases"). So, I think we've accepted that the registry is the way for users to access compiled releases.

yolile:
> I don't think this example does much more than the example in https://www.open-contracting.org/resources/ocds-1-1-codelist-mapping-template-guidance/, which is what we want people to use for mapping codelists. What do you think?

True, I found this [mapping a local codelist worked example outline](https://docs.google.com/document/d/1e2V62fLBHoqcIdhdK4g3qTGq30onNmGnc9XK7i-mgUc/edit#heading=h.n0oaaum8p8f2), which means that at some point we thought that it would be useful to have that worked example, I guess. (see also the worked examples [spreadseet](https://docs.google.com/spreadsheets/d/1CclI28-fa2pPIMjckKjexZVkbn5rSgzVi1-1PZ_zsWc/edit#gid=0))

> It's not clear that it is especially relevant for an implementer to learn that some objects follow the same schema in different parts of the data. It seems okay to learn this in the schema reference.

Yes, actually, I think that it is more relevant for them to learn about that while developing an extension, for example, which is something more advance and requires understanding/looking at the schema reference, so I'm ok with not having this section too. (fun fact: I've never managed to found a correct/nice/perfect translation for this in Spanish, so I'm happy in that sense too)

> Since packages are going to become optional #1084, it is not important to mention them in the Primer.

That is what I thought, all good then.

> Do you mean the two Example and Hint boxes? Isn't the Example covered by: https://standard.open-contracting.org/1.1/en/guidance/build/change_history/

It is partially covered, the https://standard.open-contracting.org/1.1/en/getting_started/releases_and_records/#repeating-previous-information section is not covered and I think that it contains useful guidance about **minimal updates** (recently one data user referenced this section of the documentation, so at least it is useful for some people)
And for compile releases and easy releases, maybe we can add a reference to the easy releases guidance in https://standard.open-contracting.org/1.1/en/guidance/build/change_history/?

jpmckinney:
Agreed on a call to:

* Update the codelist mapping template guidance with a better example, rather than have guidance in both the template's guidance and in the docs. Tracked in CRM-7491
* Create an issue to re-introduce a worked example for minimal updates and change history #1364

And in this PR, to:

* Link to easy_releases from change_history, for guidance on setting the release `id` 2d0b9c6a

PR #1307, opened 2021-06-15
847 record uniqueness

duncandewhurst:
Closes https://github.com/open-contracting/standard/issues/847

`records_reference.md`:

* Update the first paragraph to clarify the uniqueness rules for records.
* Update *Record structure* to clarify that: 
  * the releases list need only contain all releases **at the time of the record's publication**.
  * the versioned release need only contain a history of changes **up to the time of the record's publication**.
* Update *Compiled release* to clarify that the compiled release need only contain the latest values **at the time of the record's publication**.

`releases_and_records.md`

* Change 'While releases are never updated, records are updated each time there is a change.' to 'While releases are never updated, records **can be** updated each time there is a change.' to clarify that there is no requirement to update old records.

* Update `.description` in `record-package-schema.json` to describe only a record package, similar to `description` in `release-package-schema.json`. Previously, `.description` also paraphrased the description of a record in `definitions.Record.description`, possibly because there is no `record-schema.json` file. However, I think it's best to document the description of a record in only one place in the schema.

* Update `.description` in `definitions.Record`:
  * Align with the first sentence of `records_reference.md`.
  * Describe only the must/should/may rules for the releases list, compiled release and versioned release. 
  
  Previously, `.description` also paraphrased the descriptions from `.properties.releases`, `.properties.compiledRelease` and `.properties.versionedRelease`. However, I think it's best to document the descriptions of each property in only one place in the schema.

* Update `.description` in `definitions.Record.properties.releases` to clarify that the releases list need only contain all releases **at the time of the record's publication**

* Update `.description` in `definitions.Record.properties.compiledRelease` to clarify that the compiled release need only contain the latest values **at the time of the record's publication** and to replace 'latest version of all the contracting data' with 'latest value of each field'.

* Update `.description` in `definitions.Record.properties.versionedRelease` to clarify that the versioned release need only contain a history of changes **up to the time of the record's publication** and, since a versioned release can be published independently of a compiled release, replace 'all the data in the compiled release' with 'each field'.

PR #1308, opened 2021-06-16
857 document format codelist

### LINKED ISSUE #857: Add IANA Media Types codelist

duncandewhurst:
Closes #857


jpmckinney:
@duncandewhurst I changed the base to be 1.2-dev.

duncandewhurst:
> @duncandewhurst I changed the base to be 1.2-dev.

Thanks, I created the draft PR against 1.1-dev before pushing any commits in order to get the PR number for the changelog. Github won't let you create an 'empty' PR against, so as in https://github.com/open-contracting/standard/pull/1307 I was going to switch it to 1.2-dev after pushing to the branch. Let me know if there's a better way to get around having to do a separate commit for the changelog!

duncandewhurst:
@jpmckinney - `test_csv.py` errors because codes contain the '/' character and because the codelist lacks a description field. Please can you update the standard-maintenance-scripts?

jpmckinney:
I think it's fine to do two commits. Anyhow, the simplest method is to do a commit and create a PR, then `git reset HEAD~1` and force push the branch again.

Can you please disable whatever text editor feature/plugin removes whitespace from the end of lines? It will create unnecessary merge conflicts if another PR makes a substantive change.

* Please name the codelist `mediaType.csv` to align with IANA, and so that it is not coupled to documents.
* We should omit the "DEPRECATED" and "OBSOLETED" codes, given that they are inappropriate for use. In [RFC 6838](https://datatracker.ietf.org/doc/html/rfc6838#section-5.5):

    > Media type registrations may not be deleted; media types that are no longer believed appropriate for use can be declared OBSOLETE by a change to their "intended use" field; such media types will be clearly marked in the lists published by the IANA.

  In the codelist's description in `codelists.md`, we can simply explain that it omits deprecated and obsoleted media types. We can then remove the Notes column.

jpmckinney:
> @jpmckinney - `test_csv.py` errors because codes contain the '/' character and because the codelist lacks a description field. Please can you update the standard-maintenance-scripts?

Done, though the name needs to be changed to `mediaType.csv`.

util/fetch_document_format_codelist.py
```diff
@@ -0,0 +1,59 @@
...
+"""
+
+import csv
+import os
+import requests
+
+from helper import schema_dir
+
+
+def get_and_parse_csv(url):
```
jpmckinney: This method can be in-lined, since it is called only at one location.

util/fetch_document_format_codelist.py
```diff
@@ -0,0 +1,59 @@
...
+        'multipart',
+        'text',
+        'video'
+    ]
+
+    codes = {}
+
+    for registry in registries:
+        csvreader = get_and_parse_csv(f'https://www.iana.org/assignments/media-types/{registry}.csv')
+        next(csvreader)
```
jpmckinney: Hmm, this means we're skipping the first code of each CSV (no font/collection, etc.).

util/fetch_document_format_codelist.py
```diff
@@ -0,0 +1,59 @@
...
+        for row in csvreader:
+            code = row['Template']
+            title = row['Name']
+            # If the type has no value in the template column, construct a code from the registry and name
+            if code == "":
+                code = f'{registry}/{title}'
```
jpmckinney: Let's use Python's `logging` to log a warning.

util/fetch_document_format_codelist.py
```diff
@@ -0,0 +1,59 @@
...
+                title, notes = title.split(' (', 1)
+            else:
+                notes = None
+            codes[code] = {'Title': title, 'Notes': notes}
+
+    with open(os.path.join(schema_dir, 'codelists', 'documentFormat.csv'), 'w') as fp:
+        writer = csv.writer(fp, lineterminator='\n')
+        writer.writerow(['Code', 'Title', 'Notes'])
+        for code in codes.keys():
+            writer.writerow([code, codes[code]['Title'], codes[code]['Notes']])
```
jpmckinney:
```suggestion
        for code, obj in codes.items():
            writer.writerow([code, obj['Title'], obj['Notes']])
```

util/fetch_document_format_codelist.py
duncandewhurst: Done in https://github.com/open-contracting/standard/pull/1308/commits/bf809b43536493308a45c3dba75c6d432ceaead4

util/fetch_document_format_codelist.py
duncandewhurst: Ah, yes. Previously I was using `csv.reader`. Fixed in https://github.com/open-contracting/standard/pull/1308/commits/bf809b43536493308a45c3dba75c6d432ceaead4

util/fetch_document_format_codelist.py
duncandewhurst: Done in https://github.com/open-contracting/standard/pull/1308/commits/bf809b43536493308a45c3dba75c6d432ceaead4. Should we also log the removal of deprecated and obsolete codes?

util/fetch_document_format_codelist.py
duncandewhurst: I don't think this is necessary now that the notes column has been removed in https://github.com/open-contracting/standard/pull/1308/commits/bf809b43536493308a45c3dba75c6d432ceaead4.

duncandewhurst:
> Can you please disable whatever text editor feature/plugin removes whitespace from the end of lines? It will create unnecessary merge conflicts if another PR makes a substantive change.

Done and unintentional changes reverted in https://github.com/open-contracting/standard/pull/1308/commits/7d8cd5908592a11bdda5f8ca62f5531dabdd7af9. I've no idea why the whitespace plugin in Atom started to do that!

>     * Please name the codelist `mediaType.csv` to align with IANA, and so that it is not coupled to documents.

Done in https://github.com/open-contracting/standard/pull/1308/commits/bf809b43536493308a45c3dba75c6d432ceaead4. I had named it differently since we added the 'offline/print' code, but happy to keep the same name if you prefer.

>     * We should omit the "DEPRECATED" and "OBSOLETED" codes, given that they are inappropriate for use. In [RFC 6838](https://datatracker.ietf.org/doc/html/rfc6838#section-5.5):
>       > Media type registrations may not be deleted; media types that are no longer believed appropriate for use can be declared OBSOLETE by a change to their "intended use" field; such media types will be clearly marked in the lists published by the IANA.
>       
>       In the codelist's description in `codelists.md`, we can simply explain that it omits deprecated and obsoleted media types. We can then remove the Notes column.

Done in https://github.com/open-contracting/standard/pull/1308/commits/bf809b43536493308a45c3dba75c6d432ceaead4.

util/fetch_document_format_codelist.py
jpmckinney: Good idea!

jpmckinney:
Unfortunately, to avoid future conflicts, you will need to restart the PR rather than revert the change in a new commit. You can just download the patch (add .patch to the PR URL), and then `git apply filename.patch`.

util/fetch_document_format_codelist.py
duncandewhurst: Done in https://github.com/open-contracting/standard/pull/1314/commits/1d9a59f60e03b6bd9e0290e147437d647973b5b1

duncandewhurst:
Restarted in #1314

PR #1318, opened 2021-06-21
Deprecate versionId

### LINKED ISSUE #812: Deprecate versionId
After #810, none of our code will check `versionId`, as it is an incomplete means of getting the correct versioning behavior. In 2.0, we would then be able to remove references to `versionId` from:

* [x] standard: release-schema.json
* [x] standard: meta-schema.json
* [x] standard: meta-schema-patch.json
* [x] standard: make_versioned_release_schema.py

As part of the deprecation, we'd need to update:

* [x] ocdskit: schema_report.py

yolile:
closes #812


schema/meta-schema.json
```diff
@@ -245,7 +245,11 @@
     },
     "versionId": {
       "type": "boolean",
-      "default": false
+      "default": false,
+      "deprecated": {
```
yolile: @jpmckinney I'm not sure if I should remove the field entirely or just adding the deprecated object

jpmckinney: Remove the field entirely. We only use `deprecated` in the actual schema, not the metaschema.

jpmckinney:
Changelog entry is missing.

yolile:
> Changelog entry is missing.

added

jpmckinney:
I've also removed the note linking to the issue. I'll merge when tests pass.

PR #1319, opened 2021-06-22
rename item classification

### LINKED ISSUE #1034: Rename itemClassificationScheme.csv to classificationScheme.csv
Follow-up to #982

yolile:
closes #1034


docs/schema/codelists.md
```diff
@@ -40,16 +40,14 @@ The organizations, economic operators or other participants in a contracting pro
 :file: ../../build/current_lang/codelists/partyRole.csv
 ```
 
-### Item Classification Scheme
+### Classification Scheme
 
-Items should be classified using existing item classification schemes, such as the [EC Common Procurement Vocabulary (CPV)](https://simap.ted.europa.eu/web/simap/cpv).
```
yolile, line 45: @jpmckinney I also moved this sentence from here to the item.classification definition in the release-schema.

yolile:
Not sure if I should commit the locales too

schema/release-schema.json
```diff
@@ -1405,7 +1405,7 @@
         },
         "classification": {
           "title": "Classification",
-          "description": "The primary classification for the item.",
+          "description": "The primary classification for the item. Items should be classified using existing item classification schemes, such as the [EC Common Procurement Vocabulary (CPV)](https://simap.ted.europa.eu/web/simap/cpv).",
```
jpmckinney: I think this sentence was unnecessary in the original documentation, and we can just remove it. Any scheme used by publishers exists (it's not like they invented one just for OCDS).

jpmckinney (commented):
No need to update the PO files. We're going to have a lot to translate in 1.2, so it's not really worth the effort to edit the source and translated texts.

PR #1324, opened 2021-07-06
add tender.datePublished field

### LINKED ISSUE #892: Add tender.datePublished field
### The problem

Many countries have the information on when the tender was published. We are currently working specifically with Honduras and Paraguay, and both countries have this information available for publishing.

Having this information in a easy way to access it can have many benefits, for instance, to be able to compute some indicators that will say how many days a potential bidder has to prepare an offer (tender.tenderDate.endDate -  tender.datePublihsed).

Currenly there is no easy attribute that can be used to publish this information. In the past [there was a discussion](https://github.com/open-contracting/standard/issues/261) on how to do this, but the proposed solution is not very straigth forward and is not simple

### Proposed solution

The proposal is to create a simple extension with one attribute called **tender.datePublished** of type date which can then be used by publishers.

duncandewhurst:
Closes #892

### DIFF
--- a/docs/history/changelog.md
+++ b/docs/history/changelog.md
@@ -107,7 +107,8 @@ Per the [normative and non-normative content and changes policy](https://docs.go
 * Improve the clarity of field descriptions in the release package schema and record package schema:
   * [#1067](https://github.com/open-contracting/standard/pull/1067) `Publisher.name`, to indicate that it is the organization or department responsible for publishing the OCDS version of the data.
 * [#1257](https://github.com/open-contracting/standard/pull/1257) Remove default value of `"en"` for `language` field.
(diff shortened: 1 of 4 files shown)


mrshll1001 (approved):

duncandewhurst:
Once 1.2 is deployed we should update indicator UC047 from the [use case guide](https://docs.google.com/spreadsheets/d/1j-Y0ktZiOyhZzi-2GSabBCnzx6fF5lv8h1KYwi_Q9GM/edit#gid=1075539355) to use this field:

> The percent of tenders for which the tender documentation was added after publication of the announcement

jpmckinney:
@duncandewhurst Can you create a CRM issue for post-1.2 updates to guidance?

jpmckinney (approved):

duncandewhurst:
I've created CRM-7432

PR #1326, opened 2021-07-08
add Links field

### LINKED ISSUE #928: Add links field
## Issue

A government might only publish OCDS data for new events in a contracting process. However, they would still want to have a link back to any earlier events (in non-OCDS format).

This might occur in cases of jurisdictional changes, like municipal mergers or EU withdrawal, where the cost and complexity of converting old events might be too high.

## Proposal

Add a top-level field, similar to `relatedProcesses`, named `links`, e.g.:

```json
{
  "properties": {
    "links": {
      "title": "",
      "description": "",
      "type": "array",
      "items": {
        "$ref": "#/definitions/Link",
      }
    }
  },
  "definitions": {
    "Link": {
      "title": "",
      "description": "",
      "type": "object",
      "properties": {
        "same fields as RelatedProcess, except `title`"
      }
    }
  }
}
```

The `relationship` field would have a different codelist than `relatedProcess.csv`, with codes for:

* `prev`, Previous publication, The linked resource is a previous publication about this contracting process, that is not in OCDS format or is using a different OCID. This code is intended for use when transitioning from an old system without converting its data, which might occur in cases of jurisdictional changes like municipal mergers.

We can propose other codes like:

* `alternate`, Alternate representation, The linked resource is an alternate representation of this contracting process. For example: a local XML format.

There probably aren't any others we want to include at first, but we can be inspired by https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types

The codes should all make sense in the context of a contracting process, not in the context of the individual release. If the field were used in the context of an individual release, then we'd have to make the field `"omitWhenMerged": true`. In most cases, a `documents` field can be used to make links to external resources related to an individual release.

duncandewhurst:
Closes #928


duncandewhurst:
@jpmckinney a couple of questions on this draft PR:

1. Should `Link.properties.rel` be a string or an array? The example in https://github.com/open-contracting/standard/issues/928#issuecomment-719070552 uses a string, but according to https://json-schema.org/draft/2019-09/json-schema-hypermedia.html#rfc.section.6.2.1 it must be a string or an array. I can't immediately think of a use case for an array of relation types in the OCDS context, but an array might be safer in case a use case comes up in the future.
2. Should `linkRelationType.csv` reuse the descriptions from https://www.iana.org/assignments/link-relations/link-relations.xhtml verbatim, or should we make them specific to the OCDS context, i.e. by replacing 'resource', 'link context' and 'context' with 'release or compiled release'? Currently, I've reused them verbatim.
3. Since `links.omitWhenMerged` = true, I've omitted `null` from `links.type`, `id` from `Link.properties`, and `null` from `Links.properties.rel.type` and `Links.properties.href.type`. However, the tests report errors. What do you suggest?
4. Should I open a separate PR to merge the pagination extension per https://github.com/open-contracting/standard/issues/928#issuecomment-660284323, or add it to this one?

jpmckinney:
1. In [HTML](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel), each `link` element has a single `rel` attribute value, so I think it is safe to keep it to a string in OCDS as well. There can always just be another entry in the `links` array to add the same URL with a different `rel`. I think this case is too infrequent to make the data structure more complicated for the general case.
2. I think we should add only those for which we have a use case, as most do not make sense and are unlikely to ever be used in the context of OCDS. When adding them, we should rewrite their description to make sense in OCDS context.
3. I can update the tests once the PR is ready to get tests to pass.
4. Good question. The pagination extension takes the form of an object `{"links": {"next": "https://..."} }` on packages (this is what the [JSON API spec](https://jsonapi.org/profiles/ethanresnick/cursor-pagination/) and others do), whereas #928 takes the form of an array (following JSON Hyper Schema). I think this is okay, but I only noticed now. In any case, we should do separate PRs, unless we want to align the approach across packages and individual releases.

duncandewhurst:
Great, thanks.

For 2. I only added `prev`, `canonical`, and `alternate` per the issue, so I'll update the descriptions for those.

For 4. I can't think of a particular upside to aligning the approach across packages and individual releases. At the package level, the object is (marginally) easier to process and aligns with the JSON API spec. At the release level, using an array with an open codelist for `rel` means that we don't need to change the schema to support new link types. So a separate PR sounds good.

mrshll1001 (approved):

docs/history/changelog.md
```diff
@@ -108,6 +108,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
   * [#1067](https://github.com/open-contracting/standard/pull/1067) `Publisher.name`, to indicate that it is the organization or department responsible for publishing the OCDS version of the data.
 * [#1257](https://github.com/open-contracting/standard/pull/1257) Remove default value of `"en"` for `language` field.
 * [#1296](https://github.com/open-contracting/standard/pull/1296) Deprecate the `tender.eligibilityCriteria` field in favor of the new `tender.exclusionGrounds` field, in order to use more common terminology and improve semantics. 
+* [#1326](https://github.com/open-contracting/standard/pull/1326) Add the top-level `Links` field.
```
```suggestion
* [#1326](https://github.com/open-contracting/standard/pull/1326) Add a top-level `links` field.
```

docs/schema/codelists.md
```diff
@@ -170,6 +170,16 @@ The media type codelist is based on the [IANA Media Types](https://www.iana.org/
 :file: ../../build/current_lang/codelists/mediaType.csv
 ```
 
+### Link Relation Type
+
+The link relation type codelist is a subset of the [IANA Link Relations](https://www.iana.org/assignments/link-relations/link-relations.xhtml) list.
```
jpmckinney: Prefer "based on" since we don't copy semantics directly. The IANA h1 heading is Link Relations, but the actual list is under Link Relation Types.

```suggestion
The link relation type codelist is based on a subset of the [IANA Link Relation Types](https://www.iana.org/assignments/link-relations/link-relations.xhtml) list.
```

schema/release-schema.json
```diff
@@ -107,6 +107,16 @@
       "description": "The details of related processes: for example, if this process follows on from one or more other processes, represented under a separate open contracting identifier (ocid). This is commonly used to relate mini-competitions to their parent frameworks or individual tenders to a broader planning process.",
       "title": "Related processes",
       "type": "array"
+    },
+    "links": {
+      "title": "Links",
+      "description": "Links to related resources. In a release, links relate to the individual release. For example, a link to a canonical version of the release in another OCDS publication. In a compiled release, links relate to the entire contracting process. For example, a link to a non-OCDS publication that represents the entire contracting process.",
```
jpmckinney: Fix sentence fragments:

```suggestion
      "description": "Links to related resources. In a release, links relate to the individual release: for example, a link to a canonical version of the release in another OCDS publication. In a compiled release, links relate to the entire contracting process: for example, a link to a resource in a non-OCDS publication that represents the entire contracting process.",
```

schema/release-schema.json
```diff
@@ -2283,6 +2293,26 @@
           ]
         }
       }
+    },
+    "Link": {
+      "title": "Link",
+      "description": "A typed connection to a related resource.",
```
jpmckinney: Where is this wording from? "A reference to a related resource" might be easier to interpret.

schema/release-schema.json
```diff
@@ -2283,6 +2293,26 @@
...
+      "type": "object",
+      "properties": {
+        "rel": {
+          "title": "Link Relation Type",
```
jpmckinney: Use sentence case for title properties.

schema/release-schema.json
```diff
@@ -2283,6 +2293,26 @@
...
+          "description": "The semantics of the link, using the open [link relation type](https://standard.open-contracting.org/{{version}}/{{lang}}/schema/codelists/#link-relation-type) codelist.",
+          "type": "string",
+          "codelist": "linkRelationType.csv",
+          "openCodelist": true
+        },
+        "href": {
+          "title": "Link Target",
+          "description": "The target URI of the link.",
```
jpmckinney: Use URL instead of URI.

```suggestion
          "description": "The URL of the related resource.",
```

schema/codelists/linkRelationType.csv
```diff
@@ -0,0 +1,4 @@
+Code,Title,Description
+canonical,Canonical,Designates the preferred version of a release of compiled release.
```
jpmckinney:
```suggestion
canonical,Canonical,Designates the preferred version of a release or compiled release.
```

schema/codelists/linkRelationType.csv
```diff
@@ -0,0 +1,4 @@
...
+prev,Previous,"Indicates that a release or compiled release is a part of a series, and that the previous in the series is the link target."
+alternate,Alternate,Refers to a substitute or altnerative representation of a release or compiled release.
```
jpmckinney:
```suggestion
alternate,Alternate,Refers to a substitute or alternative representation of a release or compiled release.
```

schema/release-schema.json
```diff
@@ -2283,6 +2293,26 @@
...
       }
+    },
+    "Link": {
+      "title": "Link",
+      "description": "A typed connection to a related resource.",
+      "type": "object",
+      "properties": {
+        "rel": {
+          "title": "Link Relation Type",
+          "description": "The semantics of the link, using the open [link relation type](https://standard.open-contracting.org/{{version}}/{{lang}}/schema/codelists/#link-relation-type) codelist.",
```
jpmckinney: This description is too general.

New version is based on: "The rel attribute defines the relationship between a linked resource and the current document." https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel

```suggestion
          "description": "The relationship with the related resource, using the open [link relation type](https://standard.open-contracting.org/{{version}}/{{lang}}/schema/codelists/#link-relation-type) codelist.",
```

schema/release-schema.json
duncandewhurst: It's from https://datatracker.ietf.org/doc/html/rfc5988#section-3. I'm happy with either wording.

schema/release-schema.json
duncandewhurst: Looks good. I used the wording from https://datatracker.ietf.org/doc/html/rfc5988#section-4

schema/release-schema.json
duncandewhurst: Done in https://github.com/open-contracting/standard/pull/1326/commits/8d57e733867f03087514cd804cf646d6bf80ea29

schema/release-schema.json
```suggestion
      "description": "A reference to a related resource.",
```

jpmckinney (commented):
IETF's terminology is a bit wonky, so made this suggestion.

duncandewhurst:
@jpmckinney I think all that's left to do is to update the tests to handle:

>     3\. Since `links.omitWhenMerged` = true, I've omitted `null` from `links.type`, `id` from `Link.properties`, and `null` from `Links.properties.rel.type` and `Links.properties.href.type`

jpmckinney (approved):

PR #1330, opened 2021-07-13
Improve structure of Guidance/Map

### LINKED ISSUE #1300: Better Guidance/Map structure
I think we can improve https://standard.open-contracting.org/latest/en/guidance/map/ with the following steps:
1) Remove the title and level (not the content) "Map your data to OCDS" in Guidance/Map. Semantically, the title "Map your data to OCDS" is identical to "Map" above it, which I find confusing. Furthermore, it will allow us to reduce the overall depth of the guidance/map section. The text in "Map your data to OCDS" can be put under two new sections: "Mapping templates" (first four paragraphs) and "Localizing OCDS" (last paragraph). This will have the added benefit that https://standard.open-contracting.org/latest/en/guidance/map/localization/ will no longer be linked to in the table of contents without having a heading in the text one level above it, which is currently the (anomalous) case. The sections "Mapping organisation identifiers", "Working in parallel", "Splitting up the work" and "Mapping the hard cases" will consequently be directly under "Map". 
2) Remove "Mapping organisation identifiers" (including the content), because it only contains links to two better resources on organisation identifiers (and copy-pasted text from them), one of which is in just the neighbouring section "Mapping the hard cases" and also contains a link to the other resource. 
2) Rename "Mapping the hard cases" to "The hard cases". (Just a detail, but probs no need to repeat "Map", plus, arguably and not quite consistently with the general headline, some sections such as "Organisation identifiers" and "Personal identifiers" are not about mapping, but about filling in.) 
3) Reorder the sections by meaning, e.g. so that we first talk about the "people/organisational topics" and then "data topics". I would suggest:
    - Involve the right people 
    - Identify your data sources 
    - Working in parallel 
    - Splitting up the work
    - Mapping templates
    - Localizing OCDS
    - The hard cases
    - Extensions
    - Linked Standards
    - Wrapping up

JachymHercher:
Close https://github.com/open-contracting/standard/issues/1300

### DIFF
--- a/docs/guidance/map.md
+++ b/docs/guidance/map.md
@@ -1,6 +1,8 @@
 # Map
 
-This phase is about documenting your sources of contracting data, and documenting how that data "maps" to OCDS – that is, identifying which [data elements](https://en.wikipedia.org/wiki/Data_element) within your data sources match which OCDS [fields](../../schema/reference) and [codes](../../schema/codelists). The mapping phase is one of the longest and most important steps in the implementation process.
+This phase is about documenting your sources of contracting data, and documenting how that data "maps" to OCDS – that is, identifying which [data elements](https://en.wikipedia.org/wiki/Data_element) within your data sources match which OCDS [fields](../../schema/reference) and [codes](../../schema/codelists). The mapping phase is one of the longest and most important steps in the implementation process. 
+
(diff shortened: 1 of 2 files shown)


jpmckinney:
Re-opened as #1337 due to / in branch name causing problems with build previews.

PR #1331, opened 2021-07-14
release-schema: add Value.date field

### LINKED ISSUE #988: Add Value.date field
The unit of an amount is not only the currency, but also the date on which it was set, due to inflation. To be able to correctly sum values over longer periods of time, especially in contexts with high inflation, it is important to store the date on which the amount was set.

For the definition, we can see https://en.wikipedia.org/wiki/Value_date

cc @duncandewhurst

yolile:
closes #988

### DIFF
--- a/schema/release-schema.json
+++ b/schema/release-schema.json
@@ -2079,6 +2079,15 @@
             "ZWR",
             null
           ]
+        },
+        "date": {
+          "title": "Date",
+          "description": "The date when the value is set. The value date is used when analysing historical values due to inflation over time.",
+          "type": [
+            "string",
+            "null"
+          ],
+          "format": "date-time"
         }
       }
     },
--- a/schema/release-schema.json
+++ b/schema/release-schema.json
@@ -2082,7 +2082,7 @@
         },
         "date": {
           "title": "Date",
-          "description": "The date when the value is set. The value date is used when analysing historical values due to inflation over time.",
+          "description": "The date on which the value is set. The value date is useful to correctly sum values over longer periods of time, especially in contexts with high inflation.",
           "type": [
             "string",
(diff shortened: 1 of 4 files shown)


schema/release-schema.json
```diff
@@ -2079,6 +2079,15 @@
             "ZWR",
             null
           ]
+        },
+        "date": {
+          "title": "Date",
+          "description": "The date when the value is set. The value date is used when analysing historical values due to inflation over time.",
```
duncandewhurst: "when" -> "on which", for consistency with the descriptions of other date fields.

I think the usage example from the issue description is clearer:

```suggestion
          "description": "The date on which the value is set. The value date is useful to correctly sum values over longer periods of time, especially in contexts with high inflation.",
```

duncandewhurst (changes requested):
Did you consider expanding the description based on https://en.wikipedia.org/wiki/Value_date?

yolile:
> Did you consider expanding the description based on https://en.wikipedia.org/wiki/Value_date?

Yes, but I'm sure if "assets" is a concept that we want to introduce?

duncandewhurst (approved):

duncandewhurst:
> > Did you consider expanding the description based on https://en.wikipedia.org/wiki/Value_date?
> 
> Yes, but I'm sure if "assets" is a concept that we want to introduce?

Probably not, since `Value` can be used in many different contexts in OCDS. I think the description is clear enough as it is.

jpmckinney:
"is set" can be confused with "when we agreed this was worth $100", which might be at a very different time than "when we traded $100 for goods". This led me to more comments in #988. Closing per comment there.

PR #1332, opened 2021-07-14
Framework agreements guidance improvements

### LINKED ISSUE #1301: Improvements/fixes to the framework guidance
I have a few questions on the framework guidance at https://standard.open-contracting.org/latest/en/guidance/map/related_processes/

1) [x] 1) Are you sure the following is correct?  
> If the contracting process will establish a framework agreement with a single supplier and if there are no conditions to participate in the contracting process, set tender.procurementMethod to 'open'.

a) Isn't "with a single supplier" excessively restrictive? A contracting process under the open procedure can establish a framework agreement with multiple suppliers, in several independent scenarios: i) when winners are in a cascade ([example](https://ted.europa.eu/udl?uri=TED:NOTICE:619204-2020:TEXT:EN:HTML&src=0)), ii) with lots (e.g. same example as before), iii) when suppliers are in a consortium (e.g. same example as before); iv) with reopening of competition ([example](https://ted.europa.eu/udl?uri=TED:NOTICE:376094-2018:TEXT:EN:HTML&src=0)). 

b) Isn't "no conditions to participate" excessively restrictive? An open procedure typically involves exclusion and selection criteria same as any other procedure - the NSS worked examples contains them too. Isn't the distinction the same as described in https://standard.open-contracting.org/1.1/en/schema/codelists/#method ? (I.e. whether there is a qualification phase or not). If yes, is it even worth mentioning here? 

2) Why does https://extensions.open-contracting.org/en/extensions/competitive/master/ exist? Couldn't the same distinction be made by https://standard.open-contracting.org/1.1/en/schema/codelists/#method, where competitive corresponds to 'selective' and uncompetitive corresponds to 'direct'? I think the semantics would not need any changing; we should speak about competition the same way regardless of the "stage" it takes place in; and it means less fields to deal with. 

3) [x] 3) Drafting detail: Just for clarity's sake, in section https://standard.open-contracting.org/latest/en/guidance/map/related_processes/#award-of-a-procurement-contract-resulting-from-a-second-stage-competition I would change "tender release for the second stage" to "tender release for the invitation to participate in a second-stage competition", just to make it 100% clear that it refers to the previous paragraph. 

4) [x] In https://standard.open-contracting.org/latest/en/guidance/map/related_processes/#award-of-a-procurement-contract-resulting-from-a-second-stage-competition, shouldn't `tender.competitive` be set to `true`? 

5) [x] 5) Drafting detail: I find https://standard.open-contracting.org/latest/en/guidance/map/related_processes/#extensions important and general. My reflex would be to place it just under the flowchart above https://standard.open-contracting.org/latest/en/guidance/map/related_processes/#invitation-to-participate-in-the-first-stage-of-a-framework-agreement-procedure, consequently also removing the need to duplicate the sentence "The tender.competitive field, from the Competitive extension, is used to indicate whether the second stage involves competition." 

6) Clarification question: "Award of a procurement contract without second-stage competition" requires a new ocid. This is done because we want framework agreements contracts to be dealt with consistently, regardless of whether they involve second-stage competition or not, is that correct?

7) [x] 7) Total detail: No idea how big a hassle it is, but wouldn't the URL be better if changed from "related_processes" to something like "framework_agreements" or "framework_agreements_and_related_processes"?

yolile:
closes #1301


jpmckinney:
@JachymHercher Can you review this issue as the author of #1301?

JachymHercher (approved):
Good! I would point out three rather small things: 
1) Wouldn't "Framework agreements" be a better name than "Framework agreements and related processes"? Mainly because we talk about related processes also elsewhere (e.g. Unsuccessful procedures) and related processes are essentially just a tool for modeling FAs, which is the real issue at stake. (Three commits). 
2) 'true', not `true`: https://github.com/open-contracting/standard/pull/1332/commits/ec004641463aa21f75370868f8801782c26f49c2. (@jpmckinney please confirm that's the right approach)
3) Changelog missing, I guess even minor changes such as this one should have it? (@yolile)

yolile:
@JachymHercher thank you! I added the missing changelog entry now, @jpmckinney do you want to review the changes before merging them?

JachymHercher:
I've added some more comments in https://github.com/open-contracting/standard/issues/1301#issuecomment-888908989, perhaps we can deal with them in the same PR?

yolile:
@JachymHercher I've updated the image to remove the arrows and clarify that the "award" refers to the second stage. It is any better that way?

yolile:
@JachymHercher are you ok with an image like this?

screenshot_7

JachymHercher (approved):

JachymHercher:
@yolile Looks good, sorry for the slow reply!

jpmckinney:
@JachymHercher true and false are not codelist string values, they are booleans, which is a different data type.

@yolile Please open a new PR to correct the error (change single quotes to backticks).

In general, all PRs should be approved by me to avoid this and other mistakes.

jpmckinney:
For non-normative content, it is much less important to add a changelog entry, especially if it is just to say "improve X" without saying what was improved about it. I've updated the guidance: https://ocds-standard-development-handbook.readthedocs.io/en/latest/meta/changelog_style_guide.html#non-normative-content In this case, we can just add the PR number to another entry.

jpmckinney:
I created the PR since there were a few typos as well #1394

yolile:
> In general, all PRs should be approved by me to avoid this and other mistakes.

Ups, sorry, I thought that the Jachym approval was enough for this one

jpmckinney:
No worries! Usually I would say something like "Okay to merge after ..." to avoid ambiguity in the approval process.

PR #1334, opened 2021-07-14
Improve the structure of guidance/map/hard cases

### LINKED ISSUE #1333: Better Guidance/Map/*/Mapping the hard cases structure
The introduction of the planning process and changes in the definitions of the contracting process (https://github.com/open-contracting/standard/issues/866), contracts (https://github.com/open-contracting/standard/issues/896), awards (https://github.com/open-contracting/standard/issues/895) as well as the stages of the contracting process (https://github.com/open-contracting/standard/issues/1160) require changes in the documentation. Since Getting Started is and will be very general, the place to do this is the Guidance, specifically [Mapping the hard cases](https://standard.open-contracting.org/latest/en/guidance/map/#mapping-the-hard-cases).

Looking at the Guidance, I think this might be easier to do if we first "streamline" the Guidance with the current content and only as a second step stard updating it with new content. Here are my proposals on improving the structure of the hard cases, which follows up on the proposed improvements to the guidance/map section in general (https://github.com/open-contracting/standard/issues/1300) 

The points below describe the bigger changes, the smaller ones I suggest we discuss over the actual PR.   

1) Currently, the hard cases are listed alphabetically. I would propose ordering them into three thematic sections: cases linked to the contracting processes as a whole; to parts of the process; to organizations. Within each section, the cases should be structured by logical dependency (e.g. contracts before amendments) or by popularity (e.g. organisation classifications before beneficial ownership). I this will make it easier to navigate for users and it also corresponds to what I assume is the more typical use case for readers, i.e. reading through all (or almost all) of them when working on the mapping, not just going through a few selected cases. 

The resulting structure is at the bottom of this post. 

2) Rename "Unsuccessful tenders" to "Contracting process".
3) Restructure "Awards, contracts, buyers and suppliers".
    -  i) Split it into "Awards and contracts" and "Buyers and suppliers", as the section currently unnecessarily discusses too many things at once.
    - ii) Remove parts that are not particularly helpful (e.g. the UNCITRAL definitions). 
    - iii) Move "electronic catalogues" into a separate section, because the (very little) content is essentially independent of anything else. We could consider removing it altogether, but keeping it there (easily findable) just to answer the FAQ "and what about electronic catalogues?" is maybe not a bad idea.
    - iv) Remove the anomaly (for this level of nesting) that examples ("Award and contracts" etc.) are listed in separate webpages. This can be done by merging "Consortia suppliers" with "Buyers and suppliers"; removing the link to "Framework Agreements and Related Processes" (it is a self-standing page on the same level as "Awards, contracts, buyers and suppliers"), merging the remaining examples into the core text. The resulting page has 200 rows, which is not out of line with other pages (Framework Agreements have 300 rows, Amendments and Milestones 150 rows each).
    - v) Rename "Award notices and decisions" to "Award notices and awards (as that it what it is). 
    - vi) ("Purchase orders" should be changed as well, but I would do that when updating the section on the basis of #896). 
   
3) Consistently have uppercase only at the beginning of each heading.

____

Resulting structure: 

map/contracting_process
map/related_processes
map/pre-qualification
map/awards_contracts
map/electronic_catalogues
map/amendments
map/milestones
map/buyers_suppliers
map/organization_reference
map/organization_identifiers
map/organization_personal_identifiers
map/organizational_units
map/organization_classifications
map/beneficial_ownership

JachymHercher:
close #1333


JachymHercher:
(This PR is a bit of a "proof of concept" - I'd do another top to bottom proof read to make sure the old content fits well with the new structure before moving it from draft. On the other hand, since there will be quite a few changes (discussed in the first paragraph of https://github.com/open-contracting/standard/issues/1333), it might make sense to not spend too much time on the current version.)

jpmckinney:
Re-opened as #1338 due to / in branch name causing problems with build previews.

Issue #1154, opened 2021-01-05
Ensure tender field descriptions make sense for planning and contracting processes


jpmckinney:
Follow-up to https://github.com/open-contracting/standard/issues/866#issuecomment-751683275

> @jpmckinney: Publishers can create OCDS releases/records for planning processes, even though these are not the same as contracting processes. The contracting process(es) that result from the planning process should refer to it using `relatedProcesses` and use different OCID(s). Those contracting processes should omit the `planning` section. As for the planning processes, they should omit the `awards` and `contracts` section. Right now, in OCDS, the `tender` section covers fields relevant to describing the procedure (not just the solicitation), so planning processes might include some `tender` fields.

> @JachymHercher: I'm wondering what implications this ambiguity (`Tender` block being used both during the "planning stage" and "tender stage") has for definitions of `Tender`, `Planning` and potentially also others blocks. Perhaps their definitions should not refer to "stages" at all, but simply talk about the data that they contain.

duncandewhurst:
I think it would be helpful to explicitly state in the description of `tender` that it can be used in either a planning process or a contracting process and that its properties should be interpreted as such. We can give the different interpretations of `tender.value` discussed in #914 as an example.

JachymHercher:
> I think it would be helpful to explicitly state in the description of tender that it can be used in either a planning process or a contracting process and that its properties should be interpreted as such. We can give the different interpretations of tender.value discussed in #914 as an example.

https://github.com/open-contracting/standard/issues/1160#issuecomment-881950573 proposes a description of `tender` that explicitly mentions that it can be used within two different types of processes (and gives value as an example, of sorts).

yolile:
We somehow already do this by having a 'planning' `tender.status` code with this definition:

> A future contracting process is being considered. Early information about the process can be provided in the tender section. A process with this status might provide information on early engagement or consultation opportunities, during which the details of a subsequent tender can be shaped.
 

Maybe in the new definition we can refer to this code, and indicate that if the tender has that status, all its content refers to the planning process. I think that should apply for both, core fields but also for existing extensions to the tender block.

For context and reference, this came up during the support to Mexico City who will disclose now the enquires made during the planning/pre procurement process using the enquires extension.

JachymHercher:
> We somehow already do this by having a 'planning' tender.status code with this definition:

[...]

Issue #866, opened 2019-05-01
Clarify definition of contracting process


jpmckinney:
## Problem

A contracting process is described in terms of an 'initiation process' on http://standard.open-contracting.org/latest/en/getting_started/contracting_process/#defining-a-contracting-process:

> All the planning, tendering information, awards, contracts and contract implementation information related to a single initiation process.

> An initiation process might be a tender, a direct contract award, or a call to award a concession.

However, we never define what an 'initiation process' is, which makes the definition ambiguous. 

The only [example](http://standard.open-contracting.org/latest/en/implementation/related_processes/) given for multiple initiations is:

> for example, when a framework is setup, and then mini-competitions are used for purchases from the framework

This was introduced in #439 (related to #310) on April 24, 2017, after the peer review for 1.1, and the changes regarding the description of 'initiation process' were not mentioned in the [update](https://docs.google.com/spreadsheets/d/1x1WG2-7E134OJ6ceWV3Sn5okhn4ARIeVYI8D0ALBLHA/edit#gid=92056586) to peer reviewers. As such, we should treat the mini-competitions example as non-normative guidance.

The 'tender' code in the `initiationType` codelist would also need correction of its description. #821

## Background

In the BPMN for a restricted procedure shared by EBRD, the process could be said to be 'initiated' by the publication of a contract notice (tender notice), after which there is 'supplier qualification' (pre-qualification), 'tendering' (submitting bids, etc.), and so on. (The BPMN for an open procedure is similar, but without a supplier qualification stage.)

In the EU, the same notice is used whether there is a supplier qualification stage or not (which is why the phrase "tenders or requests to participate" is used).

In the EU, the contract establishing a framework agreement and each contract resulting from a framework agreement are considered to be the same procedure. https://github.com/eForms/eForms/issues/221#issuecomment-466436451

## Discussion

For a restricted procedure, I think it's correct to say there is only one initiation process (publication of a contract notice), not two (pre-qualification and tendering). This agrees with [this comment](https://github.com/open-contracting/standard/issues/371#issuecomment-292117603) from the peer review, and disagrees with this [later interpretation](https://github.com/open-contracting-extensions/ocds_qualification_extension/commit/9fb9d1ed8119dab8bed95fb450930aa043a762b4) in the [qualification extension](https://extensions.open-contracting.org/en/extensions/qualification/master/).

[...]

PR #1335, opened 2021-07-14
release-schema: add planning.id

### LINKED ISSUE #924: Add Planning.id field
Some countries, like Paraguay (and soon Honduras), have an identifier for the planning stage.

cc @yolile

yolile:
closes #924


schema/release-schema.json
```diff
@@ -122,6 +122,14 @@
       "description": "Information from the planning phase of the contracting process. Note that many other fields can be filled in a planning release, in the appropriate fields in other schema sections; these would likely be estimates at this stage, e.g. value in tender.",
       "type": "object",
       "properties": {
+        "id": {
+          "title": "Planning ID",
+          "description": "An identifier for this planning process. This may be the same as the ocid, or may be an internal identifier for this planning.",
```
duncandewhurst: I'm not sure the second sentence is necessary. I assume it's copied from `tender.id`, where it is necessary because the field is required, so publishers need guidance on how to populate the field even if the data source doesn't include it.

yolile: Yes, the description comes from `tender.id`, and yes, the field is not mandatory here, but, I thought that it can still be the same as the `ocid`, as the planning would be the first stage of the contracting process, so the planning id is likely to be the ocid as well, but agree, it is not necessary to mention it.

schema/release-schema.json
```diff
@@ -122,6 +122,14 @@
       "description": "Information from the planning phase of the contracting process. Note that many other fields can be filled in a planning release, in the appropriate fields in other schema sections; these would likely be estimates at this stage, e.g. value in tender.",
       "type": "object",
       "properties": {
+        "id": {
+          "title": "Planning ID",
+          "description": "An identifier for this planning stage.",
```
duncandewhurst: Is there a particular reason to use 'stage' rather than 'process' (as in the previous commit)?

According to https://github.com/open-contracting/standard/issues/866, in 1.2 the guidance will be that the `planning` section should only appear in planning processes, which are defined separately from contracting processes.

yolile: > Is there a particular reason to use 'stage' rather than 'process' (as in the previous commit)?

For consistency. Planning description uses "planning phase", Planning milestones "planning stage", but planning documents "planning process".  But you are right, happy to change it back to process, and maybe update the other fields' descriptions to process as well?

duncandewhurst: Aha, I see. Yes, let's go with "planning process" here, since the three terms are used equally and interchangeably and "planning process" aligns better with 1.2. For changing the other descriptions, maybe you can open a new issue about ensuring that the terminology in the planning section makes sense for planning processes (similar to https://github.com/open-contracting/standard/issues/1154) in case we need to make any other changes.

yolile: I've updated all the references "planning" in the release schema now

jpmckinney (approved):

Issue #1142, opened 2020-12-17
Add tender.standstillPeriod field


jpmckinney:
[UNCITRAL glossary](http://www.uncitral.org/pdf/english/texts/procurem/ml-procurement-2011/Glossary-e.pdf) defines:

Term | Definition | Reference
-- | -- | --
Standstill period | Defined in the Model Law as: “Period starting from the dispatch of a notice as required by paragraph 2 of article 22 of this Law, during which the procuring entity cannot accept the successful submission and during which suppliers or contractors can challenge, under chapter VIII of this Law, the decision so notified.” For the explanation of the term “procuring entity”, see # 62 above. For the explanation of the terms “successful submission” and “supplier or contractor”, see ## 84 and 85 below. | N/A

Its addition will depend on its availability among potential publishers.

Related to #205

jpmckinney:
If we add this field, we might want to add a worked example similar to #1344 in which an award is challenged during a standstill period. Whether we just need `statusDetails` or a new status code is to be determined.

odscrachel:
Just a note that this came up in modelling conversations with the UK Cabinet Office.

duncandewhurst:
According to the World Bank's [Benchmarking Infrastructure Development 2020](https://bpp.worldbank.org) database, the following countries' procurement legislation includes a standstill period. The data covers 40 countries and is based on applying the case study assumptions described in the [methodology](https://bpp.worldbank.org/methodology), essentially a competitive tender for a 'traditionally procured' (i.e. non-PPP) $50m infrastructure project. In most cases, it looks like the same standstill period provisions apply to all types of procurement, although in one or two the provisions are infrastructure-specific. In some cases the standstill period must be set out in the notice of intention to award and, in others, not.

Economy | Standstill period set out in the notice of intention to award
-- | --
Azerbaijan | No
Burkina Faso | No
China | No
El Salvador | Yes
Ethiopia | No
Georgia | No
Italy | Yes
Kenya | No
Kyrgyz Republic | No
Madagascar | No
Peru | No
Romania | No
Russian Federation | No
Spain | Yes
Turkey | No
Uganda | No
Ukraine | No
United Kingdom | Yes

I've extracted the relevant data into a [spreadsheet](https://docs.google.com/spreadsheets/d/1HFHIyaf9087fbkcjYfcs0OHymUlxwLXrU2uf-buQ-4k/edit#gid=1890964663), which includes details of the legal/regulatory provisions.

[...]

Issue #430, opened 2017-03-06
Tender: Suspended tender (e.g. blocked by court order)


timgdavies:
**Discussion of a possible extension to OCDS**

From OpenProcurement Ukraine:

> We have an intention of adding a new informational field for blocking all the processes related to a given auction. Our proposed name for the field is "blocked", and, in our point of view, it would be simplest for it to have two values - "true" or "false". The rationale behind such a new field is a necessity of blocking the auction in case of a court order prohibiting any further action with the auction and/or contract until further notice.

I believe the underlying issue here is:

* How to indicate when a tender process has been temporarily suspended as a result of a court order (or for some other reason).

There are two main options we could pursue:

**(1) Adding an entry to the tenderStatus codelist for use in ```tender.status```**

A tender status of 'suspended' could be paired with an extended field ```tender.statusDetails``` to describe the nature of the suspension;

This would mean that: 

* When the tender became blocked, it would cease to show up in lists of 'active' tenders;
* Applications would need to be aware of the 'suspended' tender status;

tenderStatus is a closed codelist, which means any changes have to be part of an upgrade of the standard, and should not be made by extensions. 

**(2) Adding a new field via an extension**

The suggested field from Open Procurement Ukraine is ```blocked```. 

I would suggest ```suspended``` as a more general terminology, which can be paired with 'suspendedDetails``` to provide a description of the reason for suspension.

On this approach:

* Applications that are not aware of the ```suspended``` field would continue to display the tender in a list of active tenders;
* Applications that are aware of the ```suspended``` field could use this to filter views, or display extra information about a contracting process

## Questions

Assuming we pursue option (2):

(1) Are there other uses for this kind of field, other than that raised by Open Procurement Ukraine? 

(2) Do you have any preference on terminology between 'blocked' or 'suspended' (or another option)?

(3) Should an extension also introduce any 'documentType' codelist entries to link to related reasons for suspension or documentation of the suspension/blocking of the process?

(4) Should guidance include any information on including milestones or other free-text descriptions about suspensions/blocking?

Please also comment if you have a preference for option (1).

myroslav:
2nd option is more preferable. The only question I'd have is to have suspension reasons codified somehow. Either via `suspended` field itself or via extra `suspendedCode` to accompany `suspendDetails`. Since in informational systems it is normal to have `suspendCode` and `suspendDetails` in OCDS exports derive either pogrammatically from Codelists or from dedicated description field filled by human.

[...]

PR #1344, opened 2021-07-22
Add contract suspension worked example

### LINKED ISSUE #758: Worked example (1.2): Contract suspension
In the [ageops](https://ageops.net) system in Afghanistan contracts can have status = suspended ([example](https://ageops.net/en/procurement-procedure/contract/progress/RRD%252FMOF%252FNRAP%252FKBL%252FLGR%252F055%252FC2%252F001))

This indicates that the contract is temporarily suspended, for example due to security reasons, and will either be resumed at a later date or cancelled. The system does not capture a reason or description for the suspension.

In OCDS, contract status is a closed codelist with the following codes: pending, active, cancelled, terminated.

Similar to the discussion on suspended tenders in https://github.com/open-contracting/standard/issues/430, there are two main options we could pursue:

**(1) Adding an entry to the contract status codelist**

A contract status of 'suspended' could be added to the codelist.

This would mean that:

* When the contract was suspended, it would cease to show up in lists of active contracts
* Consuming applications would need to be aware of the 'suspended' contract status

Contract status is a closed codelist, which means that any changes would have to be made as part of an upgrade to the standard.

**(2) Adding a new field via an extension**

A true/false `suspended` field could be added to the contract object to indicate whether the contract is suspended

This would mean that:

* Applications that are not aware of the new field would continue to display the contract in a list of active contracts;
* Applications that are aware of the new field could use this to filter views or display extra information about a contract

## Questions

Assuming we pursue option (2), at least until the next standard upgrade:

(1) Do any other systems or legal frameworks use the concept of a suspended contract?

(2) Does a suspended contract fit the description of the `active` code from the contract status codelist:

> This contract has been signed by all the parties, and is now legally in force.

(3) Is 'suspended' the appropriate terminology?

(4) Should an extension also introduce a free text `statusDetails` field where details of the reason for the suspension could be provided.

(5) Should an extension also introduce any 'documentType' codelist entries to link to related reasons for suspension or documentation of the suspension of the contract.

(6) Should guidance include any information on including milestones or other free-text descriptions about suspensions?

duncandewhurst:
Closes #758

As well as adding the worked example, I added a note to **Contract status** on `codelists.md` as that seems like the most proximate place to signpost publishers and users to the example.

Once agreed, we can add this example to the [statusDetails extension documentation](https://extensions.open-contracting.org/en/extensions/statusDetails/master/) for OCDS 1.1.


duncandewhurst:
@yolile @jpmckinney this worked example only addresses contract suspension, not suspended tenders (#430). Do you think we should have a separate worked example for suspended tenders, or try to have one worked example that applies to both contract suspension and suspended tenders, or can we do without a suspended tenders example given that the concept only came up once, so far?

docs/guidance/map/contract_suspension.md
```diff
@@ -0,0 +1,27 @@
+# Contract suspension
+
+A contract can be suspended after it is signed. A suspended contract is still legally in force, but work has been suspended. For example, a contract might be suspended in the following scenarios:
+
+* the buyer suspends a contract because the supplier fails to perform.
+* the supplier suspends a contract because the buyer fails to pay.
+* either party suspends a contract due to force majeure.
```
jpmckinney: Up to here, it sounds like the contract could just be 'terminatedEarly'. We'll need to explain a bit more about suspension (mainly that the work can be resumed, as I understand).

jpmckinney (commented):
In #430, I'm not clear how a process can be suspended/blocked prior to the award stage. 

Besides courts, #430 mentions "blocked by the decision of Deposit Guarantee Fund commission", but we don't have a full example of either. So, we can't do a worked example.

After the award stage, it's common for the signing of a contract to be blocked during a standstill period, where tenderers challenge the award on specific grounds. I added a comment to #1142 about adding a worked example for that scenario.

This PR is ready for a first review.

docs/guidance/map/contract_suspension.md
jpmckinney: The first paragraph from https://www.upcounsel.com/contract-suspension is clear:

> A contract suspension is the temporary cessation of performance. It's not the same as [... early] termination, which is a permanent cessation of performance.

duncandewhurst: Good catch. I've clarified the description in 17c63f726790742d83b69b30c353730c8047a6b9

mrshll1001 (approved):
I can't spot any typos or errors, and the expected files are changed/introduced.

jpmckinney (approved):
I fixed conflicts and made one commit.

jpmckinney:
I've created the follow-up issue for the extension: https://github.com/open-contracting/ocds-extensions/issues/168

After merging, please manually close #758. Since 1.2-dev is not the default branch, GitHub will not close it automatically.

PR #1352, opened 2021-07-29
Replace "party", "entity" etc with "organization"

### LINKED ISSUE #1173: partyRole codelist and elsewhere: entity vs. party (vs. organization)
The [`Organisation.roles` description](https://standard.open-contracting.org/latest/en/schema/release/#release-schema) is 

> The party's role(s) in the contracting process, using the open partyRole codelist.

The corresponding [`partyRole` codelist](https://standard.open-contracting.org/latest/en/schema/codelists/#party-role) description is 

> The organizations, economic operators or other participants in a contracting process are listed in the parties section. A single party can have one or more roles in the contracting process.

from which I assume (as well as from some older discussions) that in OCDS, "party" is preferable to "entity" or "organisation". 

However,
1) Half of the definitions in the `partyRole` codelist (i.e. 'buyer', 'procuringEntity', 'supplier', 'tenderer', 'funder') refer to entities, not parties.
2) A search through the [release schema](https://standard.open-contracting.org/schema/1__1__5/release-schema.json) gives several places where we talk about entities (excluding procuringEntity, which we take as a given because of backwards compatibility; and "[Legal Entity Identifier](https://en.wikipedia.org/wiki/Legal_Entity_Identifier)"), most prominently the title of `Organisation.id` and description of `Buyer`.
3) Possibly there are also other places in the documentation etc. where "entity" is used. 

Should we go through the definitions and documentation and change them to "party"?

JachymHercher:
closes #1173


jpmckinney:
I ran the regex ``(?<!third[ -]|rested|ested |rivate|ivate |ases, )part(y|ies)(?!": |\]\(#parties|[)`]? (section|array)|` fields|, (roles|identifier|tender)|/(details|id|1)|\.(identifier|details)|Role\.csv|Scale/|Scale codelist|_scale|Details_scale| scale extension)`` and excluded .po files, dereferenced-release-schema.json and versioned-release-validation-schema.json. (This ignores multiple forms of third-party, interested party, private party, party scale extension, and the many cases in which it is referring to the field in the schema explicitly.)

While reviewing this, I anticipated confusion between "organization ID" (the `id` field) and "organization identifier" (the `identifier` field). This confusion arises from the schema itself, rather than from this issue. I've made a commit to address this, and then another commit for miscellaneous fixes. Please see the changes in those commits.

The remaining occurrences are below.

**Likely requires a change:**

* framework_agreements.md: Uses the phrase "party to the framework agreement". I don't know if this is a term of art, or something we invented. If the latter, we should fix these (possibly introduced from edits in 1.1).

These use party/parties in the sense of contractual parties, as part of a discussion of the contract. I think they are fine:

* awards_contracts.md
* contract_suspension.md
* contractStatus.csv
* milestoneType.csv
* release-schema.json

These are fine, because they are not about organizations in OCDS:

* CLA.md
* governance.md

Mentions in deprecated fields and codes are also fine.

JachymHercher:
> framework_agreements.md: Uses the phrase "party to the framework agreement". I don't know if this is a term of art, or something we invented. If the latter, we should fix these (possibly introduced from edits in 1.1).

"Party to the framework agreement" is the expression used in the UNCITRAL definitions (quoted at the beginning of framework_agreements.md) as well as EU law (e.g. Directive 2014/24/EU). I'd say it's jargon, in the sense that it could be replaced by for example "within the framework agreement" or "that are members of the framework agreement" and nothing much would happen, but I think the current state is also more or less ok. No strong opinion.

jpmckinney (approved):

jpmckinney:
Ah, I had missed the initial occurrence. Thanks!

PR #1369, opened 2021-07-31
index: Use sphinx-panels

jpmckinney:
* [ ] Add i18n support to sphinx-panels, then translate: https://github.com/executablebooks/sphinx-panels/issues/68
* [x] Get images from @ericreese15 and add to panels (used in this [slide deck](https://docs.google.com/presentation/d/1din91eAycFO7i0CC7jfcJJ0krbhvlHyDK6fgnfQfD-U/edit), but I can't extract them)
* [x] Check that it looks okay, on desktop and mobile, and in a private browser to test visited links CSS style

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -12,12 +12,73 @@ The Open Contracting Data Standard (OCDS) enables disclosure of data and documen
 
 The OCDS is made up of several parts, designed to help you use OCDS effectively.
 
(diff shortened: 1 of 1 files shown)


jpmckinney:
sphinx-panels is obsoleted by sphinx-design. https://github.com/executablebooks/sphinx-design

Issue #1295, opened 2021-05-11
Incorrect manually entered award values


yolile:
In Ecuador, they have identified some cases where the awarded values, manually entered into the system by the procurement entities, are incorrect. They identified these errors by comparing the awarded values against the budget values (the budget values are connected to the budget system and filled automatically so that values are considered correct). In this case, as they know the values are incorrect, they could use another field to disclose this "incorrect" information and not set the award/value. 
They can have an extension instead, using another field like `valueEntered` to disclose the known incorrect award value.

cc @jpmckinney

jpmckinney:
Thanks for logging the issue!

If we hear of similar cases, we can consider a common approach and re-open this issue.

For now, I will tag it as a local extension and close it.

jpmckinney:
In the UK CRM-7381, a user can select a corporate register from a dropdown, or manually enter the register's name if it is not in the list. Since the manual entries are non-standardized, one option is to use `scheme` for dropdown entries and `schemeEntered` for manual entries. This separates data with a known issue, and makes a (narrow) semantic distinction.

PR #1372, opened 2021-08-03
Add country field to Address. Deprecate countryName.

### LINKED ISSUE #524: Add Address.country field (with ISO3166-1 alpha-2 codelist)
From workshop in Kiev:

We use address from schema.org - we need country code as well for some use cases.

Follow up with @myroslav.

duncandewhurst:
Closes #524 

Once merged:

- [x] Add the instructions for downloading ISO3166 in machine-readable form to https://ocds-standard-development-handbook.readthedocs.io/en/latest/standard/technical/deployment.html#perform-periodic-updates


duncandewhurst:
@jpmckinney @yolile a couple of questions to resolve before this PR is ready for review:

1. Currently, this PR deprecates `Address.countryName` based on the justification in https://github.com/open-contracting/standard/issues/524#issuecomment-734408821. On reflection, I have two concerns about this:
    1. I anticipate some publishers in lower-tech contexts will collect country as a free-text field, so we will have helpdesk requests about where to put this data in OCDS.
    1. ISO3166-1 is not freely/readily available, so we may be creating a barrier to data use by removing the human-readable country name from the standard.
2. I've integrated the script from https://github.com/open-contracting-extensions/ocds_countryCode_extension/tree/master/script into `manage.py`. Is https://ocds-standard-development-handbook.readthedocs.io/en/latest/standard/technical/deployment.html#perform-periodic-updates the right place to put the instructions for downloading ISO3166 in machine-readable form?

yolile:
> I anticipate some publishers in lower-tech contexts will collect country as a free-text field, so we will have helpdesk requests about where to put this data in OCDS.

Hmm, true, but, maybe, before publishing OCDS data (eg as part of the OCDS transformation process), they can map the names into their ISO codes? 

> ISO3166-1 is not freely/readily available, so we may be creating a barrier to data use by removing the human-readable country name from the standard.

Also true, but as I understand, we will publish the CSV with the codes and names, right? And I guess it also depends on the use case. For example, if the use case is to show the data in a website/mobile app etc, the developers can do the map and show the human-readable in the website using our CSV as the base, and as mentioned in this https://github.com/open-contracting/standard/issues/524#issuecomment-734430518, having the codes is better for that case. Then, another use case, let say that you want to analyze the suppliers by country or similar, it is easier to do that if the names/codes are standardized already and I think it doesn't matter too much if the name is a human-readable name or a code (same if you are planning to use the data for an AI project, etc)

> I've integrated the script from https://github.com/open-contracting-extensions/ocds_countryCode_extension/tree/master/script into manage.py. Is https://ocds-standard-development-handbook.readthedocs.io/en/latest/standard/technical/deployment.html#perform-periodic-updates the right place to put the instructions for downloading ISO3166 in machine-readable form?

Yes, I think there is a good place to put it

duncandewhurst:
> > I anticipate some publishers in lower-tech contexts will collect country as a free-text field, so we will have helpdesk requests about where to put this data in OCDS.
> 
> Hmm, true, but, maybe, before publishing OCDS data (eg as part of the OCDS transformation process), they can map the names into their ISO codes?

Ideally, yes. But, if the values are free-text (i.e. users type the country names rather than choosing them from a list) the mapping would only cover historic data - users might enter new acronyms, abbreviations or misspellings after OCDS is implemented.

> > ISO3166-1 is not freely/readily available, so we may be creating a barrier to data use by removing the human-readable country name from the standard.
> 
> Also true, but as I understand, we will publish the CSV with the codes and names, right? And I guess it also depends on the use case. For example, if the use case is to show the data in a website/mobile app etc, the developers can do the map and show the human-readable in the website using our CSV as the base, and as mentioned in this [#524 (comment)](https://github.com/open-contracting/standard/issues/524#issuecomment-734430518), having the codes is better for that case. Then, another use case, let say that you want to analyze the suppliers by country or similar, it is easier to do that if the names/codes are standardized already and I think it doesn't matter too much if the name is a human-readable name or a code (same if you are planning to use the data for an AI project, etc)

Good points, I forgot we were publishing the names ourselves. I agree that, depending on the application, codes might be better - I wasn't suggested to not add the code field. I also agree that it's not much of a barrier for more capable users to download the CSV and look up the human-readable names. However, from a spreadsheet user / low capacity perspective, users are likely to want the names in the results of their analysis and importing and joining the CSV might be a barrier.

> > I've integrated the script from https://github.com/open-contracting-extensions/ocds_countryCode_extension/tree/master/script into manage.py. Is https://ocds-standard-development-handbook.readthedocs.io/en/latest/standard/technical/deployment.html#perform-periodic-updates the right place to put the instructions for downloading ISO3166 in machine-readable form?
> 
> Yes, I think there is a good place to put it

Great. I'll do that once this PR is merged.

manage.py
```diff
@@ -456,6 +456,42 @@ def pre_commit():
     json_dump('versioned-release-validation-schema.json', get_versioned_release_schema(release_schema))
 
 
+@cli.command()
+@click.argument('path')
```
jpmckinney: This should instead use `@click.argument("file", type=click.File())`. `file` is an open file, so you can omit the `with` line and just `json.load(file)`.

manage.py
```diff
@@ -456,6 +456,42 @@ def pre_commit():
...
+def update_country(path):
+    """
+    Update schema/codelists/country.csv from PATH.
+    """
+    # https://www.iso.org/iso-3166-country-codes.html
+    # https://www.iso.org/obp/ui/#search
+    directory = Path(os.path.dirname(os.path.realpath(__file__)))
```
jpmckinney: The script already calculates this at the top of the file. Further down, you can just replace `directory / 'schema'` with `schemadir`.

manage.py
```diff
@@ -456,6 +456,42 @@ def pre_commit():
...
+
+    with open(directory / 'schema' / 'codelists' / 'country.csv', 'w') as f:
+        writer = csv.DictWriter(f, fieldnames=['Code', 'Title'], lineterminator='\n')
+        writer.writeheader()
+
+        for code in sorted(codes):
+            writer.writerow({
+                'Code': code,
+                'Title': codes[code],
+            })
```
jpmckinney:
```suggestion
        writer = csv.writer(f, lineterminator='\n')
        writer.writerow(['Code', 'Title'])
        for code in sorted(codes):
            writer.writerow([code, codes[code]])
```

docs/schema/codelists.md
```diff
@@ -182,6 +182,16 @@ The link relation type codelist is based on a subset of the [IANA Link Relation
 
 ## Closed Codelists 
 
+### Country
+
+The country codelist is used to provide the country where an address is located, using two-letter codes from [ISO3166-1](https://www.iso.org/iso-3166-country-codes.html). The country codelist adds an exceptional code for Kosovo ('XK').
```
jpmckinney: Addresses aren't located anywhere - they are locators :) It's the building that is located. X* codes are "user-assigned" in ISO3166-1 terminology.

```suggestion
The country codelist is used to provide the country component of an address, using uppercase two-letter codes from [ISO3166-1](https://www.iso.org/iso-3166-country-codes.html). The country codelist adds a user-assigned code for Kosovo ('XK').
```

manage.py
```diff
@@ -456,6 +456,42 @@ def pre_commit():
...
+
+    with open(path) as f:
+        rpc = json.load(f)[0]['rpc'][0]
+        offset = int(rpc[0])
+
+        for entry in rpc[3][1]:
+            d = entry['d']
+            codes[d[str(offset + 9)]] = re.sub(r' \(the\)|\*', '', d[str(offset + 13)])
+
+            assert d[str(offset + 9)] == d[str(offset + 15)]
```
jpmckinney: Before merging, I should try to remember why I do this (probably a check that the format didn't change).

manage.py
```diff
@@ -575,11 +611,15 @@ def update_media_type():
...
 @click.pass_context
-def update(ctx):
+def update(ctx, country_path):
     """
-    Update external codelists (currency, language, media type).
+    Update external codelists (country, currency, language, media type).
+
+    COUNTRY_PATH is the path to the file containing the ISO3166 codes.
     """
+    ctx.invoke(update_country, path=country_path)
```
jpmckinney: Since the `update_country` command requires an extra file which is a pain to retrieve manually, let's leave this command as-is. `update_country` will have to be performed separately.

schema/release-schema.json
```diff
@@ -1736,7 +1736,274 @@
...
-          "description": "The country name. For example, United States."
+          "description": "The country name. For example, United States.",
+          "deprecated": {
+            "description": "This field is deprecated, because OCDS isn't the place to annotate a country with additional information like its human-readable name. ISO3166-1 alpha-2 country codes can be provided in `.country`.",
+            "deprecatedVersion": "1.2"
+          }
+        },
+        "country": {
+          "title": "Country code",
+          "description": "The country where the address is located, from the closed [country](https://standard.open-contracting.org/{{version}}/{{lang}}/schema/codelists/#language) codelist. If you need to use a user-assigned code from ISO 3166-1 alpha-2, [create an issue](https://github.com/open-contracting/standard/issues) to discuss its addition to the country codelist.",
```
jpmckinney: Reworded first part per my earlier comment. Reworded second sentence to more clearly describe the reason for creating an issue.

```suggestion
          "description": "The country, from the closed [country](https://standard.open-contracting.org/{{version}}/{{lang}}/schema/codelists/#language) codelist. If a country is missing, [create an issue](https://github.com/open-contracting/standard/issues) to discuss the addition of a user-assigned code to the country codelist.",
```

schema/release-schema.json
```diff
@@ -1736,7 +1736,274 @@
...
             "null"
           ],
-          "description": "The country name. For example, United States."
+          "description": "The country name. For example, United States.",
+          "deprecated": {
+            "description": "This field is deprecated, because OCDS isn't the place to annotate a country with additional information like its human-readable name. ISO3166-1 alpha-2 country codes can be provided in `.country`.",
+            "deprecatedVersion": "1.2"
+          }
+        },
+        "country": {
```
jpmckinney, line 1733: This field should be before `countryName` in the schema, so that it is displayed first.

schema/release-schema.json
```diff
@@ -1736,7 +1736,274 @@
             "string",
             "null"
           ],
-          "description": "The country name. For example, United States."
+          "description": "The country name. For example, United States.",
+          "deprecated": {
+            "description": "This field is deprecated, because OCDS isn't the place to annotate a country with additional information like its human-readable name. ISO3166-1 alpha-2 country codes can be provided in `.country`.",
```
jpmckinney: Reworded to describe the value the new field brings.

```suggestion
            "description": "This field is deprecated in favor of `country`, to promote standardized country codes instead of non-standardized country names.",
```

jpmckinney (commented):
Regarding whether to have two fields or one: This is similar to the discussion in #1295 about `valueEntered`. When data is known to have quality issues, it sometimes makes sense to segment the "good" data from the "possibly bad" data.

For fields like numerical values, this is important, because major use cases become more difficult if the quality is mixed.

For address data, if the quality is bad, you just don't know the address and can't geolocate it. So, no use cases are really harmed by putting non-ISO codes (like full country names) in the `country` field. Whether the bad data is segmented or not, your use case is frustrated.

That said, if a system has only non-standardized country names, then putting their bad data in the only available field will cause lots of structural errors. (But still, the deprecated field can instead be used, which yields a deprecation warning.)

The choice really depends on the prevalence of systems that have free-text data entry for country names. We can perhaps deprecate it, and consult on whether to deprecate it or not. I added it to my note here: https://crm.open-contracting.org/issues/6532#note-14

manage.py
duncandewhurst: Nice! Done in acb42017acf26f89b53ecbeec1628c7eec4443d4

manage.py
duncandewhurst: Done in acb42017acf26f89b53ecbeec1628c7eec4443d4

manage.py
duncandewhurst: Done in acb42017acf26f89b53ecbeec1628c7eec4443d4

schema/release-schema.json
duncandewhurst: Done in ad12fb4938e981c07f9e059930219af6d342d5ad

jpmckinney (approved):
I added a couple commits.

The instructions for update-country are now part of the `--help` message, so the handbook can just remind the reader to follow the instructions in `./manage.py update-country --help`

manage.py
duncandewhurst: https://github.com/open-contracting/standard/blob/b0e10abf12e83fa7218d56a42fbc1d47b0e309ea/manage.py#L495

duncandewhurst:
PR for handbook update: https://github.com/open-contracting/standard-development-handbook/pull/247

PR #1374, opened 2021-08-03
release and record packages schemas: remove minItems

### LINKED ISSUE #419: Removing minItems constraint from release package and record package
In this [API spec issue](https://github.com/open-contracting/api-specification/issues/5) we are exploring whether release and record packages could contain an empty set of releases. 

This is helpful for an API returning a query response.

## Proposal

The ```minItems``` constraint is removed from release and record package, so that they can contain an empty set of releases.

## Process

Please indicate support or opposition for this proposal.

yolile:
Closes #419 
I'm not sure if we should also update the guidance on releases and record packages to mention when the list could be empty.


jpmckinney:
> I'm not sure if we should also update the guidance on releases and record packages to mention when the list could be empty.

Since packages will be optional #1084, I think it is not necessary to add new content about packages.

jpmckinney (approved):

PR #1376, opened 2021-08-03
release-schema: add publisher field

### LINKED ISSUE #325: Add publisher field (release schema)
Optionally a record/release should be able to have embedded within it the following properties.
- packageURI
- publishedDate
- publisher
- license
- publicationPolicy

This is important for API usage where all the releases/records do not belong to the same package or even belong to the same publisher. This is important as the API user should have easy access to this information (especially things like the licence).

These items most likely should be under a new property called possibly "packageMetadata" 
An example release with this property would look like:

```
{   
    "language": "en",
    "ocid": "ocds-213czf-000-00001",
    "id": "ocds-213czf-000-00001-05-contract",
    "date": "2010-05-10T10:30:00Z",
    "tag": [
        "contract"
    ],
    "initiationType": "tender",
    ...
    "packageMetadata": {
         "uri":"http://a.com/ocds-213czf-001-01-planning.json",
         "publishedDate":"2009-03-15T14:45:00Z",
         "publisher": {
             "scheme": "GB-COH",
             "uid": "09506232",
             "name": "Open Data Services Co-operative Limited",
             "uri": "http://standard.open-contracting.org/examples/"
          },
         "license":"http://opendatacommons.org/licenses/pddl/1.0/",
         "publicationPolicy":"https://github.com/open-contracting/sample-data/"
}
```

These properties should be excluded from the merging process when creating the versionedRelease/compiledRelease, as it is possible for a record to have many source packages. Nonetheless, the record itself could have a "packageMetadata" property also.

This could be considered as an extension (but a standadized one) and not part of the core standard as it is optional.

yolile:
closes #325

### DIFF
--- a/docs/history/changelog.md
+++ b/docs/history/changelog.md
@@ -114,6 +114,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
   * [#1165](https://github.com/open-contracting/standard/pull/1165) `statusDetails` to `Tender`, `Award` and `Contract`
   * [#1125](https://github.com/open-contracting/standard/pull/1125) `Item.unit.weight`
   * [#1326](https://github.com/open-contracting/standard/pull/1326) `links`
+  * [1376](https://github.com/open-contracting/standard/pull/1376) `publisher`
   * [#1208](https://github.com/open-contracting/standard/pull/1326) The estimated and maximum values of framework agreeemnts:
     * `tender.maximumValue`. Previously, `tender.value` was used for the maximum value. However, this led to double-counting.
     * `awards.maximumValue`. Previously, `awards.value` was used for the maximum value. However, this led to double-counting.
--- a/docs/history/changelog.md
+++ b/docs/history/changelog.md
@@ -114,7 +114,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
   * [#1165](https://github.com/open-contracting/standard/pull/1165) `statusDetails` to `Tender`, `Award` and `Contract`
   * [#1125](https://github.com/open-contracting/standard/pull/1125) `Item.unit.weight`
   * [#1326](https://github.com/open-contracting/standard/pull/1326) `links`
-  * [1376](https://github.com/open-contracting/standard/pull/1376) `publisher`
+  * [#1376](https://github.com/open-contracting/standard/pull/1376) `publisher`
   * [#1208](https://github.com/open-contracting/standard/pull/1326) The estimated and maximum values of framework agreeemnts:
     * `tender.maximumValue`. Previously, `tender.value` was used for the maximum value. However, this led to double-counting.
     * `awards.maximumValue`. Previously, `awards.value` was used for the maximum value. However, this led to double-counting.
--- a/schema/dereferenced-release-schema.json
+++ b/schema/dereferenced-release-schema.json
@@ -27,7 +27,7 @@
     },
     "publisher": {
       "title": "Publisher",
-      "description": "The original publisher of this release. Use this field to identify the original publisher of a release published as part of a package containing releases from different publishers.",
+      "description": "The original publisher of this release.",
       "type": "object",
       "properties": {
         "name": {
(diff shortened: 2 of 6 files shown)


schema/release-schema.json
```diff
@@ -25,6 +25,46 @@
       "format": "date-time",
       "omitWhenMerged": true
     },
+    "publisher": {
+      "title": "Publisher",
+      "description": "Information to uniquely identify the publisher of this specific release. This field is ought to be populated when a release package could contain releases from different publishers.",
```
duncandewhurst: Given that `publisher` is a property of a release, I don't think it's necessary to write 'specific release'. I added 'original' based on https://github.com/open-contracting/standard/issues/325#issuecomment-490661722.

I think it's better to omit the sentence with the condition about when the populate the field and encourage this field to be populated in all cases. Otherwise, republishers would need to add the field, in which case they would be changing a release, which should be immutable.

I added an alternative second sentence to preserve the purpose of the field and simplified the first sentence to reduce repetition.

```suggestion
      "description": "The original publisher of this release. Use this field to identify the original publisher of a release published as part of a package containing releases from different publishers.",
```

schema/release-schema.json
```diff
@@ -25,6 +25,46 @@
...
+        },
+        "uri": {
+          "title": "URI",
+          "description": "A URI to identify the publisher.",
+          "type": [
+            "string",
+            "null"
+          ],
+          "format": "uri"
+        }
```
duncandewhurst: I'm not sure whether it's better here to align with the publisher object in the package schema, or to use the `Identifier` object from the release schema.

duncandewhurst (changes requested):

schema/release-schema.json
yolile: I'm not sure either, maybe reusing the Identifier object is the best approach indeed. @jpmckinney what do you think?

docs/history/changelog.md
```diff
@@ -114,6 +114,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
   * [#1165](https://github.com/open-contracting/standard/pull/1165) `statusDetails` to `Tender`, `Award` and `Contract`
   * [#1125](https://github.com/open-contracting/standard/pull/1125) `Item.unit.weight`
   * [#1326](https://github.com/open-contracting/standard/pull/1326) `links`
+  * [1376](https://github.com/open-contracting/standard/pull/1376) `publisher`
```
```suggestion
  * [#1376](https://github.com/open-contracting/standard/pull/1376) `publisher`
```

schema/release-schema.json
jpmckinney: The Identifier object (which is really an OrganizationIdentifier), would be more consistent.

This will be inconsistent with the release publisher extension, but not many publishers are using it (it is an extension anyway): https://extensions.open-contracting.org/en/extensions/releasePublisher/master/schema/

Please create an issue on the extension registry to decide whether to de-register that extension after 1.2, or add a warning that it is incompatible with 1.2 (which we might have to do for others anyway).

schema/release-schema.json
```diff
@@ -25,6 +25,46 @@
       "format": "date-time",
       "omitWhenMerged": true
     },
+    "publisher": {
+      "title": "Publisher",
+      "description": "The original publisher of this release. Use this field to identify the original publisher of a release published as part of a package containing releases from different publishers.",
```
jpmckinney: Avoid instructions to publishers ("Use this field"). Just describe the semantics. The first clause is sufficient.

Also, since packages will be optional #1084, but bulk JSON lines will become an option, we shouldn't refer to packages even if we keep a second sentence.

schema/release-schema.json
yolile: Changed to `Identifier` and I created the issue https://github.com/open-contracting/extension_registry/issues/135

schema/release-schema.json
```diff
@@ -27,38 +27,18 @@
...
-          "description": "A URI to identify the publisher.",
-          "type": [
-            "string",
-            "null"
-          ],
-          "format": "uri"
+        "identifier": {
+          "title": "Identifier",
+          "description": "The identifier for this organization.",
+          "$ref": "#/definitions/Identifier"
```
jpmckinney: I think the entire object should be a `$ref` to `Identifier`. We don't need both a name and legalName.

schema/dereferenced-release-schema.json
```diff
@@ -26,68 +26,53 @@
       "omitWhenMerged": true
     },
     "publisher": {
-      "title": "Publisher",
-      "description": "The original publisher of this release.",
+      "title": "Identifier",
+      "description": "A unique identifier for a party (organization).",
```
jpmckinney: Hmm, I think my intention is for the title and description of the referring object to be preserved. I've solved this elsewhere, so I can have a look at the manage.py command.

jpmckinney: I've now added a commit.

yolile: Hmm now, all the deprecations are in the party object too, is that correct/expected?

jpmckinney: They're expected in `buyer`, etc., but not in `parties`. Where are you seeing it?

yolile: Oh you are right, they are only in `buyer`, etc and not in `parties` itself

jpmckinney (approved):

PR #1380, opened 2021-08-10
release schema: Add Document.languages, deprecate Document.language

### LINKED ISSUE #930: Publishing a document available in multiple languages
Afghanistan has two official languages, and may therefore want to publish a document that is available in both languages at the same URL e.g. contracts. 

However, `document.language` is a string rather than an array of strings. 

For options now, @duncandewhurst suggested these alternatives:
- If one language is the primary language in any sense, then put this in the `.language` field and explain in the publication policy and/or `.description` field that the document is also provided in the secondary language.
- Add two entries to the `.documents` array, one for each language.

For the future, @jpmckinney suggested that one approach to avoid having two entries for the same document is to add `.languages` (array) and deprecate `.language` (string).

See also: CRM-5197.

duncandewhurst:
Closes #930


mrshll1001 (approved):
All looks well, simple changes and couldn't spot any glaring errors.

yolile:
@duncandewhurst I think that we should update all the JSON examples that use `document.language`, as part of this PR

docs/history/changelog.md
```diff
@@ -125,12 +125,14 @@ Per the [normative and non-normative content and changes policy](https://docs.go
...
     * `contracts.estimatedValue`
   * [#1372](https://github.com/open-contracting/standard/pull/1372) `Address.country`
+  * [#1380](https://github.com/open-contracting/standard/pull/1380) `Document.languages`
 
 * Deprecate some fields:
   * [#1200](https://github.com/open-contracting/standard/pull/1200) `tender.submissionMethod`, because all codes from the `submissionMethod` codelist are deprecated.
   * [#1296](https://github.com/open-contracting/standard/pull/1296) `tender.eligibilityCriteria` in favor of the new `tender.exclusionGrounds` field, in order to use more common terminology and improve semantics. 
   * [#1372](https://github.com/open-contracting/standard/pull/1372) `Address.countryName` in favor of the new `Address.country` field, to promote standardized country codes instead of non-standardized country names.
   * [#1389](https://github.com/open-contracting/standard/pull/1389) `initiationType`, because the approach to data modelling that it supports was not pursued.
+  * [#1380](https://github.com/open-contracting/standard/pull/1380) `Document.language` in favor of the new `Document.languages` field, in order to support documents that contain multiple languages.
```
```suggestion
  * [#1380](https://github.com/open-contracting/standard/pull/1380) `Document.language` in favor of the new `Document.languages` field, to support documents in which multiple languages are used.
```

schema/release-schema.json
```diff
@@ -1111,7 +1124,11 @@
             "null"
           ],
           "codelist": "language.csv",
-          "openCodelist": true
+          "openCodelist": true,
+          "deprecated": {
+            "description": "This field is deprecated in favor of `languages` in order to support documents that contain multiple languages.",
```
```suggestion
            "description": "This field is deprecated in favor of `languages`, to support documents in which multiple languages are used.",
```

schema/release-schema.json
```diff
@@ -1103,6 +1103,19 @@
           "codelist": "mediaType.csv",
           "openCodelist": true
         },
+        "languages": {
+          "title": "Languages",
+          "description": "The languages of the linked document, from the open [language](https://standard.open-contracting.org/{{version}}/{{lang}}/schema/codelists/#language) codelist. A [BCP47 language tag](https://www.w3.org/International/articles/language-tags/) is allowed, if there is a user need for the additional information.",
```
```suggestion
          "description": "The languages used in the document, from the open [language](https://standard.open-contracting.org/{{version}}/{{lang}}/schema/codelists/#language) codelist. A [BCP47 language tag](https://www.w3.org/International/articles/language-tags/) is allowed, if there is a user need for the additional information.",
```

jpmckinney (commented):
Agree with Yohanna on updating examples. If the field is irrelevant to the example, I prefer to delete it, since it's just extra content a reader then needs to skip over.

duncandewhurst:
I reviewed all the markdown files in the documentation for references to document language in JSON examples and I couldn't find any examples to which the language specifically is relevant.

There is one markdown file which includes references to documents in JSON examples: `docs/guidance/build/change_history.md`, which expands the `tender.documents` array in the following examples:

- `docs/examples/tender.json`
- `docs/examples/records/tender.json`
- `docs/examples/tenderUpdate.json`
- `docs/examples/records/tenderUpdate.json`

Each JSON file contains a fully-populated `Document` object, but the text in the markdown file only mentions links to documents. Should we remove all but the `url` and `id` properties, since the other properties aren't relevant to change history, or should we keep them to show good practice of populating the other document fields and not just providing a URL?

jpmckinney:
We can keep those as-is. Having just `id` and `url` would not be a good implementation.

My earlier comment was specifically about `language`, since it is just metadata about the document – it is not especially informative in terms of the contracting process, so it doesn't need to appear in most cases.

duncandewhurst:
I've removed `language` from the examples to which it isn't relevant and updated it to `languages` elsewhere, with the exception of the examples that will be removed in https://github.com/open-contracting/standard/issues/1408 (`extended.json`, `minimal_updates/award.json`, `minimal_updates/tender.json` and minimal_updates/tenderUpdate.json`).

jpmckinney:
There were many files in which `language` wasn't updated. I simply removed both `languages` and `language` as there is no example to which it is relevant.

jpmckinney:
Hmm, nevermind - those might have been the top-level language field. Anyway, it will not be missed.

jpmckinney (approved):

duncandewhurst:
Yes, most examples had one or more top-level `language` fields. For the remaining occurrences, I was following your suggestion to leave the examples in which the `Document` object is referenced as-is, with the exception of updating to use `languages` rather than `language`.

PR #1382, opened 2021-08-10
Order JSON Schema keywords

jpmckinney:
I explain the preferred order in the comments in the sort_keywords function.

No changelog entry as re-ordering keywords is not a normative change.


yolile (approved):

jpmckinney:
To resolve a conflict, I reverted, committed only the management command, merged with 1.2-dev, and then ran the pre-commit command. So, nothing will have changed - the merge was just easier to resolve this way without error.

jpmckinney:
Hmm, merging this will conflict with several open PRs. I'll leave it open until they are merged. Then, I can do the same as I did before (revert, run pre-commit, commit).

duncandewhurst:
Adding blocked label, as I guess we'll want to do this once all other PRs are merged

PR #1384, opened 2021-08-11
Miscellaneous improvements to Guidance steps

### LINKED ISSUE #1379: Guidance suggestions
Whilst reviewing the publisher checklist in CRM-3919, I noticed a few areas in which we could improve or add to the implementation guidance:

1. [Split up the work](https://standard.open-contracting.org/latest/en/guidance/map/#split-up-the-work): Rename to **Complete your mapping** or **Document your mapping**. Otherwise, we have a step for completing the technical assessment ([Identify your data sources](https://standard.open-contracting.org/latest/en/guidance/map/#identify-your-data-sources)) but no step for actually completing the field-level and codelist mappings.
2.  [Perform final checks](https://standard.open-contracting.org/latest/en/guidance/publish/#perform-final-checks): Add 'Check that your package metadata conforms to the schema'. Publishers frequently populate the package metadata incorrectly, or forget to update placeholder values, so we should include a prompt here.
3. [Finalize your publication policy](https://standard.open-contracting.org/latest/en/guidance/publish/#finalize-your-publication-policy): Add 'a description of any differences in the contracting processes and fields available via different access methods and formats'. This point was prioritised in the work around the original publisher checklist and is not covered elsewhere in the documentation.

jpmckinney:
closes #1379

### DIFF
--- a/docs/guidance/design.md
+++ b/docs/guidance/design.md
@@ -88,9 +88,9 @@ Implementing OCDS is easier when your team combines both policy and technology e
 To achieve success, an OCDS implementation team ought to include the following roles, though one individual might play several roles:
 
 * A **project manager** who will oversee the project timeline and deliverables.
-* A **data expert** who will identify where data put forward for publication in OCDS exists at a technical level. They need to be familiar with the IT systems that capture and store contracting data and related documents.
-* A **procurement expert** who will identify where data put forward for publication in OCDS exists at a semantic level. They need to be familiar with procurement legislation and procedures.
-* A **policy champion** who can help navigate the policy decisions of publishing and using OCDS data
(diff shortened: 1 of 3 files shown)


yolile:
What about something like:

> As described in the Field-Level Mapping Template Guidance (introduced below), you will need **to involve at least the following roles from your [team](https://standard.open-contracting.org/latest/en/guidance/design/#build-your-team)**:

for https://standard.open-contracting.org/latest/en/guidance/map/#involve-the-right-people

jpmckinney:
The challenge is that a "technical expert who understands the data structures in OCDS" might not exist in the team at the Design stage. This is instead a role that is developed through implementation.

yolile:
Hmm true, all good then, thanks!

yolile (approved):

duncandewhurst (approved):

PR #1386, opened 2021-08-11
Guidance: Add link to OCDS implementation checklist.

duncandewhurst:
Add resource from CRM-3919

Don't merge until resource published.


yolile:
The resource is already published in both, English and Spanish

jpmckinney (approved):

PR #1387, opened 2021-08-12
Address Primer feedback

### LINKED ISSUE #1370: Primer feedback
Great to see that the Primer is live. I noticed a few issues from a quick read-through. Some are readability/copy-editing issues, some are more substantive. Happy to do a more detailed review, if that would be helpful.

https://standard.open-contracting.org/latest/en/primer/

* [x] Paragraph 1 - the bullet points mix 'OCDS' and 'the OCDS'
* [x] Paragraph 2 & 3 - repetition of 'detailed documentation to guide you in implementing the standard'

https://standard.open-contracting.org/latest/en/primer/what/

* [x] Objectives - the first paragraph isn't an objective. It speaks to the main purpose of the page, to describe what OCDS is. It would be better outside of the objectives box.
* [x] Paragraph 1 - repetition of 'Open Contracting, 'Open Data', 'Data Standard' in the first sentence and in the subsequent bullet points

https://standard.open-contracting.org/latest/en/primer/how/

* Objectives - similar issue with the first paragraph. I think it should be outside of the box.
* [x] Diagrams - the diagrams are inconsistent. One includes the planning stage, one doesn't. This could be confusing because the text doesn't address the inconsistency. In any case, there is a lot of repetition between the diagrams, so I think one diagram would suffice to communicate both the stages of a contracting process and the purpose of the ocid.
* [x] Paragraph 1 - the definition of a contracting process matches neither the old 1.1 definition or the new 1.2 definition from https://github.com/open-contracting/standard/pull/1208. It's confusing to say 'aimed at concluding one or more contracts' when the stages/diagrams include implementation.
* [x] Paragraph 4 & 5 - repetition of 'provides a common framework'
* [x] Paragaph 5 & 8 - repetition of 'the default format of the data is JSON'
* [x] Paragraph 6 - typo: 'When mapping your data to the OCDS our **sing** OCDS data'

https://standard.open-contracting.org/latest/en/primer/releases_and_records/

* Objectives - similar issue with the first paragraph. I think it should be outside of the box.
* [x] Diagram - the diagram only shows OCDS releases. The diagram in the draft shared with the helpdesk shows how releases are linked together in a record:
image
* [x] Final paragraph - the first bullet is confusing because it uses 'documents' to refer to documents linked from `tender/documents` etc., but everywhere else on the page 'documents' refers to a release or record JSON document.

jpmckinney:
closes #1370

- primer: Use "the OCDS" consistently where applicable
- primer: Remove paragraphs and bullets that repeat each other and the content on the landing page and on the last page of the primer
- primer/how: Remove image that is inconsistent due to inclusion of planning, and that is somewhat repetitive
- primer,guidance: Use "JSON text" instead of "JSON document" to avoid confusion with the documents field
- primer/releases_and_records: Use image containing both releases and records
- primer/index: Restore bullets, since the Primer also serves to outline the documentation, using same content as primer/next
- primer/what: Move significant content out of objectives. Avoid repetition of concepts. Add link to academic research. Capitalize Primer.
- primer/next,primer/releases_and_records: Add article. Capitalize Primer.
- primer/how: Align with #1208 (concluding -> implementing). Add link to demand-side assessment. Clarify normative relationship between JSON and Excel/CSV. Rephrase to avoid repetition. Capitalize Primer. Copy-edit punctuation. Fix typo.

### DIFF
--- a/docs/primer/releases_and_records.md
+++ b/docs/primer/releases_and_records.md
@@ -3,7 +3,7 @@
 ```{admonition} Objectives
 :class: note
 
-Open Contracting Data Standard (OCDS) data is published and updated over time using two JSON document types: “releases” and “records”.  This page will:
+Open Contracting Data Standard (OCDS) data is published and updated over time using two types of JSON text: “releases” and “records.” This page will:
 
 * Outline the purpose of releases and records
 * Describe how to publish a change history using releases and records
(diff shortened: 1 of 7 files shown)


docs/primer/releases_and_records.md
```diff
@@ -3,7 +3,7 @@
 ```{admonition} Objectives
 :class: note
 
-Open Contracting Data Standard (OCDS) data is published and updated over time using two JSON document types: “releases” and “records”.  This page will:
+Open Contracting Data Standard (OCDS) data is published and updated over time using two types of JSON text: “releases” and “records.” This page will:
```
duncandewhurst:
```suggestion
Open Contracting Data Standard (OCDS) data is published and updated over time using two types of JSON text: “releases” and “records”. This page will:
```

docs/primer/releases_and_records.md
```diff
@@ -14,22 +14,22 @@ Open Contracting Data Standard (OCDS) data is published and updated over time us
 <iframe width="560" height="315" src="https://www.youtube-nocookie.com/embed/OeDCLMWlY_U" title="Introducing releases and records" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
 </div>
 
-The only thing that can be truly called “OCDS data” is a JSON document that validates against the OCDS schema. The JSON documents that OCDS uses are called “releases” and “records.”
+The only thing that can be truly called “OCDS data” is JSON text that validates against the OCDS schema. The JSON texts that OCDS uses are called “releases” and “records.”
```
duncandewhurst: I'm not sure about the use of 'JSON text'. Whilst, strictly speaking, it is the correct term to cover both JSON files and JSON served via an API, I think it could be confusing for less technical readers who might associate 'text' with unstructured data.

My feedback was simply to remove the bullet point about (non-JSON) documents from this page that is otherwise about releases and records, to which documents aren't relevant.

I guess we adopted 'JSON document' from https://json-schema.org/. I don't remember it causing any confusion previously so I'm not sure we need to change it.

If we do need to change 'JSON document', I think 'JSON data' would be a better choice. For example, this paragraph could read:

> The only thing that can be truly called “OCDS data” is JSON data that validates against the OCDS schema. There are two schemas for OCDS data: “releases” and “records.”

docs/primer/releases_and_records.md
jpmckinney: That's British style punctuation. We use American style punctuation and spelling.

docs/primer/releases_and_records.md
jpmckinney: I've committed two changes. You didn't feed back to remove the bullet in #1370 - you just said it was confusing :) The bullet is now removed. To avoid going back and forth on whether "JSON document" is confusing, we now never use "JSON document" in the documentation.

duncandewhurst: Looks good!

duncandewhurst (approved):

PR #1389, opened 2021-08-19
Deprecate initiationType

### LINKED ISSUE #821: Deprecate initiationType
The current definition ("An open competitive bidding or tendering to form contracts") is overly restrictive; this definition can only be respected if `tender/procurementMethod` is 'open'.

----

More broadly, the design reason for the `initiationType` field is best described here: https://github.com/open-contracting/standard/issues/111#issuecomment-61224721

In other words (and from memory), this field exists because it was anticipated that the `tender` block would be inappropriate for some types of contracting processes. This field was added in order to indicate to data users that different features would be available within the data for such processes. For example, if `initiationType` were 'lottery', then data users would look for a `lottery` block instead of a `tender` block.

However, in practice, the `tender` field has been implemented according to its broad definition ("The activities undertaken in order to enter into a contract."), i.e. for initiation/solicitation in general. This seems like the right approach, instead of having a different block for each kind of contracting process, which makes data use harder. To my knowledge, implementers have not yet encountered contracting processes for which the `tender` block is wholly inappropriate (processes with multiple stages have issues, but the `tender` block is still appropriate for at least one stage).

A related issue might therefore be to re-assess the design approach behind the `initiationType` field more broadly. In the meantime, besides correcting the code, we might want to treat the field and codelist as 'legacy'.

cc @duncandewhurst

yolile:
closes #821


yolile:
I guess we can express the reason for deprecation in a better way, I'd be glad to hear ideas!

jpmckinney:
I re-did the commits instead of resolving the conflict with the old getting started page (which is now deleted).

jpmckinney:
I think in the schema, changelog, and code, the reason is "[This [field/code] is deprecated,] because the approach to data modelling that it supports was not pursued."

The idea was that an application would read `initiationType` and then determine that it then needs to read `tender`, or `ppp` or `contest`, etc.

But contracting processes are more similar than they are different, and it makes more sense to have a single object to which procedure-specific fields are added, instead of distinct objects. This approach is described (internally) here: https://ocds-standard-development-handbook.readthedocs.io/en/latest/standard/adr/implicit_sublasses.html

Once the deprecation reason is updated, I am happy to approve the PR.

docs/schema/codelists.md
```diff
@@ -194,7 +194,8 @@ The country codelist is used to provide the country component of an address, usi
 
 ### Initiation Type
 
-Contracting processes can be formed under a number of different processes. Currently, only 'tender' is supported in this codelist. Future versions of the standard might support other initiation types. The initiation type is used to provide information to consuming applications on the different blocks of data and releases they can expect from a contracting process.
```
jpmckinney, line 197: We didn't remove the sentence for the deprecated `submissionMethod`. That said, I'm fine with removing this paragraph, because a single-value closed codelist is meaningless anyway. At least for `submissionMethod`, a user of OCDS 1.1 data can use that sentence to interpret the publisher's intent.

yolile: Aha, and I guess "Future versions of the standard might support other ... " isn't true anymore.

docs/history/changelog.md
```diff
@@ -100,6 +100,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
 * [#1238](https://github.com/open-contracting/standard/pull/1238) Open the `releaseTag` codelist.
 * [#1353](https://github.com/open-contracting/standard/pull/1353) Replace "submit a tender" with "submit a bid" in the `method` codelist.
 * [#1209](https://github.com/open-contracting/standard/pull/1209) Replace "electronic goods" with "digital goods" in the description of 'goods' from the `procurementCategory` codelist, to align with the description in the World Trade Organization's Agreement on Government Procurement, and to avoid confusion between electronic goods like computers and digital goods like software.
+* [#1389](https://github.com/open-contracting/standard/pull/1389) Deprecate the code of the `initiationType` codelist, because the approach to data modelling that it supports was not pursued.
```
jpmckinney: Maybe also change the entry for submissionMethod to match.

```suggestion
* [#1389](https://github.com/open-contracting/standard/pull/1389) Deprecate the `initiationType` codelist, because the approach to data modelling that it supports was not pursued.
```

yolile: I've updated both, let me know if the `submissionMethod` one is what you expected

PR #1393, opened 2021-08-24
record-package-schema: remove releases in record requirement

### LINKED ISSUE #1092: Make releases optional on Record definition
## Background

For some publishers, it's only feasible/possible to prepare the compiled release – that is, a release that provides the full information about a contracting process at a given time. This might be the case if, for example, the IT system doesn't implement versioning, events, good identifiers and/or proper controls (e.g. it allows a user to just change a supplier's name instead of requiring the user to register a new supplier in the system). This also might be the case for publishers who collect non-OCDS data from a source and re-publish it as OCDS, since they don't have direct access to the source's system. 

Presently, some publishers publish these kinds of "full" releases as individual releases, e.g. each time a significant event occurs in the contracting process.

A problem can occur if new information has different array entries from earlier information, that were intended to replace that earlier information. Due to how the merge routine works for most arrays of objects, the new information wouldn't replace it, but would append to it. Due to the data quality and system limitations described earlier, it might not be possible to remedy this problem on the publisher's side.

## Discussion

The semantically appropriate way to publish such releases would be as compiled releases. However, the Record definition presently requires an array of releases.

## Proposal

Removed `"releases"` from `"required"` in the Record definition.

yolile:
closes #1092


docs/schema/records_reference.md
```diff
@@ -36,12 +36,22 @@ The following example demonstrates the package metadata and record fields.
...
-A record **should** contain a [compiledRelease](#compiled-release) object, which represents the state of the contracting process at the time of the record's publication.
+* All [releases](#releases) about the contracting process at the time of the record's publication, and/or;
+
+* A [compiledRelease](#compiled-release) object, which represents the state of the contracting process at the time of the record's publication.
+
+As such, a record functions as an index of releases and/or as the latest version of a contracting process at the time of the record's publication.
 
 A record **may** contain a [versionedRelease](#versioned-release) object, which aggregates, into a single object, all values of all fields from all releases up to the time of the record's publication. The versioned release is designed to make it easy to see how values change from one release to another, and will often be generated by data users, rather than by publishers.
 
+```{admonition} Releases
```
yolile: @jpmckinney I'm not sure if we should use `{versionadded} 1.2` here or just `{admonition}` (or nothing?), as in this case we are not adding any new field, but removing an (important!) constraint.

docs/history/changelog.md
```diff
@@ -169,6 +169,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
   * [#1307](https://github.com/open-contracting/standard/pull/1307) Clarify uniqueness rules for records.
   * [#1339](https://github.com/open-contracting/standard/pull/1339) Deprecate `packages`.
   * [#1374](https://github.com/open-contracting/standard/pull/1374) Remove `records.minItems` requirement.
+  * [#1393](https://github.com/open-contracting/standard/pull/1393) Remove `records.releases` requirement.
```
```suggestion
  * [#1393](https://github.com/open-contracting/standard/pull/1393) Remove `releases` as a required field of the `Record` object.
```

docs/schema/records_reference.md
jpmckinney: I think it's fine to just update the above text without this admonition.

docs/schema/records_reference.md
```diff
@@ -36,12 +36,22 @@ The following example demonstrates the package metadata and record fields.
 
 ## Record structure
 
-A record **must** contain an [ocid](identifiers.md#open-contracting-process-identifier-ocid) and all [releases](#releases) about the contracting process at the time of the record's publication. As such, a record functions as an index of releases about a contracting process.
+A record **must** contain an [ocid](identifiers.md#open-contracting-process-identifier-ocid) and it **should** contain either:
```
jpmckinney: We don't want just an `ocid`. It's possible to express in JSON Schema the requirement to have at least one of `releases` or `compiledRelease`, but it involves `anyOf`, which is harder to read and evaluate. So, for now, we can at least have the rule in text.

```suggestion
A record **must** contain an [ocid](identifiers.md#open-contracting-process-identifier-ocid) and one or more of:
```

docs/schema/records_reference.md
```diff
@@ -36,12 +36,22 @@ The following example demonstrates the package metadata and record fields.
...
 
-A record **should** contain a [compiledRelease](#compiled-release) object, which represents the state of the contracting process at the time of the record's publication.
+* All [releases](#releases) about the contracting process at the time of the record's publication, and/or;
+
```
jpmckinney:
```suggestion
* All [releases](#releases) about the contracting process at the time of the record's publication.
```

docs/schema/records_reference.md
```diff
@@ -36,12 +36,22 @@ The following example demonstrates the package metadata and record fields.
...
+* A [compiledRelease](#compiled-release) object, which represents the state of the contracting process at the time of the record's publication.
+
+As such, a record functions as an index of releases and/or as the latest version of a contracting process at the time of the record's publication.
```
jpmckinney: The original wording was repetitive. We don't need it.

```suggestion
```

jpmckinney (commented):
It's no longer clear that a compiledRelease "should" be included when releases are included, but we've learned that compiled releases are rarely included, or are incorrectly calculated, so I think just the original releases is fine.

jpmckinney (approved):

PR #1401, opened 2021-08-28
Replace total value with value

### LINKED ISSUE #1207: Value fields: remove "total"
Value fields in OCDS often use the term "total". I am not sure what it means and I think it might be misleading, e.g. because a "total" value field might be tax-less (see https://github.com/open-contracting/standard/issues/817) or represent just a single award, contract a lot from a contracting process that contains more of them. 

I would suggest replacing: 
- "total estimated value" with "estimated value" in `tender.value` 
- "total value" with "value" in `award.value` and `contract.value`.

JachymHercher:
closes #1207


docs/history/changelog.md
```diff
@@ -188,6 +190,8 @@ Per the [normative and non-normative content and changes policy](https://docs.go
 * [#1315](https://github.com/open-contracting/standard/pull/1315) Add rules on setting `id` and `date` for compiled releases to the merging specification.
 * [#1344](https://github.com/open-contracting/standard/pull/1344) Add contract suspension worked example.
 * [#1375](https://github.com/open-contracting/standard/pull/1375) Update guidance for empty fields in the merging documentation.
+* [#1401](https://github.com/open-contracting/standard/pull/1401) Replace "total value" with "value".
+
```
jpmckinney: Once is enough. It can be assumed that any change to the schema might be reflected in the documentation.

```suggestion
```

JachymHercher: Maybe worth adding a point to https://ocds-standard-development-handbook.readthedocs.io/en/latest/meta/changelog_style_guide.html ? Is the rule "If you change the schema changelog, don't also change the documentation changelog?" 

We could also rename the "Schema" section of the changelog to "Schema and corresponding documentation" and "Documentation" to "Documentation (only)"?

jpmckinney: I've now added this point:

> If changes are made to files under the ``schema`` directory, it is assumed that corresponding changes were made to files under the ``docs`` directory. Do not add an entry under the "Documentation" heading if the changes directly correspond to the entry under the "Codelists" and/or "Schema" headings.

jpmckinney (approved):

PR #1404, opened 2021-08-28
Moldova worked example: remove incorrect 'procuringEntity'

### LINKED ISSUE #571: Edit tender.procuringEntity description

JachymHercher:
Closes 571, in particular https://github.com/open-contracting/standard/issues/571#issuecomment-907664840

### DIFF
--- a/docs/examples/organizational-units/moldova-tender.json
+++ b/docs/examples/organizational-units/moldova-tender.json
@@ -53,10 +53,6 @@
         "procurementMethod": "open",
         "mainProcurementCategory": "goods",
         "hasEnquiries": false,
-        "procuringEntity": {
-          "name": "National Bank of Moldova - Chişinău Branch",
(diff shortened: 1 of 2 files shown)


jpmckinney:
@JachymHercher I think you missed some parts. I added another commit. Please comment if approved (since you opened the PR, you can't submit a review; and reactions like 👍 don't send a notification, so I need a comment).

JachymHercher:
Oh my, sorry and thanks for the correction.

PR #1415, opened 2021-09-19
Update stage object descriptions, release tags and planning process description

### LINKED ISSUE #866: Clarify definition of contracting process

### LINKED ISSUE #1385: Update all stage object descriptions and release tags
This issue is split from #1160. Copying @JachymHercher's comment (originally at https://github.com/open-contracting/standard/issues/1160#issuecomment-881950573):

----

The issue started with `awardStatus`, moved to [block]Status in general, then we realized that we needed to better define [blocks]. I've started from the back, so below are some observations and a proposal on new definitions for the blocks. 

1) I don't think we need a perfectly hard definition, time-wise, of where each stage starts and finishes, because a) each stage is precisely defined by the data fields it contains; b) it's hard to do. Consequently, I do not put the time borders in the first sentence of the description but only later on and introduce them with "Typically". 
1) Above, we didn't really address the question of where the planning stage starts and ends (i.e. the tender stage starts). This corresponds to the boundary between a planning and contracting process (i.e. the first "action towards concluding/implementing one or more contracts"). Nor did we discuss where the contract stage / contracting process ends.
    a.) I propose not to define the start of the planning process, because it would necessarily be very fuzzy (e.g. "After the first idea to purchase / fulfill a need.") 
    b.) Similarly, I propose not to define the end of the contracting process, because it is not necessary. The implicit approach of "publish as long as it makes sense" might be good enough. (If we wanted to go for it, it is actually a quite tricky question - I would propose "Until the end of any legal obligations stemming from the contract (e.g. maintenance, warranty, options, etc.)")
    c.) For the border between planning/tender, I use "publication of the procurement documents", together with "typically". @jpmckinney in https://github.com/open-contracting/standard/issues/1160#issuecomment-773687654 proposed also "reservation of the budget", but I would rather avoid it, as I'm not sure how widely understood "reservation" is (versus proposal, approval and apportionment) and because we only have budget information in the `Planning` block.
3) I'm also wondering about the difference between the descriptions of the blocks and the same/similar named objects nested within them. For contracts and awards, these correspond to concrete real-world objects and we have separate issues working on their definitions. For planning and tender, I take them as somewhat artificial, and would deal with it by having the descriptions identical for both blocks and objects. 
4) eForms allows bid submissions deadlines (equivalent to `tenderPeriod`) to differ per lot. OCDS currently does not (not even with the lot extensions), but once it does this will interfere with the definition of the tender stage, beacuse it currently assumes that there is a single submission deadline for the entire contracting process. I suggest we deal with it when we get there.

### Current block descriptions  

`planning`: "Information from the planning phase of the contracting process. This includes information related to the process of deciding what to contract, when and how.",

 `tender`: "The activities undertaken in order to enter into a contract.",

 `awards`: "Information from the award phase of the contracting process. There can be more than one award per contracting process e.g. because the contract is split among different providers, or because it is a standing offer.",
      
 `contracts`: "Information from the contract creation phase of the procurement process.",

### Proposed descriptions, wording 1

`planning`, `Planning`, `relaseTag` code 'planning': Planning process information about, for example, the rationale for the procurement, the budget, forecasting and market research. Typically, this information comes from actions before the publication of the procurement documents.

`tender`, `Tender`, `relaseTag` code 'tender': Contracting process information about, for example, the items to be purchased, their value, procurement method, award criteria, and various deadlines. Typically, this information comes from actions starting with the publication of the procurement documents and ending with the bid submission deadline. Alternatively, this block is used during the planning process before the procurement documents are published, e.g. to give the market a general description of the purchase or a general estimate of the purchase's value. 

`awards`, `relaseTag` code 'award': Contracting process information about the award. Typically, this information comes from actions after the bid submission deadline and ending with the award itself or, if there is a standstill period, the end of the standstill period.

`contract`, `relaseTag` code 'contract': Contracting process information about the contract and its implementation. Typically, this information comes from actions taken after the award and, if there was a standstill period, after the end of the standstill period.

### Proposed descriptions, wording 2
(This wording mentions the type of process later on. While not going "from general information to concrete information", I find it better readable.)

`planning`, `Planning`, `relaseTag` code 'planning': Information about, for example, the rationale for the procurement, the budget, forecasting and market research. This information concerns the planning process and typically comes from actions before the publication of the procurement documents.

`tender`, `Tender`, `relaseTag` code 'tender': Information about, for example, the purchased items, their value, procurement method, award criteria, and various deadlines. This information can concern the contracting process or the planning process. In contracting processes, it typically comes from actions starting with the publication of the procurement documents and ending with the bid submission deadline. In planning processes, it typically comes from actions taken before the procurement documents are published, e.g. to give the market a general description of the purchase or a broad estimate of the purchase's value. 

`awards`, `relaseTag` code 'award': Information about the award. This information concerns the contracting process and typically comes from actions after the bid submission deadline and ending with the award itself or, if there is a standstill period, the end of the standstill period. 

`contract`, `relaseTag` code 'contract': Information about the contract and its implementation. This information concerns the contracting process and typically comes from actions taken after the award and, if there was a standstill period, after the end of the standstill period.

----

In my comment above, I've clarified that bid submission falls into the tender stage and not the award stage, as discussed above. Furthermore, I've added alternative wordings and clearer identification of which blocks, fields and codes should change. The comment now covers answers to questions 1-3 in https://github.com/open-contracting/standard/issues/1160#issuecomment-792348486.

----

Questions 1-3 are copied here:

1) Contract definition covers also unsigned contracts - done. 
2) Decide where the tender and award stages end. I think they should end with bid submission and the award decision, respectively.
3) I would suggest: update the definitions of `Planning`, `Tender`, `Awards` and `Contracts` as well as the corresponding values in https://standard.open-contracting.org/latest/en/schema/codelists/#release-tag so that they take into account the start and end points of each stage and thus avoid overlaps. We can also standardize the wording.

JachymHercher:
Closes #1385

Closes #866


schema/release-schema.json
```diff
@@ -32,7 +32,7 @@
     },
     "tag": {
       "title": "Release Tag",
-      "description": "One or more values from the open [releaseTag](https://standard.open-contracting.org/{{version}}/{{lang}}/schema/codelists/#release-tag) codelist. A tag labels the release: for example, as corresponding to a particular stage of the contracting process. Codes outside the releaseTag codelist might indicate, for example, the notice or form to which the release corresponds, or the event that triggered the publication of the release.",
+      "description": "A tag labeling the release (for example, as corresponding to a stage of the contracting process), using the the open [releaseTag](https://standard.open-contracting.org/{{version}}/{{lang}}/schema/codelists/#release-tag) codelist. Codes outside the releaseTag codelist might indicate, for example, the notice or form to which the release corresponds, or the event that triggered the publication of the release. Planning processes must not contain the 'tender' code, even if they include `Tender`.",
```
jpmckinney: The edit is to the last sentence, so you will have to scroll.

```suggestion
      "description": "A tag labeling the release (for example, as corresponding to a stage of the contracting process), using the the open [releaseTag](https://standard.open-contracting.org/{{version}}/{{lang}}/schema/codelists/#release-tag) codelist. Codes outside the releaseTag codelist might indicate, for example, the notice or form to which the release corresponds, or the event that triggered the publication of the release. Planning processes must not use the 'tender' code, even if they use `tender` fields.",
```

jpmckinney (commented):
Just one suggestion. @yolile, feel free to merge if the suggestion is applied.

jpmckinney:
Regarding:

> Planning processes must not use the 'tender' code, even if they use `tender` fields.

I think the use of "must" here is okay (typically, new "must" statements are not possible in a minor release), because a planning process is identified by the release tag or the tender.status (both of which are expected to align), such that the use of the 'tender' code would imply that the planning process is actually a contracting process – meaning that this new rule can't really be broken. That said, it's still useful to express as a rule, to avoid the blurring of the types of processes.

JachymHercher:
> Just one suggestion. @yolile, feel free to merge if the suggestion is applied.

Just to be sure - did you really mean @yolile or me? :)

jpmckinney:
I meant @yolile, so that a reviewer merges, rather than the author :) But I can merge now.

Issue #909, opened 2019-08-08
Pre-qualification for multi-use list


jpmckinney:
## Context

The proposal in this issue is for pre-qualification leading to a [multi-use list](https://www.wto.org/english/docs_e/legal_e/rev-gpr-94_01_e.htm) or similar.

There is another issue for pre-qualification/pre-selection leading to a single request for tender (i.e. the pre-qualification stage produces a single-use list of qualified suppliers): #906

Compared to a single-use list, multi-use lists introduce many more considerations:

* Suppliers: Can new suppliers join the list later? If so, at any time, or at specific times? Can suppliers be removed from the list? How long is the list valid?
* Competition: Are there competitions between listed suppliers, or is there instead a set of rules to determine which supplier is used in which cases (or both)? Can off-list suppliers participate in competitions by submitting a tender, with or without a request to join the list?
* Buyers: Can multiple buyers use the list?

## Background

The GPA defines a multi-use list as:

> a list of suppliers that a procuring entity has determined satisfy the conditions for participation in that list, and that the procuring entity intends to use more than once

There are examples in Paraguay (linked from this comment: https://github.com/open-contracting/standard/issues/906#issuecomment-519671713) that seem to match this definition of a multi-use list.

The [section](https://www.wto.org/english/docs_e/legal_e/rev-gpr-94_01_e.htm#_Toc320546106) of the GPA on multi-use lists has more detail on notices and steps, notably:

> A procuring entity may maintain a multi-use list of suppliers, provided that a notice inviting interested suppliers to apply for inclusion on the list is … made available continuously

> A procuring entity shall allow suppliers to apply at any time for inclusion on a multi-use list

> Where a supplier that is not included on a multi-use list submits a request for participation in a procurement based on a multi-use list … a procuring entity shall examine the request

> A procuring entity … may use a notice inviting suppliers to apply for inclusion on a multi-use list as a notice of intended procurement

Other procedures that use a list of qualified suppliers multiple times include framework agreements; however, framework agreements might have important differences.

Note that the GPA doesn't mention framework agreements, so multi-use lists might be a super-category that covers framework agreements among others.

## Proposal

Given the similarity to framework agreements (the main difference being that a multi-use list doesn't imply a set-up agreement/contract between the government and qualified suppliers), the model for multi-use lists in OCDS should be similar to the one for framework agreements, and in particular open framework agreements (i.e. dynamic purchasing systems).

### First stage model

The proposal follows the same considerations for this first stage as in #906.

[...]

Issue #906, opened 2019-08-07
Pre-qualification/pre-selection for single use


jpmckinney:
## Context

The proposal in this issue is for pre-qualification/pre-selection leading to a single request for tender (i.e. the pre-qualification stage produces a single-use list of qualified suppliers).

There is another issue for pre-qualification/pre-selection leading to a [multi-use list](https://www.wto.org/english/docs_e/legal_e/rev-gpr-94_01_e.htm) or similar: #909

## Background

This issue is to break up #440 into smaller issues. This comment https://github.com/open-contracting/standard/issues/440#issuecomment-516674352 has a proposal for how to model two-stage procedures in OCDS without causing backwards-incompatible changes. This issue specifically focuses on procedures with a pre-qualification or pre-selection stage.

Broadly speaking, such procedures have a first stage in which potential suppliers submit requests to participate. Submitters are qualified/selected based on their submissions, according to qualification/selection criteria. If there is a limit to the number of qualified candidates, then this stage is known as pre-selection instead of pre-qualification, and involves score/rank criteria instead of pass-fail criteria. Qualified/selected candidates then submit tenders in a second stage. (See [UNCITRAL Glossary](http://www.uncitral.org/pdf/english/texts/procurem/ml-procurement-2011/Glossary-e.pdf))

Qualification/selection criteria can presently be expressed using the [requirements extension](https://extensions.open-contracting.org/en/extensions/requirements/master/).

### First stage model

Per https://github.com/open-contracting/standard/issues/440#issuecomment-516674352, the first stage should be modelled using the `tender` block. Indeed, this is already the block to which the [requirements extension](https://extensions.open-contracting.org/en/extensions/requirements/master/) adds a `criteria` field to express qualification/selection criteria, which need to be known in the first stage. (Related: #590)

The [qualification extension](https://extensions.open-contracting.org/en/extensions/qualification/master/codelists/) – which will be deprecated per the linked comment – is the present solution for pre-qualification/pre-selection and has the same structure as the `tender` block, with one difference (mentioned later). Given that it has the same structure, most fields in the `tender` block can be used without modification for the first stage of a procedure with pre-qualification/pre-selection, just as they were in that older solution. `awardCriteria`, `awardCriteriaDetails` and `awardPeriod` will still refer to the award that occurs after the second stage.

As for `tenderPeriod`, `numberOfTenderers` and `tenderers`, we need to make a decision. For `tenderPeriod`, we can either, in order of preference:

[...]

PR #1416, opened 2021-09-21
Use "supplier" consistently, e.g. use "potential supplier"

### LINKED ISSUE #1398: Inconsistent use of 'supplier' in guidance and codelist
'Supplier' is defined as an organization that has already been awarded a contract, both in OCDS 1.1 ("An entity awarded or contracted to provide goods, works or services.", https://standard.open-contracting.org/1.1/en/schema/codelists/#party-role) and in OCDS 1.2 ("An entity with which a buyer or a procuring entity decided to conclude a contract.
", https://github.com/open-contracting/standard/pull/1232). 

However, it is currently sometimes used in the sense of a company (that has not yet submitted a bit, that is submitting a bid). I have found this in the following instances:  
- https://standard.open-contracting.org/1.1/en/schema/codelists/#method (all codes)
- https://standard.open-contracting.org/1.1/en/schema/codelists/#document-type ('tenderNotice')
- https://standard.open-contracting.org/1.1/en/guidance/map/pre-qualification/ (many places)
- https://standard.open-contracting.org/latest/en/guidance/map/buyers_suppliers/#consortia-suppliers (few places)

We should change this to a correct term. There are several options: "potential supplier" (simple, but inelegant), "company" (common, but undefined in `partyRole`), organization (widely used, but too broad), economic operator (popular in the EU, but terribly ugly).

I would suggest replacing it with "potential supplier". It is understandable and concrete, we stick to the terminology defined in `organizationRole`, we don't need to add any new terms.

JachymHercher:
Closes #1398


docs/examples/pre-qualification/multi-stage.csv
```diff
@@ -1,14 +1,14 @@
...
+* Potential suppliers submit requests to participate
 
-* The procuring entity assess suppliers based on their submissions
+* The buyer or procuring entity assesses the requests
 
-* The procuring entity establishes a list of qualified or selected suppliers","* The procuring entity issues an invitation to bid to the suppliers on the list
+* The buyer or procuring entity establishes a list of qualified or selected potential suppliers","* The buyer or procuring entity issues an invitation to bid to the potential suppliers on the list
 
-* Suppliers submit bids
+* Tenderers submit bids
```
jpmckinney: "Tenderers" might make things tricky.

Ideally, "tenderers" would have been reserved for organizations that submit a bid. However, `tender.tenderers` is meant to collect the submitters in the first stage, whether it is a bid or a request to participate (see #906 and #909).

Anyhow, if we use "Potential suppliers" here, then it is consistent with the above, and it also means we're not introducing a new term in this file.

```suggestion
* Potential suppliers submit bids
```

jpmckinney (commented):
Just one comment.

jpmckinney:
I fixed a conflict with the changes in #1418. Before, the diff in this PR was:

Screen Shot 2021-09-21 at 4 11 31 PM

Now, it's just adding the word "potential" for the 'direct' code.

jpmckinney (approved):

PR #1417, opened 2021-09-21
Update `procurementMethod` description

### LINKED ISSUE #1395: `procurementMethod`: improve description and documentation
We should improve the information around `procurementMethod`: 
1) The current description is 
    > The procurement method, from the closed method codelist.

    Which doesn't say much. I would suggest something like: or "The procurement method sets out the organizations that are allowed to submit a bid."    

2) In the text in https://standard.open-contracting.org/1.1/en/schema/codelists/#method, I would remove at least the first sentence "A contracting process aims to fulfill the requirements identified at the planning stage." Concerning the note, see https://github.com/open-contracting/standard/issues/1396. 

3) In line with https://github.com/open-contracting/standard/issues/1301#issuecomment-906715985, we should add a sentence to the description along the lines of "In releases concerning second stages (e.g. in  framework agreements) or other subsequent stages, the procurement method always refers to the first stage of the process."

JachymHercher:
Closes #1395


jpmckinney (approved):
I just fixed a "the the" typo and moved changelog entries (per my commit messages).

Issue #1095, opened 2020-10-22
Improving API guidance 


odscrachel:
In ticket #1009 there was an action to 'Revisit the guidance on bulk downloads and API filters' that included

- Improve API guidance to describe API designs that would allow users to retrieve only new data, e.g. the user can pass a “since” parameter or a cursor.
- We should encourage bulk downloads in JSON format, to avoid having to scrape the API to get the full dataset.
- We should encourage endpoints to get a single release (in a release package) by release id, and a single record (in a record package) by record ocid.
- Publishers might prefer to return a search response that only lists a bit of metadata about each contracting process; the user/client would have to then follow links to get the full details of the contracting process.

Some of this seemed a bit vague to me, so I wanted to check peoples thoughts about what our API advice should be? This is outside my area of expertise, but I'm happy to implement the changes once we have decided on them. 

Current advice is [here](https://standard.open-contracting.org/latest/en/guidance/build/hosting/)

jpmckinney:
At the 2020-02 OCDS Retreat, I think this was mostly discussed by @juanpane and @duncandewhurst.

yolile:
Noting that the [current documentation](https://standard.open-contracting.org/latest/en/guidance/build/hosting/) title is "Data files and APIs" but it barely describes APIs

yolile:
and the other mention of APIs is here https://standard.open-contracting.org/latest/en/guidance/build/system_architectures/#additional-considerations

duncandewhurst:
Closed by https://github.com/open-contracting/standard/pull/1526

yolile:
The PR was merged against 1.2-dev instead of 1.1-dev

jpmckinney:
We can cherry pick the commits into 1.1-dev. I can then merge that new PR.

yolile:
closed by #1553

PR #1422, opened 2021-09-23
Merge pagination extension

### LINKED ISSUE #1327: Merge the pagination extension
Creating a new issue for this comment from #928 since the main topic of that issue will be addressed separately:

> We should also look into merging in https://github.com/open-contracting-extensions/ocds_pagination_extension.

_Originally posted by @jpmckinney in https://github.com/open-contracting/standard/issues/928#issuecomment-660284323_

duncandewhurst:
Closes #1327

`api_responses.md` includes only the content from the pagination extension readme. I figure we can add more framing as part of #1084.


pindec (approved):

pindec (commented):
Also may need to edit /docs/guidance/build.md "Establish your publication formats and access methods" section to remove the line "The OCDS pagination extension describes how to paginate OCDS data via an API."

duncandewhurst:
Good spot! Thanks, @pindec.

duncandewhurst:
@jpmckinney just flagging the outstanding review request, in case you missed it.

jpmckinney:
Not forgotten, just backlogged.

jpmckinney:
@duncandewhurst Ah, right, the reason I hadn't merged it originally was because I wasn't satisfied with the documentation of the pagination extension. We indeed revised it heavily as part of #1095, and that documentation now lives here: https://standard.open-contracting.org/staging/1.2-dev/en/guidance/build/hosting/#pagination

I think we can just remove the API responses page from this PR, change the link, and then review and merge.

duncandewhurst:
Sounds good. I've made those changes so this PR is now ready for review.

jpmckinney (approved):

PR #1427, opened 2021-10-07
Add guidance about recommended extensions.

duncandewhurst:
From a [comment](https://docs.google.com/document/d/1JQCOQzLvwEeue6yE4AugyPXQlsC08-ljfUoBwJ4bNWQ/edit?disco=AAAAP-1SRPA) in the draft guidance from https://github.com/open-contracting/extension-explorer/issues/47


jpmckinney:
"Recommended" currently has that correlation to the governance process, but it will not necessarily, going forward.

"Recommended" doesn't mean much more than that we encourage their use.

duncandewhurst:
> "Recommended" doesn't mean much more than that we encourage their use.

We should clarify that somewhere, at least for the benefit of helpdesk analysts.

jpmckinney:
Happy for suggestions as to where to clarify :)

duncandewhurst:
We could just reword the proposed sentence as:

> You are encouraged to consider whether you can use extensions tagged as **Recommended** in the Extension Explorer.

jpmckinney:
Sounds good. I think a simpler construction is:

> You are encouraged to consider using the extensions tagged as *Recommended* in the Extension Explorer.

jpmckinney (approved):

PR #1431, opened 2021-10-14
Add Link object to schema reference

### LINKED ISSUE #928: Add links field

duncandewhurst:
Closes #928

### DIFF
--- a/docs/history/changelog.md
+++ b/docs/history/changelog.md
@@ -157,7 +157,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
   * [#1324](https://github.com/open-contracting/standard/pull/1324) `tender.datePublished`
   * [#1165](https://github.com/open-contracting/standard/pull/1165) `statusDetails` to `Tender`, `Award` and `Contract`
   * [#1125](https://github.com/open-contracting/standard/pull/1125) `Item.unit.weight`
-  * [#1326](https://github.com/open-contracting/standard/pull/1326) `links`
+  * [#1326](https://github.com/open-contracting/standard/pull/1326), [#1431](https://github.com/open-contracting/standard/pull/1431) `links`
   * [#1376](https://github.com/open-contracting/standard/pull/1376) `publisher`
   * [#1208](https://github.com/open-contracting/standard/pull/1326) The estimated and maximum values of framework agreeemnts:
     * `tender.maximumValue`. Previously, `tender.value` was used for the maximum value. However, this led to double-counting.
--- a/docs/history/changelog.md
+++ b/docs/history/changelog.md
@@ -157,7 +157,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
   * [#1324](https://github.com/open-contracting/standard/pull/1324) `tender.datePublished`
   * [#1165](https://github.com/open-contracting/standard/pull/1165) `statusDetails` to `Tender`, `Award` and `Contract`
   * [#1125](https://github.com/open-contracting/standard/pull/1125) `Item.unit.weight`
-  * [#1326](https://github.com/open-contracting/standard/pull/1326), [#1431](https://github.com/open-contracting/standard/pull/1431) `links`
+  * [#1326](https://github.com/open-contracting/standard/pull/1326) [#1431](https://github.com/open-contracting/standard/pull/1431) `links`
   * [#1376](https://github.com/open-contracting/standard/pull/1376) `publisher`
   * [#1208](https://github.com/open-contracting/standard/pull/1326) The estimated and maximum values of framework agreeemnts:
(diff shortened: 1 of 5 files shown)


mrshll1001 (approved):
This is good. I think it might be worthwhile emphasising the word "same" in italics, though.

docs/history/changelog.md
```diff
@@ -157,7 +157,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
   * [#1324](https://github.com/open-contracting/standard/pull/1324) `tender.datePublished`
   * [#1165](https://github.com/open-contracting/standard/pull/1165) `statusDetails` to `Tender`, `Award` and `Contract`
   * [#1125](https://github.com/open-contracting/standard/pull/1125) `Item.unit.weight`
-  * [#1326](https://github.com/open-contracting/standard/pull/1326) `links`
+  * [#1326](https://github.com/open-contracting/standard/pull/1326), [#1431](https://github.com/open-contracting/standard/pull/1431) `links`
```
```suggestion
  * [#1326](https://github.com/open-contracting/standard/pull/1326) [#1431](https://github.com/open-contracting/standard/pull/1431) `links`
```

docs/schema/reference.md
```diff
@@ -455,3 +455,16 @@ The publisher block is used in release and record packages to identify the sourc
...
+The `Link` object is used in the `links` field:
+
+```{field-description} ../../build/current_lang/release-schema.json /properties/links
+```
+
+```{jsonschema} ../../build/current_lang/release-schema.json
+:pointer: /definitions/Link
+```
+
+Links are used to reference related resources that describe the *same* release or contracting process. Links should not be used to reference related contracting processes. To reference related processes, see [RelatedProcess](#relatedprocess).
```
jpmckinney: Put paragraphs above jsonschema tables, so that they are easier to find and more likely to be read.

This should be the first paragraph, since it is about semantics.

Can we also add a quick example after the first sentence?

docs/schema/reference.md
```diff
@@ -455,3 +455,16 @@ The publisher block is used in release and record packages to identify the sourc
 ```{jsonschema} ../../build/current_lang/release-package-schema.json
 :include: publisher
 ```
+
+### Link
+
+The `Link` object is used in the `links` field:
```
jpmckinney:
```suggestion
The entries of the top-level `links` array are `Link` objects:
```

docs/schema/reference.md
jpmckinney:
```suggestion
Links are used to reference related resources that describe the *same* release or contracting process. Links must not be used to reference related contracting processes. To reference related processes, see [RelatedProcess](#relatedprocess).
```

jpmckinney: Actually, the description in the schema is already clear: https://github.com/open-contracting/standard/blob/1.2-dev/schema/release-schema.json#L122

Instead of this paragraph, we should just cite the schema description (as in this PR using `field-description`).

To make the point about using `relatedProcesses`, we should instead add to the schema description, e.g.:

> For references to related processes, see the `relatedProcesses` array.

duncandewhurst: Done in https://github.com/open-contracting/standard/pull/1431/commits/aa09c5c8cd09205be5485575b4dc0e6a42986872

jpmckinney (approved):

PR #1432, opened 2021-10-14
Describe use case for document.languages

### LINKED ISSUE #930: Publishing a document available in multiple languages

duncandewhurst:
Closes #930

### DIFF
--- a/docs/history/changelog.md
+++ b/docs/history/changelog.md
@@ -166,7 +166,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
     * `contracts.maximumValue`
     * `contracts.estimatedValue`
   * [#1372](https://github.com/open-contracting/standard/pull/1372) `Address.country`
-  * [#1380](https://github.com/open-contracting/standard/pull/1380) `Document.languages`
+  * [#1380](https://github.com/open-contracting/standard/pull/1380), [#1432](https://github.com/open-contracting/standard/pull/1432) `Document.languages`
 
 * Deprecate some fields:
   * [#1200](https://github.com/open-contracting/standard/pull/1200) `tender.submissionMethod`, because all codes from the `submissionMethod` codelist are deprecated.
--- a/docs/history/changelog.md
+++ b/docs/history/changelog.md
@@ -166,7 +166,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
     * `contracts.maximumValue`
     * `contracts.estimatedValue`
   * [#1372](https://github.com/open-contracting/standard/pull/1372) `Address.country`
-  * [#1380](https://github.com/open-contracting/standard/pull/1380), [#1432](https://github.com/open-contracting/standard/pull/1432) `Document.languages`
+  * [#1380](https://github.com/open-contracting/standard/pull/1380) [#1432](https://github.com/open-contracting/standard/pull/1432) `Document.languages`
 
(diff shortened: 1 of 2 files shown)


mrshll1001 (commented):
The difference between documents containing multiple languages and a document published in multiple languages may not be obvious at first read. I, a native English speaker, had to read the paragraph a few times to understand the difference.

I suggest modifying the second sentence re the "documents published in multiple languages" to be clear that we're talking about a document which has several different published versions rather than a single document containing multiple languages.

If I've misunderstood this, then I think this adds weight to the need to re-phrase this slightly.

mrshll1001 (approved):
Looks great, very clear!

docs/history/changelog.md
```diff
@@ -166,7 +166,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
     * `contracts.maximumValue`
     * `contracts.estimatedValue`
   * [#1372](https://github.com/open-contracting/standard/pull/1372) `Address.country`
-  * [#1380](https://github.com/open-contracting/standard/pull/1380) `Document.languages`
+  * [#1380](https://github.com/open-contracting/standard/pull/1380), [#1432](https://github.com/open-contracting/standard/pull/1432) `Document.languages`
```
jpmckinney: No comma between PR/issue numbers.

```suggestion
  * [#1380](https://github.com/open-contracting/standard/pull/1380) [#1432](https://github.com/open-contracting/standard/pull/1432) `Document.languages`
```

docs/schema/reference.md
```diff
@@ -69,16 +69,16 @@ The majority of OCDS data is held within a release structure. One or more releas
 
 * [package](#package-metadata)
   * [release](#release)
-    * [parties](#parties)
```
jpmckinney, line 72: Please re-do your changes from a new, clean branch, but without these whitespace changes.

docs/schema/reference.md
```diff
@@ -317,6 +317,8 @@ Documents should be published at their own stable URLs, accessible for free and
 
 OCDS allows summarizing information in the document's `description` field. Providing clear summaries is a good practice, as it allows applications to display this information in a user-interface and thus enables users to read key facts without having to search through the whole document.
 
+If a document contains multiple languages, use the `languages` field to list the languages used in the document. If multiple versions of a document are published, each in a different language, add a separate `Document` object for each version of the document.
```
jpmckinney:
```suggestion
If a document contains multiple languages, use the `languages` field to list the languages used in the document. If there are multiple versions of a document, each in a different language, add a separate `Document` object for each version of the document.
```

docs/schema/reference.md
duncandewhurst: https://github.com/open-contracting/standard/pull/1440

PR #1433, opened 2021-10-14
Add Document.relatedItem

### LINKED ISSUE #816: Add Document.relatedItem field
How frequently are documents related to a specific item?

https://github.com/openprocurement/ocds_linkedDocument_extension adds Document.relatedItem (string) and Document.documentScope (string). I am curious if other publishers need this, too.

Enquiry and Criterion from the enquiries and requirements extensions, respectively, already have a `relatedItem` field.

cc @yanakapatsila

duncandewhurst:
Closes #816

### DIFF
--- a/docs/history/changelog.md
+++ b/docs/history/changelog.md
@@ -167,11 +167,11 @@ Per the [normative and non-normative content and changes policy](https://docs.go
     * `contracts.estimatedValue`
   * [#1372](https://github.com/open-contracting/standard/pull/1372) `Address.country`
   * [#1380](https://github.com/open-contracting/standard/pull/1380) `Document.languages`
-  * [#1433](https://github.com/open-contracting/standard/pull/1433) `Document.relatedItem`
+  * [#1433](https://github.com/open-contracting/standard/pull/1433) `Document.relatedItems`
 
 * Deprecate some fields:
   * [#1200](https://github.com/open-contracting/standard/pull/1200) `tender.submissionMethod`, because all codes from the `submissionMethod` codelist are deprecated.
-  * [#1296](https://github.com/open-contracting/standard/pull/1296) `tender.eligibilityCriteria` in favor of the new `tender.exclusionGrounds` field, in order to use more common terminology and improve semantics. 
+  * [#1296](https://github.com/open-contracting/standard/pull/1296) `tender.eligibilityCriteria` in favor of the new `tender.exclusionGrounds` field, in order to use more common terminology and improve semantics.
   * [#1372](https://github.com/open-contracting/standard/pull/1372) `Address.countryName` in favor of the new `Address.country` field, to promote standardized country codes instead of non-standardized country names.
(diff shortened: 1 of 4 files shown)


duncandewhurst:
One question to resolve in https://github.com/open-contracting/standard/issues/816#issuecomment-943838042

mrshll1001 (approved):

docs/history/changelog.md
```diff
@@ -167,10 +167,11 @@ Per the [normative and non-normative content and changes policy](https://docs.go
     * `contracts.estimatedValue`
   * [#1372](https://github.com/open-contracting/standard/pull/1372) `Address.country`
   * [#1380](https://github.com/open-contracting/standard/pull/1380) `Document.languages`
+  * [#1433](https://github.com/open-contracting/standard/pull/1433) `Document.relatedItems`
 
 * Deprecate some fields:
   * [#1200](https://github.com/open-contracting/standard/pull/1200) `tender.submissionMethod`, because all codes from the `submissionMethod` codelist are deprecated.
-  * [#1296](https://github.com/open-contracting/standard/pull/1296) `tender.eligibilityCriteria` in favor of the new `tender.exclusionGrounds` field, in order to use more common terminology and improve semantics. 
+  * [#1296](https://github.com/open-contracting/standard/pull/1296) `tender.eligibilityCriteria` in favor of the new `tender.exclusionGrounds` field, in order to use more common terminology and improve semantics.
```
jpmckinney, line 174: Can you disable whatever option in your editor is changing line endings? Alternatively, change editor. Please re-do the PR from a new, clean branch, but without these whitespace changes.

schema/release-schema.json
```diff
@@ -1129,6 +1129,19 @@
             "description": "This field is deprecated in favor of `languages`, to support documents in which multiple languages are used.",
             "deprecatedVersion": "1.2"
           }
+        },
+        "relatedItems": {
+          "title": "Related items",
+          "description": "If the document relates to one or more items, the IDs of the related item(s) in the items array of the parent object. For example, for a document in `tender.documents`, the `.id` of the related item(s) in `tender.items`.",
```
jpmckinney: The "if" clause is not needed, as this is true for any field. The "(s)" construction would have to be "the ID(s) of the related item(s)", but this is unnecessarily awkward to read. The plural is fine (like the description of `documents` itself). The semantics are more properly the related items, not their identifiers – their identifiers are instead just a technical (not semantic) detail.

An important distinction to make is that this field is only relevant if the document applies to a subset of items.

```suggestion
          "description": "The subset of items to which this document relates, expressed as the `id` values of the related entries in the parent object's `items` array. For example, an entry in the `tender.documents` array may be related only to entries in the `tender.items` array.",
```

docs/history/changelog.md
duncandewhurst: Well I have no idea why Atom is doing this :-(

image

duncandewhurst: https://github.com/open-contracting/standard/pull/1439/

Issue #1297, opened 2021-05-17
Linking release reference and worked examples 


pindec:
It would be helpful for people reading the [release reference](https://standard.open-contracting.org/latest/en/schema/reference/) to link to relevant [worked examples](https://standard.open-contracting.org/latest/en/guidance/map/#mapping-the-hard-cases), just as links are given to relevant extensions.

For example, in the [parties reference](https://standard.open-contracting.org/latest/en/schema/reference/#parties), the `details` field notes that extensions can provide detailed classification of parties, and the party details extensions are linked. It would also be helpful to link to the [organization classifications](https://standard.open-contracting.org/latest/en/guidance/map/organization_classifications/) worked example.

jpmckinney:
We can create a simple Sphinx extension, where we add either Markdown frontmatter or a simple directive on the worked example, and then list the marked pages in the reference section.

Possible implementation inspiration is the example `todolist` directive: https://www.sphinx-doc.org/en/master/development/tutorials/todo.html

duncandewhurst:
At the same time we can make the changes suggested in https://github.com/open-contracting/standard/pull/1434#discussion_r759687627:

> ...we want all content except the extensions list above the jsonschema directive, as otherwise it's easy to miss.)
> 
> I think we might start using a simpler format of "seealso ... How to [do X]" when linking to guidance.

jpmckinney:
> At the same time we can make the changes suggested in [#1434 (comment)](https://github.com/open-contracting/standard/pull/1434#discussion_r759687627):
> 
> > ...we want all content except the extensions list above the jsonschema directive, as otherwise it's easy to miss.)
> > I think we might start using a simpler format of "seealso ... How to [do X]" when linking to guidance.

@yolile Can you create a new issue for this above?

yolile:
Closed by #1466 and created follow up issue #1467

PR #1434, opened 2021-10-15
Add Milestone.value

### LINKED ISSUE #438: Add Milestone.value field (related to payment schedules, planned transactions)
# Summary

## Motivation

Need to express planned transactions.

## Proposal

Use milestones:

* Add a new code to milestoneType.csv
* Add a `value` field to `Milestone`

----

We've had a request to be able to support information on planned payment schedules:

> There is a chance that altogether with the contract [ ] we'd have to
> have so called payment schedule, it is something like Plan for Transactions
> of a Contract. Since there is no Invoice in OCDS but only Transaction, I'm
> thinking of extending it with state, like Tender has. Transaction in draft
> state would represent element of payment schedule, in pending state - the
> phase when Invoice is sent, and complete state when Transaction actually
> happened already.

## Current situation

At the moment we have the transactions block which, in 1.1, contains: 

* ```id``` - ID: A unique identifier for this transaction. This identifier should be possible to cross-reference against the provided data source. For IATI this is the transaction reference.
* ```source``` - Data source: Used to point either to a corresponding Fiscal Data Package, IATI file, or machine or human-readable source where users can find further information on the budget line item identifiers, or project identifiers, provided here.	
* ```date``` -  Date: The date of the transaction
* ```value```  - Value: The value of the transaction. See Value Object
* ```payer``` - Payer: An organization reference for the organization from which the funds in this transaction originate. See OrganizationReference Object
* ```payee``` - Payee: An organization reference for the organization which receives the funds in this transaction. See OrganizationReference Object
* ```uri``` -  Linked spending information: A URI pointing directly to a machine-readable record about this spending transaction.

We note that this is modelled on the [IATI transaction model](http://iatistandard.org/202/activity-standard/iati-activities/iati-activity/transaction/) although we did not use the '[transaction-type](http://iatistandard.org/202/codelists/TransactionType/)' concept from IATI to allow a distinction between 'commitments', 'disbursements', 'expenditure' and so-on (similar to the idea of a transaction.status field).

## Considerations

One of the advantages of our current model is that a user can simply sum up all the transactions to get total spend on a contract (something that gets quite complicated when there are status fields involved to be aware of). Our documentation states that transactions *"can be used to represent actual flows of money between organisations in relation to this contract."* suggesting that if we changed it to also include non-actual (i.e. planned) flows, this would not be in conformance with current use.

There has also been some work on an extension to [relate transactions and milestones](https://github.com/open-contracting/ocds_transactions_relatedMilestone_extension), such that milestones could be used to represent invoice deadlines. 

Using the proposed [budget breakdown](https://github.com/open-contracting/ocds_budget_breakdown_extension) extension it is also possible to model small payment periods that could be matched to invoice deadlines, and that allow for different sources of money to be specified. 

## Options

In response to the initial inquiry, I would suggest:

* First, looking to see whether using milestones to capture invoice deadlines could suffice, particularly with the [milestoneType](http://standard.open-contracting.org/1.1-dev/en/schema/codelists/#milestone-type) codelist introduced in 1.1 (and as an open codelist, so it could have invoice milestones added to it);

* Second, if this does not work, looking at creating an alternative property pointing to an array of Transaction objects - potentially ```paymentSchedule```. 

For really detailed post-award invoicing information, I wonder to what extend [CEN-BII can provide models](http://www.cenbii.eu/cwa-17029-bii-post-award/) to link out to.

duncandewhurst:
Closes #438


docs/schema/reference.md
```diff
@@ -402,6 +404,7 @@ Milestone information can be included in the [planning](#planning), [tender](#te
 Notes:
 
 * The `dateModified` field should be changed whenever the progress towards a milestone is reviewed, and the `status` either updated, or re-confirmed. 
+* To represent a planned payment, set `.type` to 'payment' and set `.value` to the payment's value. Once the milestone is met, add a [Transaction](#transaction) to `contracts/implementation/transactions` and set its `.id` and `.value`. For implementation milestones, the transaction can refer back to the milestone using the [transaction related milestones extension](https://extensions.open-contracting.org/en/extensions/transaction_milestones/master/).
```
yolile: I'm not sure if it is better to put that here as a note or if it is better to update the existing worked example, eg: https://standard.open-contracting.org/latest/en/guidance/map/milestones/#delivery-and-financing-data, or maybe do both things.

jpmckinney: I think best to update the worked example, and link to it from here using a `seealso` directive. This avoids repetition, and also avoids a long example in the reference.

duncandewhurst: Moved in https://github.com/open-contracting/standard/pull/1434/commits/db4c746d9e77eb2f7566ee06f6772a2c7b1f6c58

yolile (approved):

docs/guidance/map/milestones.md
```diff
@@ -33,6 +33,8 @@ At the point of contract signature, a comparison between `tender/milestones` and
 
 The `dueDate`, `dateMet`, `dateModified` and [`status`](../../schema/codelists.md#milestone-status) fields are used to track the lifecycle of the milestone.
 
+To represent a planned payment, add a `Milestone`, set its `.type` to 'payment' and set its `.value` to the payment's value. Once the milestone's `.status` is met, add a [Transaction](../../schema/reference.md#transaction) to `contracts/implementation/transactions` and set its `.id` and `.value`. For implementation milestones, the transaction can refer back to the milestone using the [transaction related milestones extension](https://extensions.open-contracting.org/en/extensions/transaction_milestones/master/).
```
jpmckinney: We don't discuss `status` previously, so it's a bit odd to use it in that sentence.

I assume it's understood that the milestone `id`, `status`, `dueDate`, etc. are set and updated as usual. Similarly, we don't need to say which fields to set on the transaction, as it's just the usual.

Alternatively, we can make a bullet list with the full, explicit instructions.

```suggestion
To represent a planned payment, add a `Milestone`, set its `.type` to 'payment' and set its `.value` to the payment's value. Once the milestone is met, add a [Transaction](../../schema/reference.md#transaction) to `contracts/implementation/transactions`. For implementation milestones, the transaction can refer back to the milestone using the [transaction-related milestones extension](https://extensions.open-contracting.org/en/extensions/transaction_milestones/master/).
```

docs/guidance/map/milestones.md
```diff
@@ -132,7 +134,7 @@ In the first implementation update release:
 
 In the second implementation update release:
 
-* The construction company has received payment for the work done so far, so the milestone for the wall restoration with type 'financing' is updated. A new `transaction` is disclosed, with the amount paid to the company.
+* The construction company has received payment for the work done so far, so the milestone for the wall restoration with type 'payment' is updated. A new `transaction` is disclosed, with the amount paid to the company. The [transaction related milestones extension](https://extensions.open-contracting.org/en/extensions/transaction_milestones/master/) is used to link the transaction to the milestone.
```
jpmckinney:
```suggestion
* The construction company has received payment for the work done so far, so the milestone for the wall restoration with type 'payment' is updated. A new `transaction` is disclosed, with the amount paid to the company. The [transaction-related milestones extension](https://extensions.open-contracting.org/en/extensions/transaction_milestones/master/) is used to link the transaction to the milestone.
```

docs/schema/reference.md
```diff
@@ -226,6 +226,8 @@ Information on subcontracts is not currently included in the core OCDS schema, b
 
 The transaction block is modelled on the [International Aid Transparency Initiative (IATI) transaction element](https://iatistandard.org/en/iati-standard/203/activity-standard/iati-activities/iati-activity/transaction/), and can be used to represent actual flows of money between organizations in relation to this contract. As with the [budget](#budget) block, this can be used to cross-reference to a third party `source` of data, and ought to re-use identifiers from that source.
 
+To represent planned payments, use [Milestones](#milestones).
```
jpmckinney: Move this after the "In most circumstances" paragraph, as this paragraph leads the reader away from Transactions, whereas "In most circumstances" is still about transactions. We might prefer to put this in a `{note}` since it's not directly about transactions.

Also, let's use this PR to move the jsonschema directive below the paragraphs, just above the extensionlist.
````suggestion
```{note}
To represent planned payments, use [Milestones](#milestones) instead.
```
````

File: docs/schema/reference.md [hunk 5]
```diff
@@ -403,6 +405,12 @@ Notes:
 
 * The `dateModified` field should be changed whenever the progress towards a milestone is reviewed, and the `status` either updated, or re-confirmed. 
 
+```{admonition} Planned payments
+:class: seealso
+
+To learn about how to represent planned payments, see the [Delivery and payment data worked example](../guidance/map/milestones.md#delivery-and-payment-data)
+```
+
```
jpmckinney: Move this and the notes above the jsonschema table. Remove the existing "Notes:" and change the bullet to a simple paragraph. (This follows from an another issue, where we want all content except the extensions list above the jsonschema directive, as otherwise it's easy to miss.)

I think we might start using a simpler format of "seealso ... How to [do X]" when linking to guidance.

```suggestion
```{seealso}
[How to represent planned payments](../guidance/map/milestones.md#delivery-and-payment-data)
```
```

File: schema/release-schema.json [hunk 6]
```diff
@@ -1006,6 +1006,11 @@
             "$ref": "#/definitions/Document"
           },
           "uniqueItems": true
+        },
+        "value": {
+          "title": "Value",
+          "description": "If the milestone represents a planned payment, the payment's value.",
```
jpmckinney: Start with the semantics, then add qualifiers.

```suggestion
          "description": "The payment's value, if the milestone represents a planned payment.",
```

docs/schema/reference.md
duncandewhurst: Changes made in https://github.com/open-contracting/standard/pull/1434/commits/d639e15b1495ffe05d50c0ef71725dfa572b6428

Both suggestions sound good. Shall I open an issue to go through `reference.md` and make those changes?

jpmckinney: I think they can be combined with https://github.com/open-contracting/standard/issues/1297

jpmckinney (approved):

Issue #848, opened 2019-03-29
Release date must be unique within a contracting process


duncandewhurst:
The merging rules begin with:

> Order all the releases that share an `ocid` by their release `date`. Starting from the oldest release (old), compare it with the next oldest release (new), and apply the following rules...

But I don't think we state anywhere that two releases under the same `ocid` must not share the same `date` to make this ordering possible.

I suggest adding this uniqueness requirement to the description of `date` in the schema and implementing it as an additional check in the data review tool.

jpmckinney:
Would 'must' be backwards-compatible, or do we need to relax to 'should' for OCDS 1.2?

Copying that this came up where a publisher "was updating the release date for all releases relating to a contracting process whenever anything changed."

Regarding the latter, that behavior is already blocked by the rule that releases are immutable and must not be changed (without a change in release `id`).

JachymHercher:
> Would 'must' be backwards-compatible, or do we need to relax to 'should' for OCDS 1.2?

Philosophically, I'd say that if you want, you have some discretion in this. Since the definition of `releaseDate` is "The date on which the information contained in the release was first recorded in, or published by, any system.", I think you could argue that the "first" implies there is only one, because things never happen in precisely the same moment. 

More pragmatically, you could check a reasonable sample of OCDS-compliant data and see whether this type of things occurs. If it does, then we probably shouldn't treat it as backwards compatible, because adding such a condition into 1.2 would mean that some old releases would stop be OCDS compliant because of a version change.

ColinMaudry:
> The date on which the information contained in the release was first recorded in

That rule also applies to cases where the releases have the same `ocid`, same `date` but different content.

> things never happen in precisely the same moment.

They do... in the world of software. For instance two releases could be generated at the same time, each describing part of a `tender`. It's not good programming, but it's possible.

ColinMaudry:
Unless data shows that it doesn't happen, I would proceed with the change without breaking compatibility ("should").

duncandewhurst:
I'm happy with a 'should' in the schema as long as we can include an additional check in the DRT to catch the type of error that prompted the issue.

ColinMaudry:
I suggest the following description for `date`, using part of the description of `id`:

> The date on which the information contained in the release was first recorded in, or published by, any system. The release date should be unique within the scope of the contracting process identified by the Open Contracting ID (ocid), for a given version of OCDS.

jpmckinney:
[...]

PR #1443, opened 2021-11-11
schema: Include release date uniqueness definition in its description

yolile:
ref #848

### DIFF
--- a/schema/release-schema.json
+++ b/schema/release-schema.json
@@ -20,7 +20,7 @@
     },
     "date": {
       "title": "Release Date",
-      "description": "The date on which the information contained in the release was first recorded in, or published by, any system. The release date should be unique within the scope of the contracting process identified by the Open Contracting ID (ocid), for a given version of OCDS. For a compiled release, the maximum `date` among the individual releases used to create the compiled release.",
+      "description": "The date on which the information contained in the release was first recorded in, or published by, any system. For a compiled release, the maximum `date` among the individual releases used to create the compiled release. The release date should be unique within the scope of the contracting process identified by the Open Contracting ID (ocid), for a given version of OCDS.",
       "type": "string",
       "format": "date-time",
       "omitWhenMerged": true
--- a/schema/release-schema.json
+++ b/schema/release-schema.json
@@ -20,7 +20,7 @@
     },
     "date": {
       "title": "Release Date",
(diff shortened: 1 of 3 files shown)


schema/release-schema.json
```diff
@@ -20,7 +20,7 @@
     },
     "date": {
       "title": "Release Date",
-      "description": "The date on which the information contained in the release was first recorded in, or published by, any system. For a compiled release, the maximum `date` among the individual releases used to create the compiled release.",
```
jpmckinney, line 23: The part about the compiled release is not a full sentence - it's the alternative definition in case of a compiled release. So, I think it has to appear before the full sentence about uniqueness.

```suggestion
      "description": "The date on which the information contained in the release was first recorded in, or published by, any system. For a compiled release, the maximum `date` among the individual releases used to create the compiled release. The release date should be unique within the scope of the contracting process identified by the Open Contracting ID (ocid), for a given version of OCDS.",
```

schema/release-schema.json
```diff
@@ -20,7 +20,7 @@
...
+      "description": "The date on which the information contained in the release was first recorded in, or published by, any system. The release date should be unique within the scope of the contracting process identified by the Open Contracting ID (ocid), for a given version of OCDS. For a compiled release, the maximum `date` among the individual releases used to create the compiled release.",
```
jpmckinney: The part about the compiled release is not a full sentence - it's the alternative definition in case of a compiled release. So, I think it has to appear before the full sentence about uniqueness.

```suggestion
      "description": "The date on which the information contained in the release was first recorded in, or published by, any system. For a compiled release, the maximum `date` among the individual releases used to create the compiled release. The release date should be unique within the scope of the contracting process identified by the Open Contracting ID (ocid), for a given version of OCDS.",
```

jpmckinney:
Also, does this close #848 or only ref it?

yolile:
> Also, does this close #848 or only ref it?

I'm not sure if we should close that issue or change it to OCDS 2.0 after we merge this one, as for OCDS 2.0 we may want to change the definition again from "should" to "must".

jpmckinney:
OCDS 2.0 has a lot of rules that might need to change, but we haven't started OCDS 2.0, so we don't need issues for those types of things yet.

PR #1450, opened 2021-11-26
Replace repeated examples in schema/merging with a link to reference/build/merging

### LINKED ISSUE #1402: schema/merging and reference/build/merging: repeated example, joining
The worked example in https://standard.open-contracting.org/latest/en/schema/merging/ is also in https://standard.open-contracting.org/latest/en/guidance/build/merging/#example-1-updates. My reflex is that we do not want to have the same information in multiple places of the standard, as it makes things unnecessary long and might feel confusing.

Shouldn't we remove this example from one of the webpages or, since the guidance page contains little besides the three worked examples, perhaps merge schema/merging into guidance/build/merging?

JachymHercher:
closes #1402


jpmckinney:
Ha, the test is failing because we specifically used the merging page to test some JavaScript. I'll change the test.

PR #974, opened 2020-01-22
Awards-related examples

duncandewhurst:
@jpmckinney tests are failing on `consortia_simple.csv` with the following errors, however I've checked the file using csvkit's `csvclean` command and it reports no errors, any ideas?

```
=================================== FAILURES ===================================

/tmp/test_csv.py:121: AssertionError: One or more codelist CSV files are invalid. See warnings below.

=============================== warnings summary ===============================

test_csv.py::test_valid

  ERROR: docs/examples/consortia_simple.csv has 5 not 8 columns in row 2

test_csv.py::test_valid

  ERROR: docs/examples/consortia_simple.csv is improperly formatted (e.g. missing trailing newline, extra quoting characters, non-"\n" line terminator):

  'awards/0/id,awards/0/title,awards/0/suppliers/0/name,awards/0/suppliers/0/identifier/scheme,awards/0/suppliers/0/identifier/scheme,awards/0/suppliers/1/name,awards/0/suppliers/0/identifier/scheme,awards/0/suppliers/0/identifier/scheme\nDHSC-2019-A18074,Medical imaging device development,SIEMENS PUBLIC LIMITED COMPANY,GB-COH,727817,MICROSOFT LIMITED,GB-COH,1624297\n'

  'awards/0/id,awards/0/title,awards/0/suppliers/0/name,awards/0/suppliers/0/identifier/scheme,awards/0/suppliers/0/identifier/scheme,awards/0/suppliers/1/name,awards/0/suppliers/0/identifier/scheme,awards/0/suppliers/0/identifier/scheme\nDHSC-2019-A18074,Medical imaging device development,SIEMENS PUBLIC LIMITED COMPANY,1624297,1624297,MICROSOFT LIMITED,1624297,1624297\n'
```

PR #1451, opened 2021-11-29
Clarify when to fill awards and when contracts

### LINKED ISSUE #1400: Guidance on awards and contracts: when to fill in one and/or the other
The section https://standard.open-contracting.org/latest/en/guidance/map/awards_contracts/#id1 starts with a general introduction and then, in the last three paragraphs, covers three different scenarios:

> If the data in the source system relates only to contract awards, then only the `award` section of OCDS ought to be populated, unless the law governing procurement permits no changes between award and signature of a contract, in which case the `contract` section can be populated.

> If the data in the source system relates only to the signed contract and no changes are permitted between the award and signature of a contract, then: 
> * Add an `Award` object to `awards` and a `Contract` object to `contracts`
> * Set `awards.id` and `contracts.id` to the same value
> * Populate the `awards.value`, `awards.contractPeriod` and `awards.items` fields with the initial contract value, period and items respectively
> * If the contract value, period or items are subsequently updated, populate the `contracts.value`, `contracts.period` and `contracts.items` fields with the updated value

> The above approach is also applicable to direct call-offs from framework agreements with multiple suppliers, since the call-off is, conceptually, an award.

Questions: 
1) In the middle paragraph, if they have data just about the signed contract, why should the users populate `awards` and not `contracts`? My reflex would be that if it's about a contract, then put it in `contracts` (as far as you can) (or both `awards` and `contracts`).  
2) In the last paragraph, why would all contracts in a direct call-off be treated the same way as "systems with signed contracts are no changes permitted between the award and the signature of a contract"? My reflex would be to treat an award and a contract within a call-off pretty much the same as any other award and contract. In the EU, there is a difference that contracts within framework agreements are not subject to a (mandatory) standstill period, but that might be different elsewhere; there can be a voluntary standstill period; and the awards and contracts could still be different as a result of amendments. 

    If awards and contracts are the same within call-offs as normally, then I would remove this paragraph and perhaps add a mention of this scenario in the "normal case", see point 4). 

3) The same issue as above is also replicated in https://standard.open-contracting.org/latest/en/guidance/map/related_processes/#add-an-award-contract-and-supplier, as briefly discussed in https://github.com/open-contracting/standard/issues/1301#issuecomment-891176354. 

4) There are no instructions for a scenario "If the data in the source system relates to both the signed contract and the award". 

    I assume this is because this is the "default" situation and thus does not require any treatment in the guidance, it should be obvious from the descriptions of the fields. Is that correct? It might be worth an intro sentence anyway.

5) There are no instructions for a scenario "If the data in the source system relates only to the signed contract and ~~no~~ changes are permitted between the award and signature of a contract

    Isn't this an omission? My guess is that such systems do exist (e.g. "contract registers" in the stricts sense of systems that just contain digitized contracts might not have any information about awards; eForms also don't really worry about awards, but rather about contracts). Furthermore, there are non-obvious instructions to give - along the lines of "i) even though you're just taking information from your contract, in OCDS, you'll need to put some information in `awards` (e.g. `suppliers`;  ii) (?) for fields that are both in `awards` and `contracts` (e.g. `value`, `items`), leave the ones in `awards` empty.

JachymHercher:
closes #1400

### DIFF
--- a/docs/guidance/map/awards_contracts.md
+++ b/docs/guidance/map/awards_contracts.md
@@ -82,7 +82,7 @@ In OCDS, **awards** and **contracts** are modelled as separate stages of the con
 
 Each contracting process can have many awards and each award can have many related contracts.
 
-OCDS separates data about the contract award and data about the signed contract into the `awards` and `contracts` sections respectively. `Award` objects (within `awards`) and `Contract` objects (within `contracts`) are connected by setting `awards.id` and `contracts.id` to the same value. Source systems can contain data on awards, on contracts, or on both.
+OCDS separates data about the contract award and data about the signed contract into the `awards` and `contracts` sections respectively. `Award` objects (within `awards`) and `Contract` objects (within `contracts`) are connected by setting `awards.id` and `contracts.awardID` to the same value. Source systems can contain data on awards, on contracts, or on both.
 
 If the data in the source systems relates: 
 * to both awards and contracts, then both `awards` and `contracts` ought to be populated. 
--- a/docs/guidance/map/awards_contracts.md
+++ b/docs/guidance/map/awards_contracts.md
@@ -87,7 +87,7 @@ OCDS separates data about the contract award and data about the signed contract 
 If the data in the source systems relates: 
 * to both awards and contracts, then both `awards` and `contracts` ought to be populated. 
 * only to contract awards, then only `awards` ought to be populated.
-* only to the contract, then `contracts`, `awards.id`, and `awards.suppliers` ought to be populated.  
+* only to contracts, then `contracts`, `awards.id`, and `awards.suppliers` ought to be populated.  
 
 If the contract (e.g. its value, period or items) is subsequently updated or amended, update only the corresponding fields in `contracts`. The fields in `awards` stay the same. 
 
--- a/docs/guidance/map/awards_contracts.md
+++ b/docs/guidance/map/awards_contracts.md
@@ -84,10 +84,10 @@ Each contracting process can have many awards and each award can have many relat
 
 OCDS separates data about the contract award and data about the signed contract into the `awards` and `contracts` sections respectively. `Award` objects (within `awards`) and `Contract` objects (within `contracts`) are connected by setting `awards.id` and `contracts.awardID` to the same value. Source systems can contain data on awards, on contracts, or on both.
 
-If the data in the source systems relates: 
-* to both awards and contracts, then both `awards` and `contracts` ought to be populated. 
-* only to contract awards, then only `awards` ought to be populated.
-* only to contracts, then `contracts`, `awards.id`, and `awards.suppliers` ought to be populated.  
+If the data in the source systems relates to: 
+* both awards and contracts, then both `awards` and `contracts` ought to be populated. 
+* only awards, then only `awards` ought to be populated.
+* only contracts, then `contracts`, `awards.id`, and `awards.suppliers` ought to be populated.  
 
 If the contract (e.g. its value, period or items) is subsequently updated or amended, update only the corresponding fields in `contracts`. The fields in `awards` stay the same. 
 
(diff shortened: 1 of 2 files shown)


docs/guidance/map/awards_contracts.md
```diff
@@ -82,18 +82,14 @@ In OCDS, **awards** and **contracts** are modelled as separate stages of the con
 
 Each contracting process can have many awards and each award can have many related contracts.
 
-OCDS separates data about the contract award and data about the signed contract into the `awards` and `contracts` sections respectively. Source systems can contain data on awards, on contracts, or on both.
+OCDS separates data about the contract award and data about the signed contract into the `awards` and `contracts` sections respectively. `Award` objects (within `awards`) and `Contract` objects (within `contracts`) are connected by setting `awards.id` and `contracts.id` to the same value. Source systems can contain data on awards, on contracts, or on both.
```
yolile:
```suggestion
OCDS separates data about the contract award and data about the signed contract into the `awards` and `contracts` sections respectively. `Award` objects (within `awards`) and `Contract` objects (within `contracts`) are connected by setting `awards.id` and `contracts.awardID` to the same value. Source systems can contain data on awards, on contracts, or on both.
```

JachymHercher: Thank you!

docs/guidance/map/awards_contracts.md
```diff
@@ -82,18 +82,14 @@ In OCDS, **awards** and **contracts** are modelled as separate stages of the con
...
 
-OCDS separates data about the contract award and data about the signed contract into the `awards` and `contracts` sections respectively. Source systems can contain data on awards, on contracts, or on both.
+OCDS separates data about the contract award and data about the signed contract into the `awards` and `contracts` sections respectively. `Award` objects (within `awards`) and `Contract` objects (within `contracts`) are connected by setting `awards.id` and `contracts.awardID` to the same value. Source systems can contain data on awards, on contracts, or on both.
 
-If the data in the source system relates only to contract awards, then only the `award` section of OCDS ought to be populated, unless the law governing procurement permits no changes between award and signature of a contract, in which case the `contract` section can be populated.
+If the data in the source systems relates: 
+* to both awards and contracts, then both `awards` and `contracts` ought to be populated. 
+* only to contract awards, then only `awards` ought to be populated.
+* only to contracts, then `contracts`, `awards.id`, and `awards.suppliers` ought to be populated.
```
jpmckinney:
```suggestion
If the data in the source systems relates to: 
* both awards and contracts, then both `awards` and `contracts` ought to be populated. 
* only awards, then only `awards` ought to be populated.
* only contracts, then `contracts`, `awards.id`, and `awards.suppliers` ought to be populated.  
```

jpmckinney:
```suggestion
If the data in the source systems relates to: 
* both awards and contracts, then `awards` and `contracts` ought to be populated. 
* only awards, then `awards` ought to be populated.
* only contracts, then `contracts`, `awards.id`, and `awards.suppliers` ought to be populated.  
```

docs/history/changelog.md
```diff
@@ -36,6 +36,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
     * [#1016](https://github.com/open-contracting/standard/pull/1016) [#1140](https://github.com/open-contracting/standard/pull/1140) Add links to OCDS profiles and linked standards.
     * [#1193](https://github.com/open-contracting/standard/pull/1193) Removed language distinguishing "core", "community", and "local" extensions.
     * [#1428](https://github.com/open-contracting/standard/pull/1428) Integrate content from [what to do with fields that don't map](https://www.open-contracting.org/2018/01/30/fields-dont-map-first/).
+  * In Awards and Contracts, clarify when to fill `awards` and when `contracts`, simplify. [#1451](https://github.com/open-contracting/standard/pull/1451).
```
jpmckinney: Just append the PR number to `* Awards and contracts [#974](https://github.com/open-contracting/standard/pull/974)` above.

JachymHercher: Done

PR #1454, opened 2021-12-02
Rename and clarify Guidance-Map-Amendments

### LINKED ISSUE #1446: Rename and clarify Guidance/Map/Amendments
Looking at https://standard.open-contracting.org/1.1/en/guidance/map/amendments/, I have two suggestions:

1) Shouldn't it be renamed to Updates and amendments, since it deals with both and these are the terms used in `releaseTag`? 
2) When publishing updates, different eProcurement systems have different approaches to what should be published when amending information: (i) just the amendment (i.e. just the new information) or (ii) unamended information and the amendment (i.e. everything that is still valid after the amendment). According to point (2) in https://github.com/open-contracting/standard/issues/1160#issuecomment-970610262, both approaches are acceptable in OCDS. Shouldn't we state this explicitly in the guidance? 

    Currently, the closest we get to this is the second bullet of https://standard.open-contracting.org/1.1/en/guidance/map/amendments/#record, but it is within a worked example and refers only to the approach for amendments, not updates. The update example itself follows approach (ii).

JachymHercher:
Closes #1446

### DIFF
--- a/docs/guidance/map/amendments.md
+++ b/docs/guidance/map/amendments.md
@@ -2,7 +2,7 @@
 
 Information about a contracting process will often change over time.
 
-Each time information changes, a new OCDS release ought to be published. The new release can either contain only newly published information or all the information valid at the time of publication (i.e. old information that is still valid and new information).  
+Each time information changes, a new OCDS release ought to be published. The new release can contain information that was previously published, in addition to the new information.
 
 There are three kinds of changes:
 
--- a/docs/guidance/map/amendments.md
+++ b/docs/guidance/map/amendments.md
@@ -1,8 +1,8 @@
-# Amendments
+# Updates and amendments
 
 Information about a contracting process will often change over time.
 
-Each time information changes, a new OCDS release ought to be published.
+Each time information changes, a new OCDS release ought to be published. The new release can either contain only newly published information or all the information valid at the time of publication (i.e. old information that is still valid and new information).  
 
 There are three kinds of changes:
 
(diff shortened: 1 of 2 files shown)


docs/history/changelog.md
```diff
@@ -36,6 +36,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
     * [#1016](https://github.com/open-contracting/standard/pull/1016) [#1140](https://github.com/open-contracting/standard/pull/1140) Add links to OCDS profiles and linked standards.
     * [#1193](https://github.com/open-contracting/standard/pull/1193) Removed language distinguishing "core", "community", and "local" extensions.
     * [#1428](https://github.com/open-contracting/standard/pull/1428) Integrate content from [what to do with fields that don't map](https://www.open-contracting.org/2018/01/30/fields-dont-map-first/).
+    * [#1454](https://github.com/open-contracting/standard/pull/1454) Rename and clarify Guidance/Map/Amendments.
```
jpmckinney: Add PR number to the `Amendments [#950](https://github.com/open-contracting/standard/pull/950)` bullet, which you can rename to "Updates and amendments"

docs/guidance/map/amendments.md
```diff
@@ -1,8 +1,8 @@
-# Amendments
+# Updates and amendments
 
 Information about a contracting process will often change over time.
 
-Each time information changes, a new OCDS release ought to be published.
+Each time information changes, a new OCDS release ought to be published. The new release can either contain only newly published information or all the information valid at the time of publication (i.e. old information that is still valid and new information).
```
jpmckinney: I think this is simpler?

```suggestion
Each time information changes, a new OCDS release ought to be published. The new release can contain information that was previously published, in addition to the new information.

JachymHercher: Yep, definitely better, thanks.

jpmckinney (approved):

PR #1461, opened 2021-12-21
Update framework agreements guidance (`procurementMethod` etc.)

### LINKED ISSUE #1301: Improvements/fixes to the framework guidance

JachymHercher:
Closes #1301


docs/guidance/map/framework_agreements.md
```diff
@@ -36,13 +36,13 @@ Framework agreements can be **open** or **closed**:
 
 Procurement systems which provide for closed framework agreements include the EU (Framework Agreements, per Directive 2014/24/EU, Article 33) and the US federal system (Indefinite-delivery contracts, per Federal Acquisition Regulation, Subpart 16.5). An example of a procurement system which provides for open framework agreements is the EU (Dynamic Purchasing Systems, per Directive 2014/24/EU, Article 33).
 
-Framework agreement procedures can involve second-stage competition, or not:
+Framework agreement procedures can involve second stage competition, or not:
```
jpmckinney: Please restore the hyphen. It's how UNCITRAL styles it, and it is easier to parse.

docs/guidance/map/framework_agreements.md
```diff
@@ -85,44 +85,45 @@ The following guidance describes how to model the different stages of a framewor
 
 ### Invitation to participate in the first stage of a framework agreement procedure
 
-* Create a release and add 'tender' to the `tag` array
+* Create a release and add 'tender' to the `tag` array.
```
jpmckinney, line 92: Please make style-only changes in a separate commit, so that I can review each commit individually and focus my effort on the "substantive" commit(s). Right now, this makes the PR much longer to review because instead of maybe 5 lines of relevant diffs I have dozens of lines of styles changes to skip over.

docs/guidance/map/framework_agreements.md
```diff
@@ -85,44 +85,45 @@ The following guidance describes how to model the different stages of a framewor
...
 * For each buyer that will participate in the framework agreement, add an `Organization` object to the `parties` array, add 'buyer' to its `.roles`, and populate its other fields.
   * If all potential buyers can participate, don't add them to the `parties` array. Instead, add a note in the [publication policy](../publish) to explain how users should interpret a contracting process with no buyers.
   * If a single buyer will participate in the framework agreement, set the `buyer.id` and `buyer.name` to match the buyer's object in the `parties` array. Otherwise, don't populate `buyer`.
 * In the `tender` section, set:
+  * `tender.procurementMethod` in line with the competitive conditions of the first stage of the framework agreement. Since framework agreements usually involve qualification, this is most often 'selective'.
```
jpmckinney:
```suggestion
  * `tender.procurementMethod` according to the competitive conditions of the first stage of the framework agreement. Since framework agreements typically involve qualification, this is most often 'selective'.
```

docs/guidance/map/framework_agreements.md
```diff
@@ -142,19 +143,19 @@ The following guidance describes how to model the different stages of a framewor
 #### Add an award, contract and supplier
 
 * Add an `Award` object to the `awards` array and a `Contract` object to the `contracts` array.
-* Set the award's `.id` and the contract's `.id` to the same value.
+* Set the award's `.id` and the contract's `.awardID` to the same value.
```
jpmckinney:
```suggestion
* Set the award's `.id` and the contract's `.awardID` to the same value, and set the contract's `.id`.
```

docs/guidance/map/framework_agreements.md
JachymHercher: Yes, sorry, I realized that as soon as I made it in the same one.

docs/guidance/map/framework_agreements.md
JachymHercher: Restored.

docs/guidance/map/framework_agreements.md
```diff
@@ -81,23 +81,24 @@ The two extensions should be declared in the package metadata:
 :language: json
 ```
 
-The following guidance describes how to model the different stages of a framework agreement procedure in OCDS.  
+The following guidance describes how to model the different stages of a framework agreement procedure in OCDS. (The guidance does cover modelling that is the same as for other types of contracts, for example populating `contracts`).
```
jpmckinney: I think the word "not" is missing after "does"? Also, we use the American spelling, "modeling". Also, to avoid confusion, do you mean other types of "procedures"?

JachymHercher: Yes, missing, sorry, added.

Modeling - oops, ok, sorry, fixed. 

"Other types of procedures" - not really. Frameworks are a particular type of contract, so "other types of contracts" was the intended formulation. To me it doesn't seem confusing, but I understand that to other it might. If you think that is the case, we can think of a formulation that avoids mentioning contracts? 

For reference, here is the definition of contract 

> "Information regarding the contract, typically between the buyer and supplier. This includes contracts describing all the contractual conditions (e.g. item, quantity, price, payment terms, time and place of delivery), as well as contracts only describing the general contractual conditions (such as a framework agreement) and those only describing the specific contractual conditions (such as a contract within a framework agreement). Communication between contractual parties that consists of minor specifications of conditions agreed previously (e.g. specifying the time or place of delivery) is not considered a contract. Amendments are considered as part of the contract that is being amended. Contracts are also used when giving prizes or other rewards (e.g. a follow-up contract) resulting from a design contest."

jpmckinney: Aha, maybe then this?

> This guidance doesn't cover the modeling of the contract (via the `contracts` array), because this information about a framework agreement is modeled the same as for other types of contracts.

JachymHercher: Sounds good! Done in https://github.com/open-contracting/standard/pull/1461/commits/bbf9bf21b19b685315c5e6b21023835a994d1d70.

jpmckinney (commented):
I figure we should remove the `### Updating the contract value, period or items` section as well?

If not, we'll need something like:

```md
(If no `Contract` object in the `contracts` array has an `awardID` matching the award's `id`, add one.)
```

JachymHercher:
Indeed, let's remove it, sorry. Done in https://github.com/open-contracting/standard/pull/1461/commits/e66f5fdbe18c6910bcc157395484b6ceea5b3cca.

I have also pushed two other commits:
- The framework guidance uses the expression "initial contract value", which we use nowhere else. I have replaced it with "award value". ("Initial contract value" also seems confusing, because I would expect it to be the value in `contract` before any amendments or updates, not the one in `award`.)
- Replaced an instance of "contract" being used where it should have actually been "award".

docs/guidance/map/framework_agreements.md
```diff
@@ -106,46 +107,43 @@ The following guidance describes how to model the different stages of a framewor
...
-* If the initial contract value, period or items are subsequently updated, populate the contract's `.value`, `.period` and `.items` with the updated values.
+* Create a release with the **same** `ocid` as the tender release *for the invitation to participate in a second-stage competition*, and add 'award' to the `tag`.
+* [Add an award and supplier](#add-an-award-and-supplier).
 
 ### Common operations
 
 #### Relate the second stage to the first stage
 
 * Add a `RelatedProcess` object to the `relatedProcesses` array, set its `.id` (to '1', for example), add 'framework' to its `.relationship` array, set its `.scheme` to 'ocid' and set its `.identifier` to the `ocid` of the invitation to participate in the framework agreement.
+* Note: second-stage contracting processes ought to be always related to the first-stage contracting process, never to other second-stage contracting processes.
```
jpmckinney: I think this second note is not needed, as the preceding sentence is explicitly about relating to "the invitation to participate in the framework agreement".

```suggestion
```

docs/guidance/map/framework_agreements.md
```diff
@@ -63,7 +63,7 @@ In some cases, complex contracting processes cannot be represented under a singl
...
+OCDS models the first and second stages of framework agreement procedures as separate contracting processes, linked together using the `relatedProcesses` array. The `tender.techniques.hasFrameworkAgreement` field, from the [Techniques](https://extensions.open-contracting.org/en/extensions/techniques/master/) extension, is used to identify contracting processes that represent the first stage of a framework agreement procedure. The presence of a related first-stage process with a `.relationship` set to 'framework' is used to identify contracting processes that represent the second stage of a framework agreement procedure. Second-stage contracting processes ought to be always related to the first-stage contracting process, never to other second-stage contracting processes.
```
jpmckinney: I think it's important to preserve the word "related process" and not "related first-stage process" because "related process" is more obviously connected to `relatedProcesses`.

````suggestion
OCDS models the first and second stages of framework agreement procedures as separate contracting processes, linked together using the `relatedProcesses` array. The `tender.techniques.hasFrameworkAgreement` field, from the [Techniques](https://extensions.open-contracting.org/en/extensions/techniques/master/) extension, indicates whether the contracting process represents the first stage of a framework agreement procedure. The presence of a related process with a `.relationship` field set to 'framework' indicates whether the contracting process represents the second stage of a framework agreement procedure.

```{note}
The 'framework' code is only used to relate a second-stage contracting process to its first-stage contracting process, never to other second-stage contracting processes.
```
````

File: docs/guidance/map/framework_agreements.md [hunk 8]
```diff
@@ -81,23 +81,24 @@ The two extensions should be declared in the package metadata:
 :language: json
 ```
 
-The following guidance describes how to model the different stages of a framework agreement procedure in OCDS.  
+The following guidance describes how to model the different stages of a framework agreement procedure in OCDS. (The guidance does not cover the modeling of the contract (using the `contracts` array), because this information about a framework agreement is modeled the same as for other types of contracts.)
```
jpmckinney:
```suggestion
The following guidance describes how to model the different stages of a framework agreement procedure in OCDS. (It does not describe how to model the contract (using the `contracts` array), because the modeling is the same for a framework agreement as for other types of contracts.)
```

docs/guidance/map/framework_agreements.md
```diff
@@ -81,23 +81,24 @@ The two extensions should be declared in the package metadata:
...
   * `tender.techniques.hasFrameworkAgreement` to `true`.
   * `tender.contractPeriod` to the duration of the framework agreement.
   * If the framework agreement is closed, set `tender.tenderPeriod.endDate` to the deadline for responses to the invitation.
   * If the framework agreement is open, set `tender.tenderPeriod.endDate` to the last date that new suppliers can be added.
 
 ### First-stage selection and addition of suppliers to the framework agreement
 
-* Create a release with the **same** `ocid` as the tender release and add 'award' to the `tag` array
+* Create a release with the **same** `ocid` as the tender release and add 'award' to the `tag` array.
 * Add an `Award` object to the `awards` array.
```
jpmckinney:
```suggestion
* Add an `Award` object to the `awards` array, setting its fields as usual.
```

docs/guidance/map/framework_agreements.md
```diff
@@ -106,46 +107,43 @@ The following guidance describes how to model the different stages of a framewor
...
+* Add an `Organization` object to the `parties` array, add 'buyer' to its `.roles` and populate its other fields with the details of the buyer for this specific award.
 * Set `buyer.id` and `buyer.name` to match the buyer's object in the `parties` array.
 
-#### Add an award, contract and supplier
+#### Add an award and supplier
 
-* Add an `Award` object to the `awards` array and a `Contract` object to the `contracts` array.
-* Set the award's `.id` and the contract's `.id` to the same value.
+* Add an `Award` object to the `awards` array.
+* Set the award's `.id`.
```
jpmckinney:
```suggestion
* Add an `Award` object to the `awards` array, setting its fields as usual.
```

docs/guidance/map/framework_agreements.md
```diff
@@ -106,46 +107,43 @@ The following guidance describes how to model the different stages of a framewor
...
 * Add an `Organization` object to the `parties` array, add 'supplier' to its `.roles` and populate its other fields.
 * Add an `OrganizationReference` object to the award's `.suppliers` array and set `.id` and `.name` to match the supplier's object in the `parties` array.
-* Populate the award's `.value`, `.contractPeriod` and `.items` with the initial contract value, period and items.
+* Populate `awards`.
```
jpmckinney:
```suggestion
```

docs/guidance/map/framework_agreements.md
JachymHercher: Sounds good, done.

JachymHercher:
Nice simplifying!

docs/guidance/map/framework_agreements.md
JachymHercher: Generally better, thanks. The one point I would slightly defend: I think speaking about "second-stage contracting processes linked to their first-stage contracting process", i.e. using the plural, is useful/educative, because it drills in the fact that there is a one-to-many relationship there. (Which is possibly not the first thing you think of when you hear "first stage" and "second stage"). On the other hand, I understand we try to speak in the singular.

I've proposed to put the plural back in https://github.com/open-contracting/standard/pull/1461/commits/f5d3209655b8549c223108b84a775a804c7a11c4.

jpmckinney: Yes, sounds good.

jpmckinney (approved):

PR #1464, opened 2021-12-22
`documentType`: Deprecate 'eligibilityCriteria', add 'exclusionGrounds' and 'selectionCriteria'

### LINKED ISSUE #901: Deprecate eligibilityCriteria and add exclusionGrounds

JachymHercher:
Closes #901


schema/codelists/documentType.csv
```diff
@@ -19,7 +19,8 @@ implementation,financialProgressReport,Financial progress reports,"Documentation
 implementation,finalAudit,Final audit,Documentation of a final audit carried out at the end of contract implementation.,,
 tender,hearingNotice,Public hearing notice,Documentation of any public hearings that took place as part of the planning for this contracting process.,,
 planning,marketStudies,Market studies,Documentation of any market studies that took place as part of the planning for this contracting process.,,
-tender,eligibilityCriteria,Eligibility criteria,Detailed documents about the eligibility of bidders.,,
+tender,eligibilityCriteria,Eligibility criteria,Detailed documents about the eligibility of bidders.,2.0,This code had unclear semantics and used ambiguous terms, and is deprecated in favor of 'exclusionGrounds'.
```
jpmckinney:
```suggestion
tender,eligibilityCriteria,Eligibility criteria,Detailed documents about the eligibility of bidders.,1.2,This code had unclear semantics and used ambiguous terms, and is deprecated in favor of 'exclusionGrounds'.
```

schema/codelists/documentType.csv
```diff
@@ -42,8 +43,9 @@ implementation,debarments,Debarments,Documentation of any debarments issued.,,
 contract,contractSummary,Contract summary,Documentation providing an overview of the key terms and sections of the contract. Commonly used for large and complex contracts.,,
 "tender, award, contract, implementation",cancellationDetails,Cancellation details,"Documentation of the arrangements, or reasons, for cancellation of a contracting process, award or specific contract.",,
 "tender, award",unsuccessfulDetails,Unsuccessful details,Documentation of the reasons for an unsuccessful award or contracting process.,,
-tender,economicSelectionCriteria,Economic selection criteria,Documentation that describes the economic selection criteria.,,
-tender,technicalSelectionCriteria,Technical selection criteria,Documentation that describes the technical selection criteria.,,
+tender,selectionCriteria,Selection criteria,Documentation that describes the minimum requirements for potential tenderers to participate in the contracting process.
```
jpmckinney: According to the style guide we prefer "potential suppliers" https://ocds-standard-development-handbook.readthedocs.io/en/latest/meta/style_guide.html#word-choice

Same for the other two codes below.

Can you make a corresponding change to the `selectionCriteria` schema field (which currently uses "tenderers")?

JachymHercher: I chose "potential tenderer" after looking at https://github.com/open-contracting/standard/issues/1398#issuecomment-924337184 (1), as selection criteria are applied to companies which I would consider "likely" tenderers, i.e. a step up from "potential suppliers"? However, it's probably unnecessarily complicated and happy to stick with "potential suppliers" everywhere? 

(Ultimately, you could probably even argue that companies sending "requests to participate" are already participating in the contracting process, which would suggest a change in `selectionCriteria`. If we were working on a blank slate, this would seem logical to me (e.g. I'd say that companies that just download procurement documents or submit questions are not participating, while companies that actively submit information about themselves are); on the other hand, since this would precisely mean that "if you submit a request to participate then you are already participating" that would be way too unintuitive / against established terminology, so it's probably a bad idea.)

jpmckinney: When I re-read that comment, I just see a reason for it to be "potential supplier" :)

I think the rest would be too great a departure from existing terminology. UNCITRAL defines "qualification criteria" as:

> Criteria used by the procuring entity in ascertainment of eligibility of suppliers or contractors to participate in the procurement proceedings, as specified in the pre-qualification or pre-selection documents where applicable and in the solicitation documents.

So from that we can take away that an application to pre-qualify cannot itself be participation. Otherwise you end up in a contradiction, whereby suppliers are already participating by having submitted an application to pre-qualify, before the buyer is able to apply the criteria for determining their eligibility to participate.

That said, UNCITRAL also defines "Participation by suppliers or contractors in the procurement proceedings" as:

> Taking part by suppliers or contractors at any stage of the procurement proceedings starting from the moment of presentation of an application to pre-qualify, application for pre-selection or a submission.

All that said, we will not reverse the tide of common practice, and so we will leave it as-is to avoid introducing our own confusing situation on top of an existing confusing situation.

JachymHercher: Done in https://github.com/open-contracting/standard/commit/20ac64ae1e398eb28e4138cfbc3e5859a6b54964. I also changed "contract to be awarded" to "contract", as I don't think "to be awarded" adds anything in this context and got there just as a copy-paste from the Directive which uses the same formulation.

jpmckinney (approved):

PR #1472, opened 2022-02-02
docs: remove planning reference from versioned release example

### LINKED ISSUE #1362: Update guidance related to the planning stage/process for OCDS 1.2
Changes that can be made in 1.1 (to avoid conflicts):

* [x] Update the unsuccessful processes worked example to have one ocid for the planning stage, another for the unsuccessful tender, and a third for the successful tender.
   For the third ocid, we can either have two entries in the related process array, for the unsuccessful tender and the planning process, or we can just have one for the unsuccessful tender.
* [x] Update all the JSON examples where the planning object is used. In cases where the planning object is not relevant to the example, it can simply be removed from the JSON.
* [x] Check whether to rephrase "planning stage" in these worked examples: milestones, organizational_units, unsuccessful_processes

Changes that need to be made in 1.2:

* [x] Update the versioned release example at the bottom of `records_reference.md`
* [x] In `codelists.md`, replace "the requirements identified at the planning stage" with "identified needs"

yolile:
closes #1362

### DIFF
--- a/docs/history/changelog.md
+++ b/docs/history/changelog.md
@@ -237,7 +237,6 @@ Per the [normative and non-normative content and changes policy](https://docs.go
   * [#1067](https://github.com/open-contracting/standard/pull/1067) `Publisher.name`, to indicate that it is the organization or department responsible for publishing the OCDS version of the data.
 
 * [#1450](https://github.com/open-contracting/standard/pull/1450) Replace a repeated example in schema/merging/ with a link to guidance/build/merging/.
-* [#1450](https://github.com/open-contracting/standard/pull/1472) Remove planning reference from the versioned release example.
 
 ### Documentation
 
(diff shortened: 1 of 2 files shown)


docs/history/changelog.md
```diff
@@ -237,6 +237,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
   * [#1067](https://github.com/open-contracting/standard/pull/1067) `Publisher.name`, to indicate that it is the organization or department responsible for publishing the OCDS version of the data.
 
 * [#1450](https://github.com/open-contracting/standard/pull/1450) Replace a repeated example in schema/merging/ with a link to guidance/build/merging/.
+* [#1450](https://github.com/open-contracting/standard/pull/1472) Remove planning reference from the versioned release example.
```
jpmckinney: I don't think we need this entry. I assume the example had been updated, but we just forgot to change the text here. We can add 1472 to that other entry (haven't checked), or we can just not mention 1472.

yolile: Apparently, the JSON file never contained a planning stage https://github.com/open-contracting/standard/commit/42b42089ca968e004359c41cd5a743f7cf0b051f, and the text was like that since the begging of time https://github.com/open-contracting/standard/commit/b83f1be62fe2a65c511fa2c6b9415d0c1a9f9609. So I will remove this entry from the changelog

jpmckinney (approved):

Issue #881, opened 2019-06-14
Add additional schema validation


jpmckinney:
* `email`: `"format": "email"`
* `unit.value->amount`: `"minimum": 0` (requires a new definition instead of Value)
* `quantity`: `"minimum": 0`
* `durationInDays`: `"minimum": 0`
* `numberOfTenderers`: `"minimum": 0`

If different definitions are used for each documents array, then there can be separate codelists for each, instead of having additional logic around the Section column.

Issue #859, opened 2019-04-04
All strings, arrays, objects should have minLength, minItems, minProperties


jpmckinney:
If a field is optional, these validation properties are ignored, unless the field is present. OCDS presently has a rule that fields *should not* be empty. This issue is about changing the rule to *must not* be empty, which is backwards-incompatible.

Issue #814, opened 2019-02-13
planning, tender and contract sections allow duplicate milestones


duncandewhurst:
In `implementation.milestones` the `uniqueItems` property is set to `true`, however the `milestones` arrays in `planning`, `tender` and `contract` do not have this property set.

This means that data with duplicate milestones (or duplicate milestone ids) in these sections doesn't result in a validation error.

The definition of `milestone.id` states:

> A local identifier for this milestone, unique within this block

So `uniqueItems` should be set to `true` in all `milestones` arrays.

Adding this to the 2.0 milestone as it could make existing data invalid.

I'll open an issue on the CoVE tracker to add an additional check for duplicate milestones ids until we can add this to the schema.

cc @pindec

jpmckinney:
I noticed this issue in extensions as well. https://github.com/open-contracting/ocdskit/issues/92 can help with this.

odscjen:
Is this todo ready for a PR to be created?

jpmckinney:
This is in the 1.2.0 Strict milestone, where the idea is to add a "strict" version of the schema alongside the main version, so that publishers can opt-in to stricter rules (otherwise, we have to create a 2.0 release to apply the strict rules).

However, we haven't started on the documentation, etc. needed for a strict version. So this depends on #1046

Issue #718, opened 2018-05-24
Package schema need not allow null values


jpmckinney:
In releases, null has a particular meaning in terms of merging. In packages, null has no special meaning and, in general, OCDS prefers to omit fields rather than include fields with null values.

This would be a backwards-incompatible change, so would be a consideration for OCDS 2.0. As the presence of null values is not a serious issue, we may chose to continue to allow them, in order to lower the implementation cost of the 2.0 upgrade.

See #677 for prior discussion.

duncandewhurst:
I think it makes sense to remove them. Otherwise, the meaning of null values would be inconsistent between packages and releases. By providing a strict schema in OCDS 1.2, fewer publishers are likely to need to do this as part of an upgrade to 2.0.

PR #1480, opened 2022-02-10
Make strict schema

### LINKED ISSUE #1046: Consider a "strict" schema to introduce JSON Schema validation keywords prior to 2.0
Several issues deal with incorrect or missing JSON Schema validation keywords. See milestone: https://github.com/open-contracting/standard/milestone/25

Typically, we would need to wait for a major version (2.0) to make changes that invalidate data from earlier versions (i.e. backwards-incompatible changes).

However, we can offer a voluntary, "strict" schema that adds in these keywords, removes deprecated fields and codes, removes deprecated types, etc.

Users of the Data Review Tool could voluntarily select the strict schema. Similarly, OCDS Helpdesk analysts could inform publishers of opportunities to improve their data by passing the stricter criteria.

duncandewhurst:
Related to #1046

Depends on https://github.com/open-contracting/ocdskit/pull/184

Related issues from 1.2.0 Strict milestone:

- [x] #881
- [x] #859
- [x] #814
- [x] #679
- [x] #718
- [x] #669
- [x] #650

TO-DO:

- [x] Remove deprecated types
- [x] Decide how to keep the strict schema up to date with changes to the regular schema
- [x] Update strict schema to match 1.2-dev
- [ ] ~~Update requirements.txt once https://github.com/open-contracting/ocdskit/pull/184 is released on PyPI~~
- [ ] ~~Add documentation~~ (To be done in a separate PR)
- [x] Update changelog
- [x] Create a compiled-release-schema.json and remove "null" types and null enums from all fields.


duncandewhurst:
>  Decide how to keep the strict schema up to date with changes to the regular schema

@jpmckinney - so far I made the changes in this PR manually, but in order to keep the strict schema up to date, I can add a script as part of `manage.py pre-commit` to derive the strict schema programmatically. Does that sound good?

jpmckinney:
Yes, let's do it that way. Then we can run the script and see if the result is currently the same as your manual preparation :)

duncandewhurst:
Unfortunately, I'd made the manual changes to an old version of the schema and since merged the latest version of 1.2-dev so it's easier to just to diff the regular and strict schemas to check that the changes are as expected:

```bash
git diff --word-diff=color --no-index schema/release-schema.json schema/strict/release-schema.json
git diff --word-diff=color --no-index schema/release-package-schema.json schema/strict/release-package-schema.json
git diff --word-diff=color --no-index schema/record-package-schema.json schema/strict/record-package-schema.json
```

duncandewhurst:
The tests are failing on an unrelated issue from the organization units example and on the schema file for the new `UnitValue` definition (suggested in https://github.com/open-contracting/standard/issues/881). I'm not sure if there's a better way to add that programmatically. Hardcoding it in `manage.py` seemed like a bad idea.

jpmckinney:
> I'm not sure if there's a better way to add that programmatically.

Add what? The UnitValue definition? I figure we can add it to the regular schema (without extra validation keywords), and then just add the validation keywords in the strict schema.

duncandewhurst:
> > I'm not sure if there's a better way to add that programmatically.
> 
> Add what? The UnitValue definition? I figure we can add it to the regular schema (without extra validation keywords), and then just add the validation keywords in the strict schema.

I've added the UnitValue definition to the regular schema and updated `manage.py` to set the `$ref` for `Unit.Value` in the strict schema.

Do you have a view on if and how we should feature the strict schema(s) in the documentation?

I figure few users would do anything with the actual JSON schema, but adding schema viewers for the strict versions of the release, release package and record package schemas seems overly duplicative. I think we should include some mention of it in the docs if we are going to include it in the DRT and helpdesk feedback processes. We could either add an admonition to https://standard.open-contracting.org/staging/1.2-dev/en/schema/# or a new page under the reference section which can give a summary of the differences to the regular schema and direct users to the DRT to check their data against the strict schema.

jpmckinney:
> Do you have a view on if and how we should feature the strict schema(s) in the documentation?

I think a concise admonition as you suggest would be best. It should briefly describe why the strict schema is preferred for new implementations (and how it can be useful to improve data quality of existing implementations).

In terms of duplication, we can consider using https://sphinx-panels.readthedocs.io/en/latest/ to allow browsing of each schema. That way it's not taking up a lot of the page.

duncandewhurst:
> In terms of duplication, we can consider using https://sphinx-panels.readthedocs.io/en/latest/ to allow browsing of each schema. That way it's not taking up a lot of the page.

I tried this approach but only the first docson widget is displayed. The second widget throws an Uncaught Error: only one instance of babel-polyfill is allowed (see https://github.com/lbovet/docson/issues/73).

In other projects that have multiple docson widgets per page we use https://github.com/OpenDataServices/docson which is a fork of an earlier version of docson that doesn't use babel-polyfill, but I see that https://github.com/open-contracting/docson makes several improvements that would be lost if we used the other fork.

I found https://www.npmjs.com/package/idempotent-babel-polyfill/ which might be a solution but my Javascript knowledge falls short of how to update docson to use it.

duncandewhurst:
@jpmckinney based on my previous comment, do you still want to use Sphinx Panels?

jpmckinney:
> In other projects that have multiple docson widgets per page we use https://github.com/OpenDataServices/docson which is a fork of an earlier version of docson that doesn't use babel-polyfill

Are you sure? I looked at several branches at https://github.com/OpenDataServices/docson/branches, comparing them to the source repository, and I don't see any removal of babel-polyfill.

Anyway, let's do this in different PRs:

1. Make the changes for `manage.py` and `UnitValue`.
2. Start drafting the changes to the documentation. When ready to merge, run the `manage.py` command. As part of this, I can sort out removing the polyfill, which I think is not necessary for today's browsers.

duncandewhurst:
> > In other projects that have multiple docson widgets per page we use https://github.com/OpenDataServices/docson which is a fork of an earlier version of docson that doesn't use babel-polyfill
> 
> Are you sure? I looked at several branches at https://github.com/OpenDataServices/docson/branches, comparing them to the source repository, and I don't see any removal of babel-polyfill.

The ODS fork doesn't remove babel-polyfill. Rather, babel-polyfill was added to the original docson repo after the ODS repo was forked and before the open-contracting repo was forked.

In https://github.com/OpenDataServices/docson/compare/master...lbovet:docson:master there is a commit from the original repo that adds babel-polyfill: https://github.com/OpenDataServices/docson/commit/0756db7653257b9d438bb0536932a8020ae46f66, which also appears in the open-contracting fork: https://github.com/open-contracting/docson/commit/0756db7653257b9d438bb0536932a8020ae46f66

> Anyway, let's do this in different PRs:
> 
>     1. Make the changes for `manage.py` and `UnitValue`.

These changes were already made in this PR: https://github.com/open-contracting/standard/pull/1480/commits/85af27bee511efb3a1b0c04e016a333aa067ae15. Is that OK?

>     2. Start drafting the changes to the documentation. When ready to merge, run the `manage.py` command. As part of this, I can sort out removing the polyfill, which I think is not necessary for today's browsers.

So create a new PR for the documentation changes?

jpmckinney:
Noting that sphinx-panels is obsolete. Need to use https://github.com/executablebooks/sphinx-design.

jpmckinney:
To clarify my earlier request in https://github.com/open-contracting/standard/pull/1480#issuecomment-1198822301

Let's try to get something merged sooner than later.

To do that, let's not generate the `strict` files in this PR. You can remove those files, and I'll "squash and merge". That means this PR is part 1 (manage.py and UnitValue).

In a new PR, we can add the documentation (the last checklist item). Once the docs are ready, we can run the `manage.py` command to generate the `strict` files, and then merge.

jpmckinney:
Note to myself that I haven't actually reviewed the changes yet, as I'm not sure if the PR is only "draft" due to the missing documentation changes.

duncandewhurst:
> Let's try to get something merged sooner than later.

Erm... I'm not sure how I missed this one, sorry! It's now ready for review. I included a changelog entry for `UnitValue` even though it makes no difference to users. Feel free to remove, or we can just update it once the part 2 PR is done.

duncandewhurst:
I've updated this PR to reflect the record schema being moved into its own file and to reflect the changes made to `manage.py` on `1.2-dev`.

Edit: I also temporarily removed `ruff-format` from `.pre-commit-config.yaml`. It can be reinstated once https://github.com/open-contracting/standard/pull/1708#issuecomment-2433776370 is done.

odscjen:
Added a new to-do from https://github.com/open-contracting/standard/issues/330#issuecomment-2450280930

duncandewhurst:
Have moved back to draft whilst I make those changes.

PR #1482, opened 2022-03-01
docs: add release schema reference examples

### LINKED ISSUE #1348: Add example data to schema reference
For example, these short snippets of common objects:

https://github.com/open-contracting/standard/blob/f73f4c9/docs/getting_started/building_blocks.md#examples

yolile:
closes #1348


yolile:
@jpmckinney added one example to the parties section to check if this is what is expected before adding it to all the sections

docs/schema/reference.md
```diff
@@ -118,6 +118,14 @@ In OCDS 1.0, the details (address, contact point, etc.) of the organizations inv
 Note that the organization references allow, but deprecate, the fields for organization details.
 ```
 
+#### Example
+
+```{jsoninclude} ../examples/release_schema_reference/release_package.json
+:jsonpointer:
```
jpmckinney: I think we want the `jsonpointer` to be the field, to avoid the clutter from the package.

In this case, I think we can expand all the fields within the party.

It might be nice to just show one party as well (or to collapse the second party).

https://standard.open-contracting.org/staging/1348-schema-examples/en/schema/reference/#parties

yolile: > In this case, I think we can expand all the fields within the party.

Makes sense; what about the fields for planning, tender, etc.? Do we want to expand them all as well? Or maybe all of them but the building blocks? e.g., planning/documents, milestones, etc

yolile: And, what about the order of the fields? Can we preserve the same order as the schema definition, or do we sort the fields alphabetically (as per `ocdskit indent`)?

yolile: And do we want the examples to have all the fields from the object? E.g., for tender, should the example contain all the existing fields from the schema?

jpmckinney: > what about the fields for planning, tender, etc.? Do we want to expand them all as well? Or maybe all of them but the building blocks?

I agree that we can collapse the multi-use sub-schema (periods, documents, milestones, etc.) to avoid repetition.

> what about the order of the fields?

`ocdskit indent` re-orders fields? It shouldn't. Anyway, I think it is best for the example file to use fields in schema order, for consistency.

> do we want the examples to have all the fields from the object?

We can have as many as makes sense. I think only the `tender` object has a large number of fields.

yolile: Done

yolile:
I tested using the `admonition` directive and the result looks like https://standard.open-contracting.org/staging/1348-schema-examples/en/schema/reference/#package-metadata in case we want to use that format instead of an "Example" subtitle and `jsoninclude` directly.

jpmckinney:
I like the use of admonition! It better chunks up the visual flow. For package metadata, we can expand `publisher`.

docs/schema/reference.md
```diff
@@ -373,6 +522,16 @@ A period has a start date, end date, and/or duration. Start and end dates are re
 
 Periods can also include a `maxExtentDate` which indicates the latest possible end date of this period, or the latest date up until which the period could be extended without an amendment.
 
+````{admonition} Example
+:class: hint
+
+```{jsoninclude} ../examples/release_schema_reference/release_package.json
+:jsonpointer: /releases/0/awards/0
```
jpmckinney: Should this jsonpointer be directly to the period object? Since it's reused it many contexts, not just award.

Same question for value later.

docs/schema/reference.md
```diff
@@ -416,6 +584,16 @@ The items block is used to list the line-items associated with a tender, award o
 
 #### Classification
 
+````{admonition} Example
```
jpmckinney: Since we now also use Classification outside the context of Items, we maybe need to change the heading level to `###` (and move it after Unit). If so, we should also make the example specific to the classification object, instead of showing the context of the item.

docs/schema/reference.md
```diff
@@ -478,6 +685,15 @@ In OCDS each contracting process can have only one planning and tender stage. Th
 
 In all these cases, the `relatedProcess` block should be used to cross-reference between the relevant open contracting processes using their `ocid`.
 
+````{admonition} Example
+:class: hint
+
+```{jsoninclude} ../examples/release_schema_reference/release_package.json
```
jpmckinney, line 689: Just noticed that `relationship` isn't an array in the example.

docs/schema/reference.md
yolile: I wasn't sure about it, as I thought it might be more useful to show the field use with context, but I agree with your point.

docs/schema/reference.md
yolile: True, changed now

docs/schema/reference.md
yolile: Fixed and passed the file through the review tool to make sure anything else was missing

jpmckinney (approved):

PR #1488, opened 2022-03-15
815 simple identifier

### LINKED ISSUE #815: Add Tender.identifier(s) and Contract.identifier(s) fields
https://github.com/gitMHOCDS/dncp-contract-code adds contracts.dncpContractCode (string)

> For each contract, the DNCP generates a contract code that differs from the id of the contractId it self, and this information is relevant because it helps users to link a contract in Hacienda with a contract in DNCP. Users are very used to working with this information in Paraguay, and as such it is needed for the Paraguayan users of the data.

https://github.com/openprocurement/ocds_contractNumber_extension adds contracts.contractNumber (string)

> This extension introduces one field that can be used to provide editable number of the contract.

The earlier discussion in #289 provides additional examples.

Rather than the approach in #289 – which is for each publisher to create their own very similar but different extension – the proposal here is to add to `Contract` either a `identifier` field for a single `Identifier`, or an `identifiers` field for an array of `Identifier`.

The two local extensions suggest only `identifier` is needed.

cc @yolile @duncandewhurst @myroslav

odscjen:
Hi Duncan,
There's still an error in the tests but I can't figure out how to fix it. This PR is related to #815. I've not updated the changelog yet as I couldn't work out how to do that.
Cheers

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,11 +6,28 @@ Visit the [OCDS Standard Development Handbook](https://ocds-standard-development
 
 ## Maintenance
 
-### Update merging examples
+Install [OCDS Kit](https://pypi.org/project/ocdskit/)
 
-You will need [OCDS Kit](https://pypi.org/project/ocdskit/) to update the examples in `docs/examples/merging`.
+Update the examples in `docs/examples/merging`:
 
 ```shell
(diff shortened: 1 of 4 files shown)


duncandewhurst (changes requested):
@odscjen - you'll need to redo this branch from `1.2-dev`. It should just be a case of pasting the new properties and definitions into `release-schema.json`, running `./manage.py pre-commit` and then updating `docs/history/changelog.md`.

odscjen:
have made the changes to a new branch and therefore a new PR, #1490

PR #1490, opened 2022-03-16
add simpleIdentifier to 1.2-dev

### LINKED ISSUE #815: Add Tender.identifier(s) and Contract.identifier(s) fields

odscjen:
closes #815 and resolves #1488

### DIFF
--- a/schema/release-schema.json
+++ b/schema/release-schema.json
@@ -213,7 +213,8 @@
           "items": {
             "$ref": "#/definitions/SimpleIdentifier"
           },
-          "uniqueItems": true
+          "uniqueItems": true,
+          "wholeListMerge": true
         },
         "title": {
           "title": "Tender title",
@@ -741,7 +742,8 @@
           "items": {
             "$ref": "#/definitions/SimpleIdentifier"
           },
-          "uniqueItems": true
+          "uniqueItems": true,
+          "wholeListMerge": true
         },
         "awardID": {
           "title": "Award ID",
@@ -2675,7 +2677,7 @@
       "properties": {
         "scheme": {
           "title": "Scheme",
-          "description": "The list, register or system from which identifiers are taken.",
+          "description": "The list, register or system from which the identifier is taken.",
           "type": [
             "string",
             "null"
--- a/schema/release-schema.json
+++ b/schema/release-schema.json
@@ -2672,7 +2672,7 @@
           ]
         },
         "id": {
-          "title": "A description of the goods, services to be provided.",
+          "title": "ID",
           "description": "The identifier taken from the scheme.",
           "type": [
             "string",
--- a/schema/release-schema.json
+++ b/schema/release-schema.json
@@ -205,6 +205,15 @@
           ],
           "minLength": 1
         },
+        "identifiers": {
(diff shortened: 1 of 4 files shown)


schema/release-schema.json
```diff
@@ -2639,6 +2657,30 @@
...
+        "scheme": {
+          "title": "Scheme",
+          "description": "The list, register or system from which identifiers are taken.",
+          "type": [
+            "string",
+            "null"
+          ]
+        },
+        "id": {
+          "title": "A description of the goods, services to be provided.",
```
duncandewhurst:
```suggestion
          "title": "ID",
```

duncandewhurst (changes requested):
@jpmckinney I think the tests need updating so that `id` not being required on `SimpleIdentifier` doesn't cause a failure.

duncandewhurst (approved):

schema/release-schema.json
```diff
@@ -2639,6 +2657,30 @@
...
       }
+    },
+    "SimpleIdentifier": {
+      "title": "Simple identifier",
+      "description": "An unambiguous reference to a resource within a given context.",
+      "type": "object",
+      "properties": {
+        "scheme": {
+          "title": "Scheme",
+          "description": "The list, register or system from which identifiers are taken.",
```
jpmckinney: Change word agreement, since this is in the context of one identifier.

```suggestion
          "description": "The list, register or system from which the identifier is taken.",
```

jpmckinney:
> @jpmckinney I think the tests need updating so that `id` not being required on `SimpleIdentifier` doesn't cause a failure.

What should the merge behavior be? For `Organization.additionalIdentifiers`, we use `"wholeListMerge": true,`. If we were to add that here, then the tests will not error.

If we want to merge by identifier, then I think `id` needs to be required.

duncandewhurst:
Ah, good point. I think `wholeListMerge` makes sense, since `SimpleIdentifier.id` need only be unique within the scope of `.scheme`.

jpmckinney:
I've made that changed and resolved my comment. I just need one approval before merge. Note my commit has some extra changes in the dereferenced file, as pre-commit hadn't been run in some earlier commit.

duncandewhurst (approved):

PR #1492, opened 2022-03-29
add awards.datePublished field

### LINKED ISSUE #1456: Add awards.datePublished field
Branching off from the discussion on adding a `tender.datePublished` field (#892), the UK also has a need to add an `awards.datePublished` field.

This will capture the date that the award was published. We will create a new extension if necessary. This will allow this information to be included in awardUpdate releases where it wouldn't be otherwise. We need to cater for analysis tools that only hold the latest release for each procurement (so the awardUpdate overwrites the award, for example).

odscjen:
adding awards.datePublished field to close issue #1456

### DIFF
--- a/schema/release-schema.json
+++ b/schema/release-schema.json
@@ -619,7 +619,7 @@
         },
         "datePublished": {
           "title": "Date published",
-          "description": "The date that the award was published.",
+          "description": "The date on which the award was published.",
           "type": [
             "string",
             "null"
--- a/schema/release-schema.json
+++ b/schema/release-schema.json
@@ -617,6 +617,15 @@
           ],
           "format": "date-time"
         },
+        "datePublished": {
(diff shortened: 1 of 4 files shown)


schema/release-schema.json
```diff
@@ -617,6 +617,15 @@
           ],
           "format": "date-time"
         },
+        "datePublished": {
+          "title": "Date published",
+          "description": "The date that the award was published.",
```
duncandewhurst: For consistency with `tender.datePublished`:

```suggestion
          "description": "The date on which the award was published.",
```

duncandewhurst (changes requested):

duncandewhurst:
Happy for you to request a review from @jpmckinney once you commit my suggestion.

jpmckinney (approved):

Issue #395, opened 2016-10-10
contractStatus: Add termination codes


timgdavies:
@lindseyAM has raised the concern that some implementers are finding the codelist entry for [Contract Status](http://standard.open-contracting.org/latest/en/schema/codelists/#contract-status) of `terminated` contains too negative a connotation. 

Lindsey states:

> there was feedback that terminate was too negative (sounding like the contract was ended for failure to perform etc. rather than that the contract simply ended) 

The current codelist is shown below:

| Code | Title | Description |
| --- | --- | --- |
| pending | Pending | This contract has been proposed, but is not yet in force. It may be awaiting signature. |
| active | Active | This contract has been signed by all the parties, and is now legally in force. |
| cancelled | Cancelled | This contract has been cancelled prior to being signed. |
| terminated | Terminated | This contract was signed and in force, and has now come to a close. This may be due to successful completion of the contract, or may be early termination due to some non-completion. |

Views on this issue are welcome.

timgdavies:
@ekoner Is there a quick way for us to check if particular organisations are using or not using 'terminated' as a code in their data?

chrisalexsmith:
I agree with the concern. Suggest 'terminated' is more appropriate to an active contract that has for some reason been cancelled after signature (either in part or in full) and that 'completed' better describes a  contract where all obligations have been fulfilled. In fact it might be helpful to differentiate between the two types.

timgdavies:
ResourceContracts.org have also described a need to capture a distinction between:

* Contracts terminated prior to completion
* Contracts terminated at completion

There is also the case of 'expired' contracts. I.e. not formally terminated in any way, but past the dates when they would be active. 

## Considerations

There are a couple of considerations here:

1. We need to be careful to distinguish the concept of contract status (in terms of whether it is force or not) with concepts of contract outcome (actual delivery of the contracted services being complete or not)

2. We need to think about how this data would be supplied: do systems store information that would allow them to choose between termination statuses (this requires some degree of implementation information to be managed)

3. We need to think about how users would draw on this data: **if** we can be sure there will be reliable data generated by some systems on contract status (distinguishing completed and terminated contracts), are there clear use cases for this - and forms of analysis it enables? And analysis that could not be completed using other existing OCDS fields? (i.e. we want to avoid modelling the same thing in multiple ways). 

[...]

PR #1509, opened 2022-05-13
`tenderStatus`, `awardStatus` `contractStatus`: update codes

### LINKED ISSUE #1160: Update all status codelists

JachymHercher:
* Related to #1160


schema/codelists/awardStatus.csv
```diff
@@ -1,5 +1,6 @@
 Code,Title,Description
-pending,Pending,"This award has been proposed, but is not yet in force. This can be due to a cooling off period, or some other process."
```
jpmckinney, line 2: This is a closed codelist, so the code must be deprecated, not removed. I've added the columns, but I could use your help in drafting the value of the "Deprecation note" cell.

docs/history/changelog.md
```diff
@@ -146,6 +146,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
 * [#1200](https://github.com/open-contracting/standard/pull/1200) Deprecate the `submissionMethod` codelist, because its codes were either not submission methods, or not supported by use cases.
 * [#1209](https://github.com/open-contracting/standard/pull/1209) Replace "electronic goods" with "digital goods" in the description of 'goods' from the `procurementCategory` codelist, to align with the description in the World Trade Organization's Agreement on Government Procurement, and to avoid confusion between electronic goods like computers and digital goods like software.
 * [#1389](https://github.com/open-contracting/standard/pull/1389) Deprecate the `initiationType` codelist, because the approach to data modelling that it supports was not pursued.
+* [#1509](https://github.com/open-contracting/standard/pull/1509) Update the `tenderStatus`, `awardStatus` and `contractStatus` codelists, including major change of descriptions, removal of 'planning' and 'pending' and the addition of 'withdrawn', 'complete', 'cancelled' and 'unsuccessful' to codelists where they were not before.
```
jpmckinney: Let's create a bulleted list (like for `releaseTag` etc.) about these codelists, and add the above bullet about 'terminatedEarly' and 'terminatedSuccessfully' to this list.

contractStatus has no 'complete' code, unlike the others. Should we rename 'terminatedSuccessfully'? It's part of 1.2, so can be renamed.

Instead of removal, we can only do deprecation.

schema/codelists/tenderStatus.csv
```diff
@@ -1,8 +1,8 @@
...
-planned,Planned,"A contracting process is scheduled, but is not yet taking place. Details of the anticipated dates can be provided in the tender block."
-active,Active,A tender process is currently taking place.
-cancelled,Cancelled,The tender process has been cancelled.
-unsuccessful,Unsuccessful,The tender process was unsuccessful.
-complete,Complete,The tender process is complete.
-withdrawn,Withdrawn,No further information on this process is available under this ocid.
+Code,Title,Description,Deprecated,Deprecation note
+planning,Planning,"A future contracting process is being considered. Early information about the process can be provided in the tender section. A process with this status might provide information on early engagement or consultation opportunities, during which the details of a subsequent tender can be shaped.",1.2,
+planned,Planned,"The contracting process is scheduled, but is not yet taking place. Details of the anticipated dates can be provided in the tender block.",,
+active,Active,The procurement documents have been published and the bid submission deadline has not yet passed.,,
```
jpmckinney: OCDS has been used for PPPs and can ostensibly be used for sale of state assets, etc. Is there another term we can use that works as well as "procurement documents"? If not, then we can keep it as-is, as OCDS is primarily about procurement.

----

Also, for 'planned' we end with "is not yet taking place", and for 'active' we start with "The procurement documents have been published".

I guess it's possible that a process is planned, but the documents are never published (maybe they plan to do a direct award?), such that we need a less specific event for the end of the 'planned' status.

Just flagging in case there's some way to improve here.

schema/codelists/tenderStatus.csv
```diff
@@ -1,8 +1,8 @@
...
-planned,Planned,"A contracting process is scheduled, but is not yet taking place. Details of the anticipated dates can be provided in the tender block."
-active,Active,A tender process is currently taking place.
-cancelled,Cancelled,The tender process has been cancelled.
-unsuccessful,Unsuccessful,The tender process was unsuccessful.
-complete,Complete,The tender process is complete.
-withdrawn,Withdrawn,No further information on this process is available under this ocid.
+Code,Title,Description,Deprecated,Deprecation note
+planning,Planning,"A future contracting process is being considered. Early information about the process can be provided in the tender section. A process with this status might provide information on early engagement or consultation opportunities, during which the details of a subsequent tender can be shaped.",1.2,
+planned,Planned,"The contracting process is scheduled, but is not yet taking place. Details of the anticipated dates can be provided in the tender block.",,
```
jpmckinney: While we're here, let's remove reference material that pertains to a different field (`tenderPeriod`).

```suggestion
planned,Planned,"The contracting process is scheduled, but is not yet taking place.",,
```

schema/codelists/tenderStatus.csv
jpmckinney: Comma between full clauses. Present tense preferred over present perfect.

```suggestion
active,Active,"The procurement documents are published, and the bid submission deadline has not yet passed.",,
```

schema/codelists/tenderStatus.csv
```diff
@@ -1,8 +1,8 @@
...
+active,Active,The procurement documents have been published and the bid submission deadline has not yet passed.,,
+complete,Complete,The bid submission deadline has passed.,,
+cancelled,Cancelled,"The contracting process has been cancelled by the buyer or the procuring entity (e.g. because of a change in needs, insufficient funds, or technical or procedural errors) after the procurement documents have been published, but not later than the bid submission deadline.",,
+unsuccessful,Unsuccessful,"The contracting process failed after the procurement documents have been published, but not later than the bid submission deadline (e.g. no bids were received, all bids were withdrawn by the bidders, all bids were rejected by the buyer or the procuring entity).",,
```
jpmckinney: Use past tense for the past, instead of present perfect.

```suggestion
unsuccessful,Unsuccessful,"The contracting process failed after the procurement documents were published, but not later than the bid submission deadline (e.g. no bids were received, all bids were withdrawn by the bidders, all bids were rejected by the buyer or the procuring entity).",,
```

schema/codelists/tenderStatus.csv
```diff
@@ -1,8 +1,8 @@
...
-complete,Complete,The tender process is complete.
-withdrawn,Withdrawn,No further information on this process is available under this ocid.
+Code,Title,Description,Deprecated,Deprecation note
+planning,Planning,"A future contracting process is being considered. Early information about the process can be provided in the tender section. A process with this status might provide information on early engagement or consultation opportunities, during which the details of a subsequent tender can be shaped.",1.2,
+planned,Planned,"The contracting process is scheduled, but is not yet taking place. Details of the anticipated dates can be provided in the tender block.",,
+active,Active,The procurement documents have been published and the bid submission deadline has not yet passed.,,
+complete,Complete,The bid submission deadline has passed.,,
+cancelled,Cancelled,"The contracting process has been cancelled by the buyer or the procuring entity (e.g. because of a change in needs, insufficient funds, or technical or procedural errors) after the procurement documents have been published, but not later than the bid submission deadline.",,
```
jpmckinney:
```suggestion
cancelled,Cancelled,"The contracting process was cancelled by the buyer or the procuring entity (e.g. because of a change in needs, insufficient funds, or technical or procedural errors) after the procurement documents were published, but not later than the bid submission deadline.",,
```

schema/codelists/contractStatus.csv
```diff
@@ -1,7 +1,9 @@
-Code,Title,Description
-pending,Pending,"This contract has been proposed, but is not yet in force. It might be awaiting signature."
-active,Active,"This contract has been concluded by all the parties, and is now legally in force."
-cancelled,Cancelled,This contract has been cancelled prior to being concluded.
-terminated,Terminated,"This contract was concluded and in force, and has now come to a close. This might be due to the successful completion of the contract, or might be early termination due to some non-completion."
-terminatedEarly,Terminated early,"This contract was concluded and in force, and is terminated early due to some non-completion."
-terminatedSuccessfully,Terminated successfuly,"This contract was concluded and in force, and is terminated due to its successful completion."
+Code,Title,Description,Deprecated,Deprecation note
+pending,Pending,"This contract has been proposed, but is not yet in force. It might be awaiting signature.",1.2,
+active,Active,"The award has been made and the standstill period, if any, has ended. The contract has not yet terminated.",,
```
jpmckinney:
```suggestion
active,Active,"The award was made and the standstill period, if any, has ended. The contract is not yet terminated.",,
```

schema/codelists/contractStatus.csv
```diff
@@ -1,7 +1,9 @@
...
+terminated,Terminated,"The contract was concluded and in force, and has now come to a close. This might be due to the successful completion of the contract, or might be early termination due to some non-completion.",,
```
jpmckinney: With the change to 'active', we no longer have the words "in force" or "concluded". Should we just say, "The contract was active" for the terminated* codes? I'm also not fond of the second clause, but I won't edit it yet.

schema/codelists/contractStatus.csv
```diff
@@ -1,7 +1,9 @@
...
+terminatedSuccessfully,Terminated successfully,"The contract was concluded and in force, and is terminated due to its successful completion.",,
+terminatedEarly,Terminated early,"The contract was concluded and in force, and is terminated early due to some non-completion.",,
+cancelled,Cancelled,"The contracting process has been cancelled by the buyer or the procuring entity (e.g. because of a change in needs, insufficient funds, or technical or procedural errors) after the award has been made and the standstill period, if any, has ended.",,
```
jpmckinney:
```suggestion
cancelled,Cancelled,"The contracting process was cancelled by the buyer or the procuring entity (e.g. because of a change in needs, insufficient funds, or technical or procedural errors) after the award was made and the standstill period, if any, had ended.",,
```

schema/codelists/contractStatus.csv
```diff
@@ -1,7 +1,9 @@
...
+unsuccessful,Unsuccessful,"The contract failed (e.g. the winning bidder(s) refused to sign the contract) after the award had been made and the standstill period, if any, had ended, but before the contract was signed and in force.",,
```
jpmckinney:
```suggestion
unsuccessful,Unsuccessful,"The contract failed (e.g. the winning bidder(s) refused to sign the contract) after the award was made and the standstill period, if any, had ended, but before the contract was signed and in force.",,
```

jpmckinney: Similar to above, we mention "in force" without any code causing this state. This is also the only place we refer to signature.

schema/codelists/awardStatus.csv
```diff
@@ -1,5 +1,7 @@
-Code,Title,Description
-pending,Pending,"This award has been proposed, but is not yet in force. This can be due to a cooling off period, or some other process."
-active,Active,"This award has been made, and is currently in force."
-cancelled,Cancelled,This award has been cancelled.
-unsuccessful,Unsuccessful,"This award could not be successfully made. If items or supplier details are included within the award section, then these narrow the scope of the unsuccessful award (i.e. the award of noted items, or an award to the noted supplier, was unsuccessful, but there can be other successful awards for different items listed in the tender, or to different suppliers)."
+Code,Title,Description,Deprecated,Deprecation note
+pending,Pending,"This award has been proposed, but is not yet in force. This can be due to a cooling off period, or some other process.",1.2,
+active,Active,"The bid submission deadline has passed, but the award has not yet been made or the standstill period, if any, has not yet ended.",,
```
jpmckinney:
```suggestion
active,Active,"The bid submission deadline has passed, but the award is not yet made or the standstill period, if any, has not yet ended.",,
```

schema/codelists/awardStatus.csv
```diff
@@ -1,5 +1,7 @@
...
+complete,Complete,"The award has been made and the standstill period, if any, has ended.",,
```
jpmckinney:
```suggestion
complete,Complete,"The award was made and the standstill period, if any, has ended.",,
```

schema/codelists/awardStatus.csv
```diff
@@ -1,5 +1,7 @@
...
+cancelled,Cancelled,"This award has been cancelled by the buyer or the procuring entity (e.g. because of a change in needs, insufficient funds, or technical or procedural errors) after the bid submission deadline, but before the award had been made or before the standstill period, if any, had ended.",,
```
jpmckinney:
```suggestion
cancelled,Cancelled,"The award was cancelled by the buyer or the procuring entity (e.g. because of a change in needs, insufficient funds, or technical or procedural errors) after the bid submission deadline, but before the award was made or before the standstill period, if any, had ended.",,
```

schema/codelists/awardStatus.csv
```diff
@@ -1,5 +1,7 @@
...
+unsuccessful,Unsuccessful,"The award failed (e.g. a review body cancelled the process during the standstill period) after the bid submission deadline, but before the award had been made or before the standstill period, if any, had ended.",,
```
jpmckinney:
```suggestion
unsuccessful,Unsuccessful,"The award failed (e.g. a review body cancelled the process during the standstill period) after the bid submission deadline, but before the award was made or before the standstill period, if any, had ended.",,
```

schema/codelists/tenderStatus.csv
jpmckinney: Unlike all the others, here the parenthetical is at the end. Can we move it, for easier comparison by a reader looking to distinguish the two options?

jpmckinney (commented):
As mentioned in a comment, we need a "Deprecation note" for the deprecated codes in the CSV files. We should repeat this note in the changelog (it can be abbreviated if needed).

schema/codelists/awardStatus.csv
JachymHercher: Omg, sorry. Totally forgot to deprecate instead of remove. Deprecation notes added in https://github.com/open-contracting/standard/pull/1509/commits/f5aa66f42d8c05bc4935a78be96f0850b7d1f7c4.

docs/history/changelog.md
JachymHercher: Change-log done in https://github.com/open-contracting/standard/pull/1509/commits/f5aa66f42d8c05bc4935a78be96f0850b7d1f7c4. 

I think 'complete' would be better than `terminatedSuccessfully' and indeed nicely consistent. However, we went for 'terminatedSuccessfully' (in https://github.com/open-contracting/standard/issues/395#issuecomment-767203422) to make it clear that those are subcodes of 'terminated'. Do we want to break that? (My tendency would be to go for 'complete' and deprecate 'terminated' or consult on it, if we think we'll get useful feedback.)

schema/codelists/tenderStatus.csv
JachymHercher: We could say "important documents (for example procurement documents)" or "contracting documents (for example procurement documents)". (We can also add other examples than procurement documents, I just don't know the terms :). ) Note that we use "procurement documetns" also in `planning`, `Planning`, `tender` and `Tender`, so we'd need to change them there.

Concerning the procurement documents, e.g. for a direct award, you are right. In the stage definitions, it's supposed to fall under the looseness of "This information **typically** concerns...". However, in the status codes, this flexibility is no longer there. I guess we could replace "_procurement_ documents are published" everywhere (i.e codes and `Tender` etc.) with "_procurement_ documents are available to potential suppliers"? 

(Not "potential tenderers", even though we could - see https://github.com/open-contracting/standard/issues/1398#issuecomment-924337184 - just not to introduce expressions we don't need and OCDS currently doesn't use?)

schema/codelists/tenderStatus.csv
JachymHercher: Done in https://github.com/open-contracting/standard/pull/1509/commits/7722294d340880bb8ce1355b349dec1577db040b.

schema/codelists/contractStatus.csv
JachymHercher: I think we shouldn't say "active", because part of the 'active' definition is "Not yet terminated". 

For the broader question see next comment.

schema/codelists/contractStatus.csv
JachymHercher: Changed the one occurrence "signed and in force" to "concluded and in force" in https://github.com/open-contracting/standard/pull/1509/commits/ee1630e7554421a77272351a79538061b3b7321c. 

______

More broadly: I think we never properly discussed https://github.com/open-contracting/standard/issues/1160#issuecomment-887489821:

> I would have a tendency to drop 'pending' from both awardStatus and contractStatus. I'm assuming its main goal is to distinguish releases where an award has already been made or a contract already signed, but this information is clear from dateSigned and awardDate. The same argument could probably also be made against 'planned', 'active', 'complete' though, which would then lead to status changing into "reasons for failure"...?

The flow we're trying to model is award decision --> end of standstill (1) --> signature by last party --> contract in force --> contract completed/terminated (2).  In my proposal, at point 1, we set `awardStatus` = 'complete' and `contractStatus` to 'active'. Then we do nothing until 2, when we set `contractStatus` to 'terminated*'. 

Looking at it again, and if there is user need, we can keep/rephrase codes to cover:
- `awardStatus`: awarded, but still in standstill (by slightly rephrasing 1.1 'pending' from `awardStatus`) 
- `contractStatus`: signed, but not yet in force (by rephrasing 1.1 'pending' from `contractStatus` to correspond to this point and not to the next point)
- `contractStatus`: in force, but not yet terminated (by sticking closer to the 1.1 description of 'active') 

I guess the main point is: what do we want to track through `*status` and what is enough to have taken care of by another field (e.g. `dateSignature`, `dateAward`)? (Looking at it now, I'm actually more inclined to track it all in *status, especially since that is the status quo.)

schema/codelists/tenderStatus.csv
JachymHercher: Done in https://github.com/open-contracting/standard/pull/1509/commits/8aee47456cc68f8907f997c7801245201f29b81e. (The sentence before the parentheses is "The contracting process failed." is extremely broad, which is why I originally put the parentheses at the end, but you have large breadth also in the other stages, so it probably isn't a good reason.)

docs/history/changelog.md
jpmckinney: I've updated my note in CRM-6532 to make this part of the consultation (with a link to the relevant comment in #1496 . For now, we'll leave it as decided in #395.

schema/codelists/tenderStatus.csv
jpmckinney: I think "contracting documents (for example, procurement documents)" is fine, and can be changed in all necessary locations. (It seems some governments even use the term "procurement contracting documents".)

Yes, let's go with "The contracting documents are available to potential suppliers."

Does this mean 'planned' will change to "... but the contracting documents are not yet available to potential suppliers." ?

JachymHercher: Done in the last few commits. (I left "procurement documents" in som examples or exampe-like guidance. I did not do a massive review of all the uses of "publish".)

duncandewhurst:
@jpmckinney based on https://github.com/open-contracting/standard/issues/1160#issuecomment-1598012872, I think that the next step for this PR is for you to review the changes since your last review and for me to address any feedback.

jpmckinney:
I've honestly lost track of the status of this PR. From your reading, does it resolve the part of #1160 that it intends to address?

I see that I made commits to make it mergeable. And I see that Jáchym had accepted my earlier suggestions, and then made other changes (I assume in response to other suggestions).

If that's all correct, then, yes, I should read it once more.

docs/history/changelog.md
```diff
@@ -144,7 +144,15 @@ Per the [normative and non-normative content and changes policy](https://docs.go
...
+    * Deprecated: 'pending' in `awardStatus` and `contractStatus`. Deprecated because the code (which would follow 'active') does not provide any additional information compared to `awardDate` and `dateSigned` and, in the case of `contractStatus` has unclear semantics ("proposed").
+    * Added: 'withdrawn' in `awardStatus` and `contractStatus`, as information can stop being published at any moment also for only individual awards / contracts.
+    * Added: 'complete' in `awardStatus`, as there was no code suggesting that the award successfully happened and was done (the equivalent in `contractStatus` is 'terminated'.)
+    * Added 'unsuccessful' in `contractStatus`, as a failure not caused by the buyer can happen also in the contract stage (e.g. the winning bidder(s) refuses to sign the contract). 
+  * [#1201](https://github.com/open-contracting/standard/pull/1201) Add 'terminatedEarly' and 'terminatedSuccessfully' to the `contractStatus` codelist, to distinguish between successful completion and early termination of the contract.
+
```
jpmckinney: The changelog should be entirely rewritten. These entries were before the decision to deprecate status fields in favor of only having finalStatus fields.

schema/codelists/tenderStatus.csv
```diff
@@ -1,8 +1,8 @@
...
-planned,Planned,"A contracting process is scheduled, but is not yet taking place. Details of the anticipated dates can be provided in the tender block."
-active,Active,A tender process is currently taking place.
-cancelled,Cancelled,The tender process has been cancelled.
-unsuccessful,Unsuccessful,The tender process was unsuccessful.
-complete,Complete,The tender process is complete.
-withdrawn,Withdrawn,No further information on this process is available under this ocid.
+Code,Title,Description,Deprecated,Deprecation note
+planning,Planning,"A future contracting process is being considered. Early information about the process can be provided in the tender section. A process with this status might provide information on early engagement or consultation opportunities, during which the details of a subsequent tender can be shaped.",1.2,"This code has been deprecated, because it is unnecessary since contracting processes are split from planning processes (which do not have any states themselves)."
```
jpmckinney:
```suggestion
planning,Planning,"A future contracting process is being considered. Early information about the process can be provided in the tender section. A process with this status might provide information on early engagement or consultation opportunities, during which the details of a subsequent tender can be shaped.",1.2,"Deprecated as unnecessary, since contracting processes are separate from planning processes (which do not have any states)."
```

schema/codelists/tenderStatus.csv
```diff
@@ -1,8 +1,8 @@
...
+planned,Planned,"The contracting process is scheduled, but contracting documents (for example, planning documents) are not yet available to potential suppliers.",1.2,Deprecated because the same information can be determined by comparing `tender/datePublished` (new in 1.2) to `date` (or its absence)
```
jpmckinney:
```suggestion
planned,Planned,"The contracting process is scheduled, but contracting documents (for example, planning documents) are not yet available to potential suppliers.",1.2,Deprecated because the same information can be determined by the absence of `tender/datePublished` (new in 1.2) or by comparing it to `date`.
```

schema/codelists/contractStatus.csv
```diff
@@ -1,7 +1,5 @@
-Code,Title,Description
-pending,Pending,"This contract has been proposed, but is not yet in force. It might be awaiting signature."
-active,Active,"This contract has been concluded by all the parties, and is now legally in force."
-cancelled,Cancelled,This contract has been cancelled prior to being concluded.
-terminated,Terminated,"This contract was concluded and in force, and has now come to a close. This might be due to the successful completion of the contract, or might be early termination due to some non-completion."
-terminatedEarly,Terminated early,"This contract was concluded and in force, and is terminated early due to some non-completion."
-terminatedSuccessfully,Terminated successfully,"This contract was concluded and in force, and is terminated due to its successful completion."
+Code,Title,Description,Deprecated,Deprecation note
+pending,Pending,"The award was made and the standstill period, if any, has ended. The contract is not yet concluded (for example, not signed by all parties).",1.2,Deprecated because the same information can be determined by the absence of a `contracts.dateSigned`.
```
jpmckinney:
```suggestion
pending,Pending,"The award was made and the standstill period, if any, has ended. The contract is not yet concluded (for example, not signed by all parties).",1.2,Deprecated because the same information can be determined by the absence of `contracts.dateSigned`.
```

schema/codelists/tenderStatus.csv
```diff
@@ -1,8 +1,8 @@
...
+active,Active,"The contracting documents (for example, procurement documents) are available to potential suppliers, and the bid submission deadline has not yet passed.",1.2,Deprecated because the same information can be determined by comparing `tender/datePublished` (new in 1.2) and `tender/tenderPeriod/endDate` to `date`.
+complete,Complete,The bid submission deadline has passed.,1.2,Deprecated because the same information can be determined by comparing `tender/tenderPeriod/endDate` to `date`
```
jpmckinney:
```suggestion
complete,Complete,The bid submission deadline has passed.,1.2,Deprecated because the same information can be determined by comparing `tender/tenderPeriod/endDate` to `date`.
```

schema/codelists/tenderStatus.csv
```diff
@@ -1,8 +1,8 @@
...
+cancelled,Cancelled,"The contracting process was cancelled by the buyer or the procuring entity (for example, because of a change in needs, insufficient funds, or technical or procedural errors) after the contracting documents (for example, procurement documents) were available to potential supppliers, but not later than the bid submission deadline.",1.2,OCDS 1.2 replaces this code with a new field for final states.
```
jpmckinney:
```suggestion
cancelled,Cancelled,"The contracting process was cancelled by the buyer or the procuring entity (for example, because of a change in needs, insufficient funds, or technical or procedural errors) after the contracting documents (for example, procurement documents) were available to potential suppliers, but not later than the bid submission deadline.",1.2,OCDS 1.2 replaces this code with a new field for final states.
```

jpmckinney (commented):
Besides my suggestions, I think only the changelog needs to be updated.

Merging this PR does leave 1.2-dev in an unstable position (fields are deprecated, without their replacements being implemented).

That said, this PR is over a year old, so better to get this first set of changes committed.

docs/history/changelog.md
jpmckinney: Since we haven't yet implemented the replacements for the deprecated fields, we can maybe just leave a simple changelog entry for now, and write something longer when doing the next PR.

schema/codelists/tenderStatus.csv
duncandewhurst: I find the example of planning documents to be a bit confusing here since my understanding is that `tender.status` should only be published as part of a contracting process. Could it instead be replaced with bidding documents (to align with the document type codelist)?

Regarding the deprecation notice, my understanding is that `tender.datePublished` is only to be used retrospectively and that future notice dates should be communicated using `tender.communication.futureNoticeDate` so I'm struggling to understand how comparing `tender/datePublished` to `date` could be used to determine this state.

docs/history/changelog.md
duncandewhurst: Done in https://github.com/open-contracting/standard/pull/1509/commits/5fd7ac0a4dc30535f63a80b87c77620afcb597d8. I also added entries for the other changes in this PR.

duncandewhurst:
> From your reading, does it resolve the part of #1160 that it intends to address?

It looks to address points 1 and 2 from https://github.com/open-contracting/standard/issues/1160#issuecomment-1139793598, plus a few other changes that I've listed in the changelog.

schema/codelists/tenderStatus.csv
jpmckinney: I'm not sure why that was used in e33c21c0b255bc337879c92a63028f03f00435aa but we can change to "procurement documents" like in the 'active' code.

For deprecation, we can just leave it as the absence of `tender.datePublished`. There are certainly many dates that would mean the tender is active.

jpmckinney (commented):
I just noticed that the deprecation notices inconsistently use dot and slash notation. I think our style guide is to use dot notation here.

schema/codelists/tenderStatus.csv
duncandewhurst: Sounds good. I've made those changes in f6217cc57e6c6ab6121d8f681d29366e1bd86ced

jpmckinney: 34ee1350796350408d617c0bc89739949fd1f78e

schema/codelists/contractStatus.csv
```diff
@@ -1,7 +1,5 @@
...
-cancelled,Cancelled,This contract has been cancelled prior to being concluded.
-terminated,Terminated,"This contract was concluded and in force, and has now come to a close. This might be due to the successful completion of the contract, or might be early termination due to some non-completion."
-terminatedEarly,Terminated early,"This contract was concluded and in force, and is terminated early due to some non-completion."
-terminatedSuccessfully,Terminated successfully,"This contract was concluded and in force, and is terminated due to its successful completion."
+Code,Title,Description,Deprecated,Deprecation note
+pending,Pending,"The award was made and the standstill period, if any, has ended. The contract is not yet concluded (for example, not signed by all parties).",1.2,Deprecated because the same information can be determined by the absence of `contracts.dateSigned`.
+active,Active,"The contract was concluded (for example, signed by all parties). The contract is not yet terminated.",1.2,Deprecated because the same information can be determined by comparing `contracts.period` to `date`.
```
jpmckinney:
```suggestion
active,Active,"The contract was concluded (for example, signed by all parties). The contract is not yet terminated.",1.2,Deprecated because the same information can be determined by the presence of `contracts.dateSigned` and by comparing `contracts.period` to `date`.
```

schema/codelists/tenderStatus.csv
```diff
@@ -1,8 +1,8 @@
...
-withdrawn,Withdrawn,No further information on this process is available under this ocid.
+Code,Title,Description,Deprecated,Deprecation note
+planning,Planning,"A future contracting process is being considered. Early information about the process can be provided in the tender section. A process with this status might provide information on early engagement or consultation opportunities, during which the details of a subsequent tender can be shaped.",1.2,"Deprecated as unnecessary, since contracting processes are separate from planning processes (which do not have any states)."
+planned,Planned,"The contracting process is scheduled, but contracting documents (for example, procurement documents) are not yet available to potential suppliers.",1.2,Deprecated because the same information can be determined by the absence of `tender.datePublished` (new in 1.2).
+active,Active,"The contracting documents (for example, procurement documents) are available to potential suppliers, and the bid submission deadline has not yet passed.",1.2,Deprecated because the same information can be determined by comparing `tender.datePublished` (new in 1.2) and `tender/tenderPeriod/endDate` to `date`.
```
jpmckinney:
```suggestion
active,Active,"The contracting documents (for example, procurement documents) are available to potential suppliers, and the bid submission deadline has not yet passed.",1.2,Deprecated because the same information can be determined by the presence of `tender.datePublished` (new in 1.2) and by comparing `tender/tenderPeriod/endDate` to `date`.
```

schema/codelists/tenderStatus.csv
```diff
@@ -1,8 +1,8 @@
...
+complete,Complete,The bid submission deadline has passed.,1.2,Deprecated because the same information can be determined by comparing `tender.tenderPeriod.endDate` to `date`.
```
jpmckinney:
```suggestion
complete,Complete,The bid submission deadline has passed.,1.2,Deprecated because the same information can be determined by the presence of `tender.datePublished` (new in 1.2) and by comparing `tender.tenderPeriod.endDate` to `date`.
```

schema/codelists/tenderStatus.csv
```diff
@@ -1,8 +1,8 @@
...
+cancelled,Cancelled,"The contracting process was cancelled by the buyer or the procuring entity (for example, because of a change in needs, insufficient funds, or technical or procedural errors) after the contracting documents (for example, procurement documents) were available to potential suppliers, but not later than the bid submission deadline.",1.2,OCDS 1.2 replaces this code with a new field for final states.
+unsuccessful,Unsuccessful,"The contracting process failed (for example, no bids were received, all bids were withdrawn by the bidders, all bids were rejected by the buyer or the procuring entity) after the contracting documents (for example, procurement documents) were available to potential suppliers, but not later than the bid submission deadline.",1.2,OCDS 1.2 replaces this code with a new field for final states.
+withdrawn,Withdrawn,No further information on the contracting process is available under this ocid.,1.2,OCDS 1.2 replaces this code with a new field for final states.
```
jpmckinney:
```suggestion
withdrawn,Withdrawn,No further information on the contracting process is available under this ocid.,1.2,This code is deprecated, because it describes the status of the publication, not of the contracting process.
```

jpmckinney (commented):
If all suggestions accepted, we can merge.

I mostly made deprecation notes more consistent.

PR #1512, opened 2022-05-17
Change wording to multi-stage

### LINKED ISSUE #1511: Change "multiple stage procedures" to "multi-stage procedures"
In line with 4) of https://github.com/open-contracting/standard/issues/1301#issuecomment-999083865, we should not speak about "multiple stage procedures", but "multiple-stage procedures". 

Also, rather than using "multiple-stage" (which I unwittingly introduced myself), we should stick to "multi-stage". The latter is used more often in OCDS as well as having more results on google.

JachymHercher:
Closes #1511


jpmckinney (approved):

jpmckinney:
I fixed two occurrences in 6c9511f.

PR #1167, opened 2021-01-14
Improve the field description of the Budget object

ColinMaudry:
`Budget`,` Budget.id`, `Budget.description`, `Budget.amount`

Closes #701

PR #1513, opened 2022-05-18
Review use of "contracting process", replace with "contracting (or planning) process"

### LINKED ISSUE #1409: Review usage of "contracting process" to account for planning process
@JachymHercher in #866:

From https://github.com/open-contracting/standard/issues/866#issuecomment-778792793: 
> 3. I have replaced "contracting process" with "contracting and/or planning process" only in a few prominent places where I thought it would be useful. The rest I propose doing once the "Getting Started" section is reviewed.

> The getting section is reviewed, so we can do this now. However, on second thought, we should probably also wait for a moment when there are no - or very few - unmerged PRs, because since this change will go through all parts of the standard, it would otherwise cause plenty of conflicts?

> Since there are almost 600 results for "contracting process", before going through them, two questions:
&nbsp;&nbsp;&nbsp;&nbsp; a) are we sure we want to replace them? I would, as just defining somewhere that "a contracting process can be a contracting or planning process" would be cyclical and ignored anyway. 
&nbsp;&nbsp;&nbsp;&nbsp; b) what should we replace them with? I'd go for "contracting (or planning) process", "contracting or planning process" or "contracting/planning process"; in this order of preference (I think the parentheses nicely hide the less important process and make the text more easily readable, because one doesn't get sidetracked by the "or"; backslashes are a bit hard to read). Just calling it "process" doesn't seem clear enough. 

> Detail: there is a repeating formulation "within the scope of the contracting process". I would leave out "the scope of", it seems unnecessary.

### LINKED ISSUE #1517: Align wording of `budget.project` and `budget.projectID`
The description of `budget.project` 

> The name of the project through which the eventual contracts will be funded (if applicable). Some organizations maintain a registry of projects, and the data should use the name by which the project is known in that registry.

The description of `budget.projectID` 

> An external identifier for the project that the future contracting process will be part of, or will be funded via (if applicable). Some organizations maintain a registry of projects, and the data should use the identifier from the relevant registry of projects. Although an integer is allowed, it is recommended to use a string.

Shouldn't the descriptions, especially the first sentence, be more similar? I would expect them to be practically identical, just the former with a textual name and the latter with an identifier (for example, the first sentence of the project could be changed to "The identfier of the project through which the eventual contracts will be funded."). 

(I would also remove the "if applicable")

JachymHercher:
Closes #1409

Closes #1517


schema/codelists/documentType.csv
```diff
@@ -1,15 +1,15 @@
...
-tender,exclusionGrounds,Exclusion grounds,"Documentation that describes the criteria regarding the situation of a tenderer that can lead to its exclusion from the contracting process. For example: criminal convictions, bankruptcy, presence on a blacklist or failure to pay taxes.",,
-tender,selectionCriteria,Selection criteria,Documentation that describes the minimum requirements for potential suppliers to participate in the contracting process.,,
+tender,exclusionGrounds,Exclusion grounds,"Documentation that describes the criteria regarding the situation of a tenderer that can lead to its exclusion from the future contracting process. For example: criminal convictions, bankruptcy, presence on a blacklist or failure to pay taxes.",,
+tender,selectionCriteria,Selection criteria,Documentation that describes the minimum requirements for potential suppliers to participate in the future contracting process.,,
```
jpmckinney: I don't think these are about the future but the present process? (Section is "tender"). That said, I know that these can also be published in a PIN (and so maybe we should add "planning" to Section, but in another PR). With that in mind, should we change to "contracting activity" to make it less tied to OCDS' particular way of splitting processes?

schema/codelists/documentType.csv
```diff
@@ -23,20 +23,20 @@ implementation,physicalProgressReport,Physical progress reports,"Documentation o
...
 tender,clarifications,Clarifications to bidders questions,Documentation that provides replies to issues raised in pre-bid conferences or an enquiry processes.,,
 tender,shortlistedFirms,Shortlisted firms,Documentation providing information on shortlisted firms. Structured versions of this information can be provided using the bids extension.,,
-"planning, implementation",environmentalImpact,Environmental impact,"Documentation of assessments of the environmental impacts (e.g. impacts on flora, fauna & woodlands, areas of natural beauty, carbon emissions etc.) and mitigation measures (e.g. pollution control, low carbon solutions, sustainable timber etc.) for the contracting process.",,
+"planning, implementation",environmentalImpact,Environmental impact,"Documentation of assessments of the environmental impacts (e.g. impacts on flora, fauna & woodlands, areas of natural beauty, carbon emissions etc.) and mitigation measures (e.g. pollution control, low carbon solutions, sustainable timber etc.) for the future contracting process.",,
```
jpmckinney: So here the Section is "planning, implementation", because you can measure environmental impact both before and after the intervention. Would "contracting activity" work?

schema/codelists/documentType.csv
```diff
@@ -1,15 +1,15 @@
...
 planning,plannedProcurementNotice,Notice of planned procurement,"Notice published by the buyer or procuring entity regarding their plans for future procurement. This notice is called a notice of planned procurement (for example, in the WTO's General Procurement Agreement), a future opportunity notice, a planning notice, a prior information notice or a periodic indicative notice (for example, in the EU). In some cases, the buyer or procuring entity can use the notice of planned procurement as a tender notice.",,
-planning,marketEngagementNotice,Market engagement notice,"A notice published by the buyer or procuring entity to consult the market with the aim to prepare the contracting process. This type of activity is also called market consultation, early engagement, pre-procurement dialogue and information gathering.",,
+planning,marketEngagementNotice,Market engagement notice,"A notice published by the buyer or procuring entity to consult the market with the aim to prepare the future contracting process. This type of activity is also called market consultation, early engagement, pre-procurement dialogue and information gathering.",,
```
jpmckinney: Depending on whether we prefer "contracting activity" in all cases or only some cases, if we do keep "future contracting process", the article should be "a" not "the", since we don't know if it exists yet (and there could even be multiple).

This applies to all other occurrences in this file (and probably other files).

schema/codelists/partyRole.csv
```diff
@@ -1,21 +1,21 @@
...
+enquirer,Enquirer,An organization that has made an enquiry during the enquiry phase of a contracting (or planning) process.,,
 payer,Payer,An organization making a payment from a transaction.,,
 payee,Payee,An organization in receipt of a payment from a transaction.,,
 reviewBody,Review body,An organization responsible for the review of this procurement process. This organization often has a role in any challenges made to the contract award.,,
 interestedParty,Interested party,"An organization that has expressed an interest in the contracting process: for example, by purchasing tender documents or submitting clarification questions.",,
-contractImplementationManager,Contract implementation manager,The organization responsible for managing the implementation of the contract on behalf of the buyer. (This may be different from the procuring entity that manages the contracting process.),,
+contractImplementationManager,Contract implementation manager,The organization responsible for managing the implementation of the contract on behalf of the buyer. (This may be different from the procuring entity that manages the contracting (or planning) process.),,
```
jpmckinney: I'm not sure that a planning process would ever have a contractImplementationManager.

schema/codelists/partyRole.csv
```diff
@@ -1,21 +1,21 @@
...
-procuringEntity,Procuring entity,"The organization managing the contracting process. If an organization is both a buyer and a procuring entity (as can be the case in simple contracting processes), its roles should include 'buyer', but not 'procuringEntity'.",,
+procuringEntity,Procuring entity,"The organization managing the contracting (or planning) process. If an organization is both a buyer and a procuring entity (as can be the case in simple contracting processes), its roles should include 'buyer', but not 'procuringEntity'.",,
 supplier,Supplier,An organization with which a buyer or a procuring entity decided to conclude a contract.,,
 tenderer,Tenderer,An organization that submitted a bid.,,
 funder,Funder,The funder is an organization providing money or finance for this contracting process.,,1.2
-enquirer,Enquirer,An organization that has made an enquiry during the enquiry phase of a contracting process.,,
+enquirer,Enquirer,An organization that has made an enquiry during the enquiry phase of a contracting (or planning) process.,,
```
jpmckinney: Do planning processes have an enquiry phase?

schema/codelists/partyRole.csv
```diff
@@ -1,21 +1,21 @@
...
 lender,Lender,"An institution that finances the contracting process through debt financing, like traditional debt or project bonds. This role typically relates to financing a public-private partnership project.",https://ppp-certification.com/ppp-certification-guide/glossary,
 equityInvestor,Equity investor,"An investor that finances the contracting process through equity financing: that is, capital in exchange for a share of ownership. This role typically relates to financing a public-private partnership project.",https://ppp-certification.com/ppp-certification-guide/glossary,
-notifiedPotentialSupplier,Notified potential supplier,"A potential supplier that has been notified about the contracting process: for example, through a non-exclusive invitation to bid.",,
+notifiedPotentialSupplier,Notified potential supplier,"A potential supplier that has been notified about the contracting (or planning) process: for example, through a non-exclusive invitation to bid.",,
```
jpmckinney, line 21: Are potential suppliers notified about planning processes?

schema/codelists/extendedProcurementCategory.csv
```diff
@@ -1,5 +1,5 @@
 Code,Title,Description
-goods,Goods and supplies,This contracting process involves physical or electronic goods or supplies.
-works,Works,"This contracting process involves construction, repair, rehabilitation, demolition, restoration or maintenance of some asset or infrastructure."
-services,Services,"This contracting process involves professional services of some form, generally contracted for on the basis of measurable outputs or deliverables. When the consultingServices code is also available or in use for data from a particular dataset, the service code must only be used for non-consulting services."
-consultingServices,Consulting services,This contracting process involves professional services provided on a consultancy basis.
+goods,Goods and supplies,This contracting (or planning) process involves physical or electronic goods or supplies.
```
jpmckinney: For this codelist, the planning process doesn't involve goods/construction/etc. It's the future contracting process that will involve those.

What if we instead just change "This" to "The"? That way, from perspective of a planning process, it refers to the (future) contracting process. From the perspective of the contracting process, it refers to itself.

Going back to my comments in `documentType.csv`, maybe we could just state somewhere:

> When the description of a field or code refers to a contracting process:
>
> * If the release is for a contracting process, "contracting process" refers to itself.
> * If the release is for a planning process, "contracting process" refers to a future contracting process that will be related to the planning process.

schema/codelists/procurementCategory.csv
```diff
@@ -1,4 +1,4 @@
 Code,Title,Description
-goods,Goods and supplies,The primary object of this contracting process involves physical or digital goods or supplies.
-works,Works,"The primary object of this contracting process involves construction, repair, rehabilitation, demolition, restoration or maintenance of some asset or infrastructure."
-services,Services,"The primary object of this contracting process involves professional services of some form, generally contracted for on the basis of measurable outputs or deliverables."
+goods,Goods and supplies,The primary object of this contracting (or planning) process involves physical or digital goods or supplies.
```
jpmckinney: Same concerns as for extendedProcurementCategory.

schema/release-schema.json
```diff
@@ -2580,7 +2580,7 @@
           "openCodelist": true
         },
         "title": {
-          "description": "The title of the related process, where referencing an open contracting process, this field should match the tender/title field in the related process.",
+          "description": "The title of the related process, where referencing an open contracting (or planning) process, this field should match the tender/title field in the related process.",
```
jpmckinney: We can omit "open" here https://github.com/open-contracting/standard/issues/1409#issuecomment-1130229088

schema/release-schema.json
```diff
@@ -2557,7 +2557,7 @@
       }
     },
     "RelatedProcess": {
-      "description": "A reference to a related contracting process: generally one preceding or following on from the current process.",
+      "description": "A reference to a related contracting (or planning) process: generally one preceding or following on from the current process.",
```
jpmckinney, line 2560: FYI, this field can be used to refer to non-OCDS data (e.g. when the UK transitioned, they referred to TED notices). That said, I think it's fine to use "contracting (or planning) process", since the interpretation should be of any corresponding concept in TED, etc.

schema/release-schema.json
```diff
@@ -1224,28 +1224,28 @@
...
+          "description": "The name of the project through which the future contracting process will be funded (if applicable). Some organizations maintain a registry of projects, and the data should use the name by which the project is known in that registry.",
           "type": [
             "string",
             "null"
           ]
         },
         "projectID": {
           "title": "Project identifier",
-          "description": "An external identifier for the project that this contracting process forms part of, or is funded via (if applicable). Some organizations maintain a registry of projects, and the data should use the identifier from the relevant registry of projects. Although an integer is allowed, it is recommended to use a string.",
+          "description": "An external identifier for the project that the future contracting process will be part of, or will be funded via (if applicable). Some organizations maintain a registry of projects, and the data should use the identifier from the relevant registry of projects. Although an integer is allowed, it is recommended to use a string.",
```
jpmckinney: To match prepositions of `project`. As done previously elsewhere, we can also remove "(if applicable)".

```suggestion
          "description": "An external identifier for the project of which the future contracting process will be part, or through which it will be funded. Some organizations maintain a registry of projects, and the data should use the identifier from the relevant registry of projects. Although an integer is allowed, it is recommended to use a string.",
```

schema/release-schema.json
```diff
@@ -1224,28 +1224,28 @@
...
+          "description": "The value reserved in the budget to be used in a future contracting process. A negative value indicates anticipated income to the budget as a result of this contracting process, rather than expenditure. Where the budget is drawn from multiple sources or extends over multiple years, the budget breakdown extension can be used. This field should not be used to report the total value of the budget line funding this contracting process.",
           "$ref": "#/definitions/Value"
         },
         "project": {
           "title": "Project title",
-          "description": "The name of the project through which this contracting process is funded (if applicable). Some organizations maintain a registry of projects, and the data should use the name by which the project is known in that registry.",
+          "description": "The name of the project through which the future contracting process will be funded (if applicable). Some organizations maintain a registry of projects, and the data should use the name by which the project is known in that registry.",
```
jpmckinney: Should we do this for `project` and `projectID` to correspond? If so, please also add a changelog entry.

```suggestion
          "description": "The name of the project of which the future contracting process will be part, or through which it will be funded. Some organizations maintain a registry of projects, and the data should use the name by which the project is known in that registry.",
```

schema/release-schema.json
```diff
@@ -1224,28 +1224,28 @@
         },
         "description": {
           "title": "Budget Source",
-          "description": "A short free-text description of the budget allocation for this contracting process. This may be used to provide human-readable information on the budget category allocated to this contracting process, and/or, information about the nature and source of the allocation (e.g. conditional, confirmed; any official authorizations given to the budget allocation).",
+          "description": "A short free-text description of the budget allocation which will be used in a future contracting process. This may be used to provide human-readable information on the budget category allocated to the future contracts, and/or, information about the nature and source of the allocation (e.g. conditional, confirmed; any official authorizations given to the budget allocation).",
```
jpmckinney: To match wording of `amount`.

```suggestion
          "description": "A short free-text description of the budget to be used in a future contracting process. This may be used to provide human-readable information on: the budget category for the future contracts and/or the nature and source of the budget allocation (e.g. conditional, confirmed, or any official authorizations given to the budget allocation).",
```

schema/release-schema.json
```diff
@@ -1211,11 +1211,11 @@
     "Budget": {
       "type": "object",
       "title": "Budget information",
-      "description": "This section contains basic information about the budget estimated for, or allocated to, this contracting process at the present time. Further documentation and data about how budgets have been allocated to a contracting process should be published outside of OCDS data, according to the best available standards.",
+      "description": "This section contains basic information about the budget that will be used in a future contracting process. Further documentation and data about how budgets have been allocated should be published outside of OCDS data, according to the best available standards.",
```
```suggestion
      "description": "This section contains basic information about the budget to be used in a future contracting process. Further documentation and data about how budgets have been allocated should be published outside of OCDS data, according to the best available standards.",
```

schema/release-schema.json
```diff
@@ -1211,11 +1211,11 @@
...
       "properties": {
         "id": {
           "title": "ID",
-          "description": "An identifier for the budget line item which provides funds for this contracting process. This identifier should be possible to cross-reference against formal budget documents.",
+          "description": "An identifier for the budget line item which will be used in a future contracting process. This identifier should be possible to cross-reference against formal budget documents.",
```
jpmckinney: Good god, we really didn't use consistent wording (budget, budget line item, budget allocation).

```suggestion
          "description": "An identifier of the budget to be used in a future contracting process. This identifier should be possible to cross-reference against formal budget documents.",
```

schema/release-schema.json
```diff
@@ -1254,7 +1254,7 @@
         },
         "uri": {
           "title": "Linked budget information",
-          "description": "A URI pointing directly to a machine-readable record about the budget line-item or line-items that fund this contracting process. Information can be provided in a range of formats, including using IATI, the Open Fiscal Data Standard or any other standard which provides structured data on budget sources. Human readable documents can be included using the planning.documents block.",
+          "description": "A URI pointing directly to a machine-readable record about the budget line-item or line-items that will be used in the future contracting process. Information can be provided in a range of formats, including using IATI, the Open Fiscal Data Standard or any other standard which provides structured data on budget sources. Human readable documents can be included using the planning.documents block.",
```
jpmckinney: The description of `Budget` had been changed to "according to the best available standards" in #1167. We forgot to remove the specific examples here.

```suggestion
          "description": "The URL of a machine-readable resource about the budget to be used in the future contracting process. For human-readable documents, see the `planning.documents` array.",
```

schema/release-schema.json
```diff
@@ -160,7 +160,7 @@
         },
         "budget": {
           "title": "Budget",
-          "description": "Details of the budget that funds the planned contracting process.",
+          "description": "Details of the budget that will be used in a future contracting process.",
```
```suggestion
          "description": "Details of the budget to be used in a future contracting process.",
```

schema/release-schema.json
```diff
@@ -13,14 +13,14 @@
...
       "type": "string",
       "minLength": 1,
       "omitWhenMerged": true
     },
     "date": {
       "title": "Release Date",
-      "description": "The date on which the information contained in the release was first recorded in, or published by, any system. For a compiled release, the maximum `date` among the individual releases used to create the compiled release. The release date should be unique within the scope of the contracting process identified by the Open Contracting ID (ocid), for a given version of OCDS.",
+      "description": "The date on which the information contained in the release was first recorded in, or published by, any system. For a compiled release, the maximum `date` among the individual releases used to create the compiled release. The release date should be unique within the scope of the (contracting or planning) process, for a given version of OCDS.",
```
jpmckinney, line 23: Why not this? Same question for `id`.

```suggestion
      "description": "The date on which the information contained in the release was first recorded in, or published by, any system. For a compiled release, the maximum `date` among the individual releases used to create the compiled release. The release date should be unique within the scope of the contracting (or planning) process, for a given version of OCDS.",
```

jpmckinney (commented):
@JachymHercher I've reviewed the schema section first. Once we resolve those, I think it'll be easier to review the rest.

schema/codelists/documentType.csv
JachymHercher: Sorry, this was my mistake. (I got mentally sidetracked along the lines of "participating in a process", "requesting to participate in a process", "requesting to participate in a future process?" etc., see https://github.com/open-contracting/standard/pull/1464#discussion_r774787527.) 

Replacing process with activities - I rather wouldn't. First, I think we should try to stick to the defined term, if we can. Second, I wouldn't do anything special for these EU PINs, because it's not all the EU PINs, but just for the "PINs used as call for competition", which the eForms classifies under Competition forms, so they wouldn't start a planning process but a contracting process. (Regrettably, this information is also in "PIN used to shorten time limits", which is classified as a planning notice, but the scenario is quite narrow, jurisdiction-specific and - I think - not hugely used, so I wouldn't change our wording just because of it.)

schema/codelists/documentType.csv
JachymHercher: I would try to avoid using activities, see above. However, in this issue, I think we have an easy way out: somewhat similarly to the discussion on budgets, the environmental impact isn't of the of the process per se, but rather of the actual purchase (i.e. items). So I would replace the "for contracting process" with "of the works, goods or services".

Proposed in https://github.com/open-contracting/standard/pull/1513/commits/be8c64f5a9bb0998c0e4d93000023f74efec0fe6. Related issue opened in https://github.com/open-contracting/standard/issues/1521#issue-1242889108.

schema/codelists/partyRole.csv
JachymHercher: Not sure. I don't really have an "enquiry phase" in my mental model of contracting (which is quite possibly my fault because questions don't have fields in eForms and they are not really part of the Directives, so I don't know much about them).

For me, the only thing linked to questions is a deadline (typically a few weeks before bid submission, with the reasoning being that tenderers need time to reflect the answers in their bids) and then the modalities of how/where to send them. I don't really think of a deadline only after which you can _start_ asking questions. 

Obviously, you will be asking different / more concrete questions after the procurement documents have been published (contracting process, e.g. "could you clarify sentence XYZ") than before (planning process, e.g. "could you clarify vague paragraph XYZ in the PIN", "are you aware of solution X that would solve your need but might end up not being applicable in your future contracting process because of thing ABC that you mention in your planning process" or perhaps even "do you have any news on when the contracting process will be launched"?), but for me both seem similar enough to group them under one code.

In this PR, I would leave it as is, and then deal with it in more detail in a follow-up issue: https://github.com/open-contracting/standard/issues/1522

schema/codelists/partyRole.csv
JachymHercher: Sure? "Dear company that participated in previous computer procedure / company that responsed to the prior market consultation, here is a link to our PIN about the procedures we are planning to run next year. Get ready & happy for your feedback."

schema/codelists/documentType.csv
JachymHercher: Done in https://github.com/open-contracting/standard/pull/1513/commits/93b050c385e810bad6fb03b1ccc04fd19a8276ce

schema/codelists/documentType.csv
JachymHercher: So far, I'm assuming we keep "future contracting process", so I've changed the article to indefinite. Done in https://github.com/open-contracting/standard/pull/1513/commits/37b97510f2b049533f6e278a22225ee9987e1c3a

schema/codelists/partyRole.csv
JachymHercher: Ok, removed in https://github.com/open-contracting/standard/pull/1513/commits/16d72e242f12f8642c19175d2c97bc3942d2dbb2

(My wording doesn't require there to be a contractImplementationManager (CIM) in the planning process - I was rather thinking that it's probably true/consistent to say that if there is a CIM in a contracting process, then the CIM might be different from the procuring entity that manages the contracting (_and/or_ planning) process. However, these are typically the same; instead of "manages" we might need to say "manages"/"managed" etc. - it doesn't matter. Removed :). )

schema/codelists/extendedProcurementCategory.csv
JachymHercher: > For this codelist, the planning process doesn't involve goods/construction/etc. It's the future contracting process that will involve those.

Hm, for me already the planning process does? I guess I understand "involves" as "is about", so I read the codes as "process is about this stuff we wan't to buy". A contracting process then may end up with a contract with might end up with actual implementation, while a planning process may just end up with a contracting process, but that seems fine by me. 

____ 

> What if we instead just change "This" to "The"? That way, from perspective of a planning process, it refers to the (future) contracting process. From the perspective of the contracting process, it refers to itself.

I don't have a big issues with this, but I don't see the need that much either, see above. Also, even if it is "the contracting process", there is still the issue of "involves" vs. "will involve". 

(We are also inconsistent in the use of this/the for contracting/planning processes, the alignment of which could be another issue.) 

___ 

> When the description of a field or code refers to a contracting process:

If the release is for a contracting process, "contracting process" refers to itself.
If the release is for a planning process, "contracting process" refers to a future contracting process that will be related to the planning process.

Same as above - not hugely against, but don't have a strong urge to do it, either. We should still make sure it makes sense in individual fields, because people are unlikely to remember a "general" disclaimer as such.

schema/release-schema.json
JachymHercher: I was planning to do it later within https://github.com/open-contracting/standard/issues/1517, but we can do it here as well. Done in https://github.com/open-contracting/standard/pull/1513/commits/756b5202b2cea363b576b7618d186e281d01b554.

schema/release-schema.json
JachymHercher: Done in https://github.com/open-contracting/standard/commit/756b5202b2cea363b576b7618d186e281d01b554

schema/release-schema.json
JachymHercher: For me, the same reason as in other "ambiguous" cases (point 4 of https://github.com/open-contracting/standard/issues/1409#issuecomment-1129735868). I feel a bit uncertain about both whether the ambiguity in general exists and whether we're solving it with the "(contracting or planning)", so I was glad it was fine by you, but if there is some nuance here I'm quite likely missing it.

schema/codelists/procurementCategory.csv
JachymHercher: same as above

schema/codelists/documentType.csv
```diff
@@ -23,20 +23,20 @@ implementation,physicalProgressReport,Physical progress reports,"Documentation o
...
+planning,marketStudies,Market studies,Documentation of any market studies that took place as part of the planning for a future contracting process.,,
 tender,clarifications,Clarifications to bidders questions,Documentation that provides replies to issues raised in pre-bid conferences or an enquiry processes.,,
 tender,shortlistedFirms,Shortlisted firms,Documentation providing information on shortlisted firms. Structured versions of this information can be provided using the bids extension.,,
-"planning, implementation",environmentalImpact,Environmental impact,"Documentation of assessments of the environmental impacts (e.g. impacts on flora, fauna & woodlands, areas of natural beauty, carbon emissions etc.) and mitigation measures (e.g. pollution control, low carbon solutions, sustainable timber etc.) for the contracting process.",,
+"planning, implementation",environmentalImpact,Environmental impact,"Documentation of assessments of the environmental impacts (e.g. impacts on flora, fauna & woodlands, areas of natural beauty, carbon emissions etc.) and mitigation measures (e.g. pollution control, low carbon solutions, sustainable timber etc.) of the works, goods or services.",,
```
jpmckinney, line 29: Elsewhere we use this ordering:

```suggestion
"planning, implementation",environmentalImpact,Environmental impact,"Documentation of assessments of the environmental impacts (e.g. impacts on flora, fauna & woodlands, areas of natural beauty, carbon emissions etc.) and mitigation measures (e.g. pollution control, low carbon solutions, sustainable timber etc.) of the goods, works or services.",,
```

schema/codelists/partyRole.csv
jpmckinney: Yes, it was a true statement, but it kind of violates the Gricean maxim of quantity :)

> Do not make your contribution more informative than is required

https://psychology.fandom.com/wiki/Gricean_maxims#Maxim_of_Quality

schema/codelists/partyRole.csv
jpmckinney: I prefer to not include the change in this PR. If needed, we can do the change in #1522.

schema/codelists/partyRole.csv
jpmckinney: Sounds good. I guess they are "potential" in terms of the eventual contracting process.

schema/codelists/extendedProcurementCategory.csv
jpmckinney: We can change to "is about" (or maybe better "is for"?), and keep the "(or planning)".

While we're at it, let's change to "The", which is more common.

The primary meaning of "Involve" is "include (something) as a necessary part or result" – but no works, etc. result from a planning process.

schema/release-schema.json
jpmckinney: Ah, yes, I forgot.

schema/codelists/documentType.csv
jpmckinney: We can decide in #1521

schema/codelists/partyRole.csv
JachymHercher: Ok. Removed in https://github.com/open-contracting/standard/pull/1513/commits/c5d811a90b71818b6ee980d521154e03dae91a18

schema/codelists/extendedProcurementCategory.csv
JachymHercher: I've added "is for" in https://github.com/open-contracting/standard/pull/1513/commits/de94047a0a4980c4699a238553592a669eb0a424 and "the instead of this" in https://github.com/open-contracting/standard/pull/1513/commits/da953c4ab1e177015914dfda9f39c107fe09ec45. 

Also, I've opened https://github.com/open-contracting/standard/issues/1525.

schema/codelists/procurementCategory.csv
JachymHercher: I've replaced "involve" with "be", e.g. "The primary object are physical or digital goods." Sounds fine by me? 

I've also opened https://github.com/open-contracting/standard/issues/1525.

docs/primer/releases_and_records.md
```diff
@@ -23,20 +23,20 @@ Open Contracting Data Standard (OCDS) data is published and updated over time us
 
 The only thing that can be truly called “OCDS data” is JSON data that validates against the OCDS schema. There are two schemas for OCDS data: “releases” and “records.”
 
-OCDS publishers are encouraged to use releases and records to publish data in near real-time and to provide a change history. Tracking change over time is important to users of contracting data because there can be many changes in the life of a contracting process. Knowing how and when certain attributes have changed can help users to understand the process, identify inefficiencies, and spot red flags for corruption.
+OCDS publishers are encouraged to use releases and records to publish data in near real-time and to provide a change history. Tracking change over time is important to users of contracting data because there can be many changes in the life of a contracting (or planning) process. Knowing how and when certain attributes have changed can help users to understand the process, identify inefficiencies, and spot red flags for corruption.
```
jpmckinney: We don't define planning process in the primer. Should we just leave these as-is?

jpmckinney (commented):
I assume we want to change this text on schema/reference (maybe using "future" - though I can't remember if there is an existing issue recently created):

> The planning section can be used to describe the background to a contracting process. This can include details of the budget from which funds are drawn, or related projects for this contracting process. Background documents such as a needs assessment, feasibility study and project plan can also be included in this section.

We also need to update:

* `ocid` and `releases` in `record-package-schema.json`
* `ocid` in `release-schema.json`

docs/primer/releases_and_records.md
JachymHercher: I've updated https://standard.open-contracting.org/staging/1.2-dev/en/schema/reference/#planning (mainly reused the description of `planning`), removed the "or planning" in the primer and added "or planning" in `releases` in `record-package-schema.json` in https://github.com/open-contracting/standard/pull/1513/commits/bdfd38f4d23ec645ed1d9784d8d143cc3f7e54a8.

I'm not sure `ocid` needs updating - the planning process is in the second sentence there?

schema/codelists/partyRole.csv
jpmckinney: > I'm not sure ocid needs updating - the planning process is in the second sentence there?

Oh, yes, I was lazy and stopped reading after the first sentence :)

docs/primer/releases_and_records.md
jpmckinney: > I'm not sure ocid needs updating - the planning process is in the second sentence there?

Oh, yes, I was lazy and stopped reading after the first sentence :)

PR #1519, opened 2022-05-18
adding amountNet and amountGross to value, updating value definition

### LINKED ISSUE #817: Update Value.amount description to suggest/recommend tax-free amounts

odscjen:
closes #817


docs/history/changelog.md
```diff
@@ -200,6 +201,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
   * [#1189](https://github.com/open-contracting/standard/pull/1189) `Document.url`, to use precise wording and to describe how to indicate a part of a document.
   * [#1229](https://github.com/open-contracting/standard/pull/1229) Standardize the descriptions of `planning.documents`, `tender.documents`, `awards.documents`, `contracts.documents`, and `contracts.implementation.documents`.
   * [#1335](https://github.com/open-contracting/standard/pull/1335) Standardize the descriptions of `planning`, `planning.rationale`, `planning.budget`, `planning.documents`, and `planning.milestones`.
+  * [#1519](https://github.com/open-contracting/standard/pull/1519) `value.amount`, to clarify amount from tender/award/contract notcie document.
```
mrshll1001: Typo on line 204: notcie (notice)

mrshll1001 (changes requested):
One very minor change in the form of a typo, otherwise what I'd expect based on the issue,.

docs/history/changelog.md
odscjen: thanks, corrected that now

mrshll1001 (approved):
This looks fine, typo amended.

jpmckinney (commented):
We can omit "as a number" as this is repetitive with the `type` validation keyword.

I know "The amount as stated in the tender notice / award notice / contract notice document as a number" is the text I suggested in https://github.com/open-contracting/standard/issues/817#issuecomment-542353977, but `Value` objects can appear in many different places, and the numbers might not originate in these specific documents in those cases. The number might also originate from an electronic system, such that there is no "document" per say. 

Unfortunately, as an alternative, I don't think we can use turns of phrase like "original" / "unadjusted" / "unmodified" amount or similar, as we do allow the `amount` of `Value` objects to change over time, and it might just confuse readers as to the semantics.

Maybe something like "The amount as entered into a system or as published in a document."

This PR should also be supported by changes to the Markdown page for the Value object, to explain how the three fields ought to be used/interpreted (e.g. there are three fields so that, when it is unknown whether taxes are included or not, it can nonetheless be entered into `amount` – please be more eloquent than me :)).

The Markdown page should encourage publishers to describe their semantics for `amount`, e.g. a publish that has access to the `amountNet` and `amountGross` can have a policy of entering the `amountNet` in the `amount`. (We can perhaps also encourage this approach of using `amount` for `amountNet` when all are available, for consistency across datasets.)

docs/schema/reference.md
```diff
@@ -655,7 +655,11 @@ For delivery milestones, if there is a time frame for delivery, use `.dueAfterDa
 
 ### Value
 
-Financial values should be published with a currency attached. 
+Financial values should be published with a currency attached. There are three fields available for providing the value amount:
+
+* When it is known that the value includes all taxes use `value/amountGross`;
+* When it is known that the value excludes all taxes use `value/amountNet`;
+* When it is unknown if the value includes all taxes or not use `value/amount`.
```
jpmckinney: The original change doesn't cover all the scenarios discussed in #817. Most of the text was just repeating the semantics of the fields, which doesn't demand repetition.

````suggestion
`currency`, `amountNet` and `amountGross` should be provided, wherever possible.

`amount` is defined as:

```{field-description} ../../build/current_lang/release-schema.json /definitions/Value/properties/amount
```

If both the `amountNet` and the `amountGross` match this definition, enter the `amountNet` as the `amount`.
````

File: docs/history/changelog.md [hunk 3]
```diff
@@ -186,6 +186,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
   * [#1455](https://github.com/open-contracting/standard/pull/1455) `Milestone.dueAfterDate`
   * [#1434](https://github.com/open-contracting/standard/pull/1434) `Milestone.value`
   * [#1490](https://github.com/open-contracting/standard/pull/1490) `SimpleIdentifier`
+  * [#1519](https://github.com/open-contracting/standard/pull/1519) `amountNet` and `amountGross` to `value`
```
```suggestion
  * [#1519](https://github.com/open-contracting/standard/pull/1519) `Value.amountNet` and `Value.amountGross`
```

odscjen:
@jpmckinney I think all the requested changes to this PR were made so just bumping back for your final review

jpmckinney (approved):

jpmckinney:
Not sure why I left this PR open so long. I think I had paused my work on OCDS 1.2 around the time the updates were made.

This new description of `amount` resolves some of the ambiguities that I was maybe unsure about (e.g. how to interpret `amount` depending on which other fields were available).

Issue #105, opened 2014-10-02
I'm not clear on the difference between `itemsContracted` and `deliverables`.


birdsarah:
From @jpmckinney #61

PR #1530, opened 2022-06-22
update a language in schema and docs to consistently refer to good(s)…

### LINKED ISSUE #1521: Use consistently "works, goods or services"
We don't have a consistent expression for "stuff to be bought". We should. 

___ 

`item` is defined as:

> "A good, service, or work to be contracted."

and the `procurementCategory` contains the codes 'works', 'services', and 'goods and supplies'.

___ 

Currently, OCDS uses the following (a few of which might be specific for a reason, but most are probably not): 
- good and services (11 matches)
- goods, works or services (8 matches)
- supplies, services or works (3 matches) 
- goods, works and services (2 matches)
- goods and supplies (in the codename), goods or supplies (in the description) (2 matches)  
- goods and services (5 matches, out of which 4 deprecated)

The GPA uses "goods or services" and, rarely "goods and services". 

The EU directives use "works, supplies or services" and, rarely, "works, supplies and services". (I think the EU mentions works first because they are probably the most important/valuable segment. The reason why the EU can have works while the GPA does not is that works are just a combination of supplies and services.)

___ 

I think we should follow the following principles:
1. With the exception of the definition of 'goods' in `procurementCategory`, we should use just one of "goods" and "supplies". Goods is aligned with the GPA and seems more common, so I would stick to that. 
1. Since "works" are in OCDS, we should mention them along "goods" and "services" when we want to say "stuff to be bought". 
1. "Works" should either be first or last, but not between "goods" and "services" (because it's a bit of a different animal). 
1. "Goods" before "services" seems to be a typical way to phrase things, so we should stick to that.

With that in mind, we should either say "works, goods or services" or "goods, services or works". (With the "or" replaced by "and" depending on the context.) I would go for the former, but largely because I'm very much used to it from the EU (and because I like the argument that it is the most valuable part of procurement.)

odscjen:
…, service(s) and/or work(s)

Where this is referred to in `Item` I've used singular as while an item can be multiple this is covered in `item.quantity` so `Item`, `item.description` and `item.unit` are only ever referring to a single instance of the item so never plural. Otherwise all instances of this phrase are plural with either "and" or "or" depending on context

~~The description for CPV in `classificationScheme.csv` has not been altered as the whole description is lifted from the official scheme url.~~

closes #1521


jpmckinney:
I suggest using the plural in all cases, since an item might be e.g. 100 doses of a drug or something like cleaning services, which cover a range of activities/services that are not typically broken down further.

schema/release-schema.json
```diff
@@ -68,7 +68,7 @@
     },
     "buyer": {
       "title": "Buyer",
-      "description": "The organization aiming to conclude a contract with a supplier or to use the goods, works or services resulting from the contract.",
+      "description": "The organization aiming to conclude a contract with a supplier or to use the goods, services and works resulting from the contract.",
```
duncandewhurst: Was the conjunction changed from 'or' to 'and' deliberately?

schema/release-schema.json
```diff
@@ -823,7 +823,7 @@
         },
         "items": {
           "title": "Items contracted",
-          "description": "The goods, services, and any intangible outcomes in this contract. If the items contracted are identical to the items awarded, this field should be omitted.",
+          "description": "The goods, services and works in this contract. If the items contracted are identical to the items awarded, this field should be omitted.",
```
duncandewhurst: Works and intangible outcomes seem like different concepts, but I don't know the reason that intangible outcomes are mentioned here but not elsewhere. @jpmckinney any ideas?

schema/release-schema.json
```diff
@@ -1369,7 +1369,7 @@
           "description": "(Deprecated in 1.1. Use transaction.payer instead.) The Organization Identifier for the organization from which the funds in this transaction originate. Expressed following the Organizational Identifier standard - consult the documentation and the codelist.",
           "$ref": "#/definitions/Identifier",
           "deprecated": {
-            "description": "This field has been replaced by the `transaction.payer` field to resolve ambiguity arising from 'provider' being interpreted as relating to the goods or services procured rather than the flow of funds between the parties.",
+            "description": "This field has been replaced by the `transaction.payer` field to resolve ambiguity arising from 'provider' being interpreted as relating to the goods, services and works procured rather than the flow of funds between the parties.",
```
duncandewhurst: Was the conjunction changed from 'or' to 'and' deliberately?

schema/release-schema.json
```diff
@@ -1378,7 +1378,7 @@
           "description": "(Deprecated in 1.1. Use transaction.payee instead). The Organization Identifier for the organization which receives the funds in this transaction. Expressed following the Organizational Identifier standard - consult the documentation and the codelist.",
           "$ref": "#/definitions/Identifier",
           "deprecated": {
-            "description": "This field has been replaced by the `transaction.payee` field to resolve ambiguity arising from 'receiver' being interpreted as relating to the goods or services procured rather than the flow of funds between the parties.",
+            "description": "This field has been replaced by the `transaction.payee` field to resolve ambiguity arising from 'receiver' being interpreted as relating to the goods , services and works procured rather than the flow of funds between the parties.",
```
duncandewhurst: Was the conjunction changed from 'or' to 'and' deliberately?

schema/release-schema.json
```diff
@@ -1547,7 +1547,7 @@
     "Item": {
       "title": "Item",
       "type": "object",
-      "description": "A good, service, or work to be contracted.",
+      "description": "A single instance of the goods, services, or works to be contracted.",
```
duncandewhurst: Per https://github.com/open-contracting/standard/pull/1530#issuecomment-1163380962 an item can be plural so I'm not sure 'single instance' is quite the right phrasing. Another option could be to use the wording from https://standard.open-contracting.org/latest/en/schema/reference/#item:

```suggestion
      "description": "A line-item of the goods, services, or works to be contracted.",
```

docs/history/changelog.md
```diff
@@ -203,6 +210,8 @@ Per the [normative and non-normative content and changes policy](https://docs.go
   * [#1189](https://github.com/open-contracting/standard/pull/1189) `Document.url`, to use precise wording and to describe how to indicate a part of a document.
   * [#1229](https://github.com/open-contracting/standard/pull/1229) Standardize the descriptions of `planning.documents`, `tender.documents`, `awards.documents`, `contracts.documents`, and `contracts.implementation.documents`.
   * [#1335](https://github.com/open-contracting/standard/pull/1335) Standardize the descriptions of `planning`, `planning.rationale`, `planning.budget`, `planning.documents`, and `planning.milestones`.
+  * [#1529](https://github.com/open-contracting/standard/pull/1529) `Budget`, `Budget.id`, `Budget.description`, `Budget.amount`, `Budget.project`, `Budget.uri`, `Budget.source`, `tender.value`, `tender.minValue`, `awards.value`, `Transaction`, to replace "contracting process" with "contract(s)"
+  * [#1530](https://github.com/open-contracting/standard/pull/1530) `Buyer`, `tender.items`, `awards.items`, `contracts.items`, `Item`, `Item.description`, `Item.unit`, `Unit`, `transaction.providerOrganization`, `transaction.receiverOrganization` standardize reference to good(s), service(s) and/or work(s)
```
duncandewhurst:
```suggestion
  * [#1530](https://github.com/open-contracting/standard/pull/1530) `Buyer`, `tender.items`, `awards.items`, `contracts.items`, `Item`, `Item.description`, `Item.unit`, `Unit`, `transaction.providerOrganization`, `transaction.receiverOrganization` standardize reference to goods, services and/or works.
```

duncandewhurst (changes requested):

schema/release-schema.json
jpmckinney: The wording seems to have originated in https://github.com/open-contracting/standard/issues/105#issuecomment-59630318

Anyway, a service (i.e. an activity) is intangible, e.g. advisory services, so it's already covered. We shouldn't introduce a unique concept at the stage of contract items.

schema/release-schema.json
odscjen: yes because it is possible for the buyer to be procuring a mixture of goods, services and works in a single contract

schema/release-schema.json
odscjen: yes, again I thought it was possible for a mixture of these to be the focus of the transaction

schema/release-schema.json
odscjen: yes, as previous comment

schema/release-schema.json
odscjen: I was attempting to find a way to express that an item will be a single type. To take the example of "cleaning services", these can take many forms but if the classification being used is simply "cleaning services" without further disambiguation then this is being presented as a single, unique type of service. "Type" didn't feel like the right word though. "Line-item" could work but it is then a bit of a circular description, "an item is a line-item...".

schema/release-schema.json
jpmckinney: The original "or" is not an exclusive-or; as such, it functions as "one or more".

"and" would typically mean "all" here, unless it's read as "goods (if any), services (if any) and works (if any)".

I think "or" reads better.

odscjen: fair enough, I'll change this back to 'or'

schema/release-schema.json
duncandewhurst: Thanks, I agree. @odscrachel taking account of https://github.com/open-contracting/standard/pull/1530#discussion_r908510347, we should update this to:

```suggestion
          "description": "The goods, services or works in this contract. If the items contracted are identical to the items awarded, this field should be omitted.",
```

schema/release-schema.json
duncandewhurst: Per https://github.com/open-contracting/standard/pull/1530#discussion_r908510347, let's keep 'or'

schema/release-schema.json
duncandewhurst: Per https://github.com/open-contracting/standard/pull/1530#discussion_r908510347, let's keep 'or'

schema/release-schema.json
duncandewhurst: Hmm, yes, it's a tricky one. I checked the [glossary crosswalk](https://docs.google.com/spreadsheets/d/1Nh_HjyYuNk8wuu_Tb0kcPDyV6j8pAMgXqJ3MgwpXZjQ/edit?skip_itp2_check=true#gid=0) and the item concept lacks a definition. I also looked at a few definitions of 'line item' ([Collins](https://www.collinsdictionary.com/dictionary/english/line-item), [Wiktionary](https://en.wiktionary.org/wiki/line_item), [Law Insider](https://www.lawinsider.com/dictionary/purchase-order-line)), which are quite circular, too.

How about:

```suggestion
      "description": "A specific instance of the goods, services, or works to be contracted.",
```
That works for the examples given in https://github.com/open-contracting/standard/pull/1530#issuecomment-1163380962

duncandewhurst (changes requested):
See my replies to the previous round of review comments.

duncandewhurst (approved):
@jpmckinney you might want to check that you're happy with the outcome of https://github.com/open-contracting/standard/pull/1530#discussion_r905676571

jpmckinney (dismissed):

jpmckinney:
> @jpmckinney you might want to check that you're happy with the outcome of [#1530 (comment)](https://github.com/open-contracting/standard/pull/1530#discussion_r905676571)

From the links in https://github.com/open-contracting/standard/pull/1530#discussion_r908960713

Collins:

> a specific item, esp., an amount listed separately, in a budget, appropriation bill, etc.

Wiktionary:

> An item appearing on a single line in any schedule of information.
> One item on an order, regardless of quantity.

Law Insider:

> a single position on a Purchase Order
> a line in a Purchase Order that specifies the Product(s) and quantity of Product(s) being ordered

The e-Procurement Ontology defines the related concept of Lot as:

> A qualitative, quantitative or strategic subdivision of the goods, services or works to be procured, allowing the award of one or more contracts.

While an item and a lot are sometimes used to model the same thing, a single lot can contain a mix of items (especially if we define a procurement without divisions as nonetheless having a single lot). And whereas a lot can be a mix, an item is expected to be more homogenous. So while "subdivision" is a handy way to resolve the wording of Item, we probably want to reserve it for the definition of Lot.

The word "instance" already implies "single" and "specific" ("an example or single occurrence of something", "a particular case"), so we have to do away with it for the reasons in the discussion.

Given that all available definitions are circular, I am fine with circularity. We don't define the word "contract" either, instead deferring to whatever legal regime. It's fine for the title of a concept to be so clear that it requires no description.

So let's go with "line-item" as suggested in one comment.

jpmckinney (commented):
Removing approval per last comment.

jpmckinney (approved):
I made the commit myself as folks are out sick.

PR #1531, opened 2022-06-24
Update the guidance on change history

### LINKED ISSUE #1412: Update the guidance on releases being optional on Record
In OCDS 1.2 the releases are now optional on Record in favor of allowing the publishers to publish only compile releases. We need to update the relevant guidance on this.

From https://github.com/open-contracting/standard/issues/1092#issuecomment-910907474:

>  it would be relevant to make the approach of only publishing compiled releases known on some of the change history or easy releases pages.

yolile:
Closes #1412


docs/guidance/build/change_history_options.md
```diff
@@ -0,0 +1,11 @@
...
+
+In order to publish a change history, your data sources need to track what changed and when, for example, by implementing events and/or last modified dates on database records. The [system architecture](../build/system_architectures) for your OCDS implementation also needs to store historic releases. If you cannot store historic releases, you can [publish full releases and update the release `id` and `date` each time there is a change](../build/full_updates).
+
+However, publishing a change history is not always possible. Sometimes, it is only possible to publish the full information about a contracting process at a given time This might be the case if, for example, your data sources don't implement versioning, events, good identifiers, or proper controls (e.g. allowing users to just change a supplier's name instead of requiring the user to register a new supplier in the system). This also might be the case if you collect non-OCDS data from a data source and re-publish it as OCDS and you don't have direct access to the data source.
```
duncandewhurst:
```suggestion
However, publishing a change history is not always possible. Sometimes, it is only possible to publish the full information about a contracting process at a given time. This might be the case if, for example, your data sources don't implement versioning, events, good identifiers, or proper controls (e.g. allowing users to just change a supplier's name instead of requiring the user to register a new supplier in the system). This also might be the case if you collect non-OCDS data from a data source and re-publish it as OCDS and you don't have direct access to the data source.
```

docs/guidance/build/full_updates.md
```diff
@@ -1,8 +1,8 @@
-```{workedexample} 'Easy' Releases
+```{workedexample} Full updates
 :tags: release
 ```
 
-# 'Easy' Releases
+# Full updates
 
 OCDS encourages the use of the [releases and records](../../primer/releases_and_records) model in order to publish up-to-date, timely data. However, sometimes publishers can't fully support the model, because historic data of contracting (or planning) processes is not stored in the source system(s). In such case, the publisher can produce only one release for each process, and the release gets overridden with new updates.
```
duncandewhurst:
```suggestion
OCDS encourages publishers to use the [releases and records](../../primer/releases_and_records) to publish both a change history and the current state of each contracting (or planning) process. However, sometimes publishers can't publish a change history because their [system architecture](../build/system_architectures) does not store historic releases. In such cases, implementers can publish full updates by making a single release per process available and updating the release `id` and `date` each time there is a change.
```

docs/guidance/build/incremental_updates.md
```diff
@@ -1,8 +1,8 @@
-```{workedexample} Change history
+```{workedexample} Incremental updates
 :tags: release,record
 ```
 
-# Change history
+# Incremental updates
 
 The following example shows how to publish data about a contracting process. Publish a release for each event in the process and update the record each time you publish a release.
```
duncandewhurst:
```suggestion
The following example shows how to publish data using the releases and records model: publish a release for each event in the process and update the record each time you publish a release.
```

docs/guidance/build/no_change_history.md
```diff
@@ -0,0 +1,23 @@
+```{workedexample} No change history
+:tags: record
+```
+
+# No change history
+
+OCDS supports and encourages the publication of the change history of a contracting process. But, for some publishers, publishing this information is not feasible. For example, if your system doesn’t implement versioning or track changes, or if you are collecting non-OCDS data from a source and republishing it as OCDS. Similarly, if you publish historical information that won’t change anymore, you might have only the latest state of the contracting process, without its change history.
+
+If this is your case, the best way to publish this information in OCDS is in a compiled release, as a compiled release represents the state of the contracting process at the time of publication.
```
duncandewhurst:
```suggestion
If you cannot use releases to publish incremental or full updates for the reasons outlined in [Change History Implementation Options](change_history_options.md), you ought to publish the current state of each contracting process using compiled releases.
```

docs/guidance/map/amendments.md
```diff
@@ -124,7 +124,7 @@ Note that the `compiledRelease` contains all the items, included the latest one
 
 ### Example 3: Amendments in a Easy Releases scenario
```
duncandewhurst:
```suggestion
### Example 3: Amendments in full updates
```

docs/history/changelog.md
```diff
@@ -262,6 +262,7 @@ Per the [normative and non-normative content and changes policy](https://docs.go
 * [#1375](https://github.com/open-contracting/standard/pull/1375) Update guidance for empty fields in the merging documentation.
 * [#1466](https://github.com/open-contracting/standard/pull/1466) Reference worked examples in release and record reference documentation.
 * [#1466](https://github.com/open-contracting/standard/pull/1482) Add examples in release reference documentation.
+* [#1531](https://github.com/open-contracting/standard/pull/1531) Update wording on change history. Add guidance on change history options. Add a working example on no change history.
```
duncandewhurst:
```suggestion
* [#1531](https://github.com/open-contracting/standard/pull/1531) Update wording on change history. Add guidance on change history options. Add worked example on no change history.
```

docs/schema/merging.md
```diff
@@ -76,7 +76,7 @@ The structure of the versioned release is described by the {download}`versioned
 
 ### Merge routine
 
-To create a compiled or versioned release, you must:
+To create a compiled or versioned release when multiples individual releases are available, you must:
```
duncandewhurst:
```suggestion
To create a compiled or versioned release when multiple individual releases are available, you must:
```

docs/guidance/build/change_history_options.md
```diff
@@ -0,0 +1,11 @@
+# Change History Implementation Options
```
duncandewhurst: On re-reading, I think this page is still quite confusing. I suggest we replace it with the following, which is more closely aligned with the suggestion in https://github.com/open-contracting/standard/issues/1412#issuecomment-925014101. If you agree, we can update the worked examples accordingly.

---

# Change history implementation options

The [releases and records model]() allows publishers to disclose the change history and current state of a contracting process by publishing:

* Updates for changes to the process, known as [releases]().
* The latest values of each field, known as a [compiled release]().
* A version history of each field, known as a [versioned release]().

## Releases or compiled releases

To meet the widest range of user needs the preferred approach is to [publish at least releases and compiled releases](change_history.md).

To publish releases, you need to be able to identify what changed and when in your data sources, for example, by using events or last modified dates. However, it isn't always possible to identify what changed and when, for example because your data sources don't implement versioning, events, good identifiers, or proper controls. In which case, publishing releases can lead to incoherent data due to how the [merge routine](merge_routine.md) handles array entries.

Therefore, if you cannot identify what changed and when in your data sources, you ought to publish [compiled releases only](no_change_history.md).

## Incremental or full updates

If you can publish releases, the preferred approach is to implement a [system architecture](system_architecture.md) that can store and provide access to historic releases. In which case, you can choose to publish either incremental updates or full updates. Incremental updates are releases that contain only new or changed fields and full updates are releases that contain new, changed and unchanged fields.

If your system architecture cannot store and provide access to historic releases, then you need to [publish full updates](full_updates.md). Otherwise, users would not be able to access the values of unchanged fields.

docs/guidance/build/incremental_updates.md
```diff
@@ -1,8 +1,8 @@
-```{workedexample} Change history
+```{workedexample} Incremental updates
```
duncandewhurst: This page isn't really about incremental updates. It's about the whole releases and records model. In fact, the examples actually show full updates.

duncandewhurst (changes requested):

docs/guidance/build/full_updates.md
yolile: This change wasn't part of this PR but it makes sense, but the addition of "contracting (or planning) processes" was part of another PR and I guess we want to keep that expression

docs/guidance/build/incremental_updates.md
yolile: >  In fact, the examples actually show full updates.

Hmm, well, the contract and implementation releases don't include the info about the past stages. I guess we can change the award release to remove the tender section from there and we should be fine?

docs/guidance/build/full_updates.md
jpmckinney: Yes, see #1409. The handbook has also been updated on the use of this term: https://ocds-standard-development-handbook.readthedocs.io/en/latest/meta/style_guide.html#word-choice

docs/guidance/build/change_history_options.md
yolile: It looks good to me. Maybe before the "Releases or compiled releases" section, we can add a sentence:
"As part of your OCDS implementation process, you have two main decisions: whether to have a change history or not and if you have a change history, whether to have incremental updates or full updates."
@jpmckinney, any thoughts?

jpmckinney: The heading should be "Individual or compiled releases" rather than "releases or compiled releases".

That said, as you suggest, the question is more "change history or not", and the conclusion is "individual (and compiled) releases, or only compiled releases".

So I would reframe that first section around a "change history or not" perspective, rather than leading with the possible conclusions.

duncandewhurst: > The heading should be "Individual or compiled releases" rather than "releases or compiled releases".

Sounds good

> That said, as you suggest, the question is more "change history or not", and the conclusion is "individual (and compiled) releases, or only compiled releases".

I think that we might create some confusion by referring to the publication of individual releases without storing and publishing historic releases (i.e. the old 'easy releases' model) as a change history. I know that model means that a 3rd party can maintain a change history if they scrape the data regularly, but from the publisher's point of view it looks more like just publishing the latest version of the data albeit with careful attention to setting ids etc.

I also think that, in terms of the implications for the software they need to write, the decision about whether or not to store and publish historic releases is probably more significant for a publisher than deciding between publishing individual or compiled releases.

jpmckinney: Good points! So are the decisions then:

* Change history or not (i.e. having multiple releases available at once about the same contracting process, or not)

If no change history, then:

* Individual releases ("easy releases") or compiled releases

If change history, then:

* Incremental updates or full updates

duncandewhurst: Yes, I think that's right. I've taken a pass at reframing along those lines below. I also added some criteria for choosing between incremental and full updates. I think there are a couple of other decisions for implementers that choose to publish a change history, but I think the page gets complicated if we start to include them:

* Whether to publish records (as well as releases)
* If publishing records:
  * Whether to publish compiled releases
  * Whether to publish versioned releases

---

# Change history implementation options

To implement OCDS, you need to decide whether or not to publish a change history.

Using the [releases and records model](), you can publish:

* Updates for changes to the process, known as [releases]().
* The latest values of each field, known as a [compiled release]().
* A version history of each field, known as a [versioned release]().

To publish a change history, you need to implement a [system architecture](system_architecture.md) that can store and provide access to historic releases. You also need to be able to identify what changed and when in your data sources, for example, by using events or last modified dates.

However, it isn't always possible to identify what changed and when, for example, because your data sources don't implement versioning, events, good identifiers, or proper controls. In which case, publishing releases can lead to incoherent data due to how the [merge routine](merge_routine.md) handles array entries.

If you choose to publish a change history, you need to decide whether to publish [incremental or full updates]().

If you choose not to publish a change history, you need to decide whether to publish [individual releases or compiled releases]().

## Incremental or full updates

Incremental updates are releases that contain only new or changed fields and full updates are releases that contain new, changed and unchanged fields.

Incremental updates are smaller than full updates, which can reduce the size of your datastore and reduce download times for data users. If you anticipate many releases per contracting process or very many contracting processes, you ought to consider incremental updates.

If you choose to publish incremental updates, the preferred approach is to also publish complied releases so that users can easily access the latest state of the contracting process.

## Individual releases or compiled releases

If you can identify what changed and when in your data sources, then you ought to [publish individual releases](). To publish individual releases, you need to update the release identifier each time there is a change to the data about a contracting process. That way, data users can easily identify when there is a change. You also need to pay careful attention to consistent use of identifiers for objects in arrays.

If you cannot identify what changed and when, then you ought to [publish compiled releases only]().

jpmckinney: > I think there are a couple of other decisions for implementers that choose to publish a change history, but I think the page gets complicated if we start to include them:
> 
> * Whether to publish records (as well as releases)
> * If publishing records:
>   
>   * Whether to publish compiled releases
>   * Whether to publish versioned releases

Agreed, let's leave records out for this page.

I've made comments and small edits below.

On word choice: Use "determine" when the meaning is "establish" or "ascertain" in general. Use "identify" only when you are establishing the identity of something or when working with identifiers.

> # Change history implementation options

Since one option is "no change history", this page title could use some work. I don't have a suggestion atm.

> First, you need to decide whether or not to publish a change history.

> Using the releases and records model, you can publish:
> 
> * Updates for changes to the process, known as releases.
> * The latest values of each field, known as a compiled release.
> * A version history of each field, known as a versioned release.

I don't think these bullets serve a purpose, and they break up the logical flow, so we can remove them. If they are here to remind readers of terms' definitions, we can instead describe the terms the first time they are used.

> To publish a change history, you need to be able to determine what changed and when within your data sources: for example, by using event-based triggers or last-modified dates. You also need to use a [system architecture](system_architectures) that has a [separate OCDS datastore](system_architectures.md#separate-ocds-datastore).

The second sentence now follows logically from the first, and I made the second sentence more explicit about the requirement.

> Sometimes, it isn't possible to determine what changed and when: for example, because your data sources don't implement triggers, timestamps, versioning, or consistent identifiers. If identifiers are inconsistent, publishing a change history can lead to incoherent data, due to how the [merge routine](../../schema/merging) handles array entries.

Not sure what "proper controls" implies, so I removed it. If we add it back, we should give an example.

> If you choose to publish a change history, you need to decide whether to publish [incremental or full updates](#incremental-or-full-updates).
> 
> If you choose not to publish a change history, you need to decide whether to publish [individual releases or compiled releases](#individual-releases-or-compiled-releases).
> 
> ## Incremental or full updates
>
> A full update is a [release](../../schema/reference) that contains all the available information about the contracting (or planning) process. An incremental update is a release that contains a subset of the available information, focusing on what's new or changed.

An incremental update isn't *only* new or changed, since it might re-state, for example, the startDate when only the endDate changed. I reworded.

> Incremental updates are smaller than full updates. Smaller releases reduce the OCDS datastore's size and users' download times. If you anticipate many releases per contracting process or very many contracting processes, you ought to consider incremental updates.
> 
> If you choose to publish incremental updates, the preferred approach is to also publish [complied releases](../../schema/records_reference.md#compiled-release), so that users can easily access the latest state of the contracting process.
> 
> ## Individual releases or compiled releases
>
> If you can determine what changed and when within your data sources – but cannot implement a separate OCDS datastore – then you ought to publish individual releases. To publish individual releases, you need to update the release identifier along with each change to the data about a contracting process. That way, users can track each update to the contracting process. You also need to ensure that identifiers are consistent for objects in arrays.

I added a second condition to the first sentence, as otherwise they should implement a change history.

> If you cannot determine what changed and when, then you ought to publish compiled releases only.

duncandewhurst: Those updates look good to me, with one addition to the second sentence below in order to make the relationship between publishing a change history and having a separate OCDS datastore explicit:

> To publish a change history, you need to be able to determine what changed and when within your data sources: for example, by using event-based triggers or last-modified dates. You also need to use a [system architecture](https://github.com/open-contracting/standard/pull/system_architectures) that has a [separate OCDS datastore](https://github.com/open-contracting/standard/pull/system_architectures.md#separate-ocds-datastore) so that you can provide access to historic releases.

jpmckinney: Sounds good!

yolile:
Thank you, @duncandewhurst, for the changes. I think this is ready for James to review again, right?
And should we add a link to the new page from the [Determine data needs](https://standard.open-contracting.org/latest/en/guidance/design/#determine-data-needs) section? (particularly to the "Whether to publish a change history" item)

jpmckinney:
> And should we add a link to the new page from the [Determine data needs](https://standard.open-contracting.org/latest/en/guidance/design/#determine-data-needs) section? (particularly to the "Whether to publish a change history" item)

Yes!

jpmckinney (commented):
I rewrote `change_history_options` per the discussion here: https://github.com/open-contracting/standard/pull/1531#discussion_r914394363

That page now uses "full updates" to mean "full updates in the context of a change history" – in contrast to incremental updates. It also uses "individual releases" to mean "individual releases, when you can't publish a change history" – in contrast to "compiled releases".

However, that usage is inconsistent with the other pages.

I suggest renaming the other pages:

* Incremental updates -> Change history. As Duncan commented, the examples are a mix of full and incremental updates. In any case, we don't have a page for each of these (and I don't think we need one, as the distinction is fairly minor).
* Full updates -> Individual releases with no change history
* No change history -> Compiled releases with no change history

To be clear, if a publisher follows the 'easy releases' approach, then they aren't publishing a change history (at best, a third-party can reconstruct a change history).

There might need to be some corresponding changes to the introductory paragraphs on those other pages.

duncandewhurst:
@yolile, @jpmckinney are you happy for me to pick up and complete this PR?

yolile:
@duncandewhurst Yes, I'm!

docs/guidance/build/incremental_updates.md
duncandewhurst: Resolving as the page has been renamed.

docs/guidance/build/change_history_options.md
duncandewhurst: Resolving as the content has been redrafted.

duncandewhurst:
To-do:

- [x] Review built version of documentation
- [x] Check guidance on setting `.date` in guidance on hashing (no last modified date)
- [x] Cherry pick updates to amendments example and update JSON directives in example 3

duncandewhurst:
@jpmckinney this is ready for your review. I've made the changes suggested in your last comment, simplified the examples in line with our discussions elsewhere, and made various copy edits for consistency and clarity.

Two things to flag:

1. The old easy releases guidance describes how to use a hash function to update release `id` in the absence of a last modified date. However, it doesn't explain how to set `date`. In the new individual releases guidance, I've recommended setting `date` to the date the release was generated (i.e. the date the user downloaded the data). Happy to discuss if there might be a better approach.
2. In the individual releases guidance, I've preserved the recommendation to prefix `.id` with the ocid so that users don't need to filter on `ocid` to find releases belonging to the same contracting process. However, I'm inclined to remove that recommendation so that the examples align with the rule that `.id` need only be unique within the scope of the contracting process. It seems like more work to filter on part of a field (`.id`) than on a separate field. What do you think?

Edit: I forgot to mention, this PR includes some commits from https://github.com/open-contracting/standard/pull/1661/ which is also awaiting review. Whichever is reviewed first, I'll make the necessary updates to the other one.

jpmckinney:
> In the individual releases guidance, I've preserved the recommendation to prefix .id with the ocid so that users don't need to filter on ocid to find releases belonging to the same contracting process. However, I'm inclined to remove that recommendation so that the examples align with the rule that .id need only be unique within the scope of the contracting process. It seems like more work to filter on part of a field (.id) than on a separate field. What do you think?

Agree - I prefer your suggested simplification.

duncandewhurst:
Great. I've done that in https://github.com/open-contracting/standard/pull/1531/commits/c8483823d93fe5f3d8414fb1f52feb26c5d78338

jpmckinney:
Noting that amendments.md includes changes from #1661, but my suggested edits in that PR don't overlap, so there shouldn't be a merge conflict. amendments.md has an extra commits in this PR, which can be reviewed independently: 

* https://github.com/open-contracting/standard/commit/3c64c95d0f84fe6de9b807273a3d75d98c6eee8e#diff-d3ab3df28ed0eba11bf5bcc05a619b2e8048d802f73ce950116f800fba79dd6d (small) 
* https://github.com/open-contracting/standard/commit/104d2a48a075476fab384f6804934b11efc18b99#diff-d3ab3df28ed0eba11bf5bcc05a619b2e8048d802f73ce950116f800fba79dd6d (merge)
* https://github.com/open-contracting/standard/commit/0cc23e17878db6dfa73074624efb0e8c21b85c54#diff-d3ab3df28ed0eba11bf5bcc05a619b2e8048d802f73ce950116f800fba79dd6d
* https://github.com/open-contracting/standard/commit/6f51570b2ccf9f1b8b4ea1657c21de17ab86fc41#diff-d3ab3df28ed0eba11bf5bcc05a619b2e8048d802f73ce950116f800fba79dd6d

PR #1558, opened 2022-10-28
Upgrade jsonref

jpmckinney:
- build: Upgrade to latest profile template
- chore: Upgrade to jsonref 1.0 API


jpmckinney:
This only affects the build, not the content, so merging.

PR #1637, opened 2023-09-11
add sdg and sdgTarget to classificationScheme codelist

### LINKED ISSUE #1098: itemClassificationScheme: Add codes for SDG and SDGTarget
The [Himachal Pradesh Health Procurement Performance Index](https://hp.openbudgetsindia.org/#/procurement/analysis) tags contracting process with the SDG targets they relate to.

To model this in OCDS they are using the [tender classification extension](https://extensions.open-contracting.org/en/extensions/tenderClassification/master/) with `.scheme` set to 'SDG' or 'SDGTarget', depending on whether the contracting process is tagged with a goal or a target.

**Proposal**

Add 'SDG' and 'SDGTarget' to the itemClassificationScheme codelist (since https://github.com/open-contracting/standard/pull/1033 this codelist can be used in contexts other than items):

Code | Title | Description | Source | Category
-- | -- | -- | -- | -- 
SDG | Sustainable Development Goals | The Sustainable Development Goals are a collection of 17 interlinked goals set by the United Nations General Assembly and designed to be a "blueprint to achieve a better and more sustainable future for all". | https://sdgs.un.org/goals | contractingProcess
SDGTarget | Sustainable Development Goal Targets | The Sustainable Development Goal Targets are specific targets for each Sustainable Development Goal. | https://sdgs.un.org/goals | contractingProcess

There's an Excel sheet of the goals, targets and indicators available at https://unstats.un.org/sdgs/indicators/indicators-list/, so that might be a better source to reference than the main SDGs site.

cc @pindec

odscjen:
closes #1098 

Copied from OC4IDS codelist and used https://unstats.un.org/sdgs/indicators/indicators-list/ as the source rather than https://sdgs.un.org/goals


duncandewhurst (approved):

jpmckinney (commented):
The Category is `purpose` but OCDS doesn't have a `purpose` field like OC4IDS.

I think it's meant to be `item`? The other options so far in this codelist are `legalBasis` and `organization`.

odscjen:
Good catch, but I think it should be 'tender' as its the tender that's being classified not any particular items within it. Tender is used as the category in the [Tender classification](https://extensions.open-contracting.org/en/extensions/tenderClassification/master/codelists/) extension

jpmckinney:
We haven't used 'tender' yet (except in that extension), and many of the 'item' classifications can be used in both 'tender' and 'item' context.

For example, the EU profile uses tender.classification with CPV codes, because the EU doesn't subdivide procedures or lots into individual items.

So, I would use 'item', since presumably individual items can be tagged as SDG-related.

jpmckinney:
On second thought, if we only expect a procedure to be tagged with SDGs, then we can leave it as 'tender'.

jpmckinney (approved):

PR #1642, opened 2023-10-09
Add planning final status fields and codelist

### LINKED ISSUE #1582: Date of final state change
Splitting from #1160 

Copying some discussion from CRM-8486.

A publisher wants to disclose the date on which the process reached its final state. This date might differ from the date on which the notice is published (`date`).

We can either have:

1. One field for each state (`dateCancelled` and `dateWithdrawn` – presumably, `dateUnsuccessful` is not needed, as this occurs simultaneously with the `tenderPeriod/endDate` if there are no qualified bids)
2. One field for the final state (since the final states are disjoint – i.e. a process can't be both cancelled and withdrawn, etc. – we only ever have one date to model). The field name would follow whatever name we choose for the status.

I would go with 2.

### LINKED ISSUE #1583: Planning final status codelist
Related: #1160 CRM-8486

A buyer can publish a notice about a future procurement, and then later decide to not follow through on that future procurement.

The model needs to be able to disambiguate whether it's the planning process itself that is cancelled, or the future (non-existent) process. The tricky part is that it's hard to make a statement about something that doesn't exist; we can't mint an OCID for the contracting process and then set its `tender/status` to 'cancelled', because that implies that the contracting process had at least a minimal existence.

After #1160 and #1582, we can add `planning/finalStatus` and `planning/finalStatusDate`, with one code being 'cancelled' (as in the buyer didn't go ahead with the planning process, e.g. to hold consultations for a new recreational facility, which it then decided not to go ahead with) and 'unsuccessful' (as in the planning process didn't yield a contracting process).

duncandewhurst:
* Closes #1583 
* Related to #1582


duncandewhurst:
The spell-check failure is addressed by https://github.com/open-contracting/standard/pull/1641

schema/codelists/planningFinalStatus.csv
```diff
@@ -0,0 +1,3 @@
+Code,Title,Description
+cancelled,Cancelled,"The planning process was cancelled and did not yeild a contracting process. For example, a buyer did not go ahead with planned consultations about the construction of a new recreational facility and decided not to build the facility."
```
odscjen:
```suggestion
cancelled,Cancelled,"The planning process was cancelled and did not yield a contracting process. For example, a buyer did not go ahead with planned consultations about the construction of a new recreational facility and decided not to build the facility."
```

schema/release-schema.json
```diff
@@ -178,6 +178,30 @@
           "items": {
             "$ref": "#/definitions/Milestone"
           }
+        },
+        "finalStatus": {
+          "title": "Planning final status",
+          "description": "The final status of the planning process, from the closed [planningFinalStatus](https://standard.open-contracting.org/{{version}}/{{lang}}/schema/codelists/#planning-final-status) codelist.",
```
odscjen:
```suggestion
          "description": "The final status of the planning process, using the closed [planningFinalStatus](https://standard.open-contracting.org/{{version}}/{{lang}}/schema/codelists/#planning-final-status) codelist.",
```

as per the style guide

odscjen (changes requested):
My reading of https://github.com/open-contracting/standard/issues/1582 and the associated CRM issue is that it's not just the planning final status date that's needed by also a `tender.finalStatusDate`

duncandewhurst:
Ah, good spot. I shouldn't have marked #1582 as being closed by this PR. I figure it makes sense to do those other final status date fields when adding the new final status fields in #1160.

Alternatively, we could do all of the final status and final status date fields in this PR. @jpmckinney let me know what you prefer.

jpmckinney:
My comment in #1583 was to close it "After https://github.com/open-contracting/standard/issues/1160 and https://github.com/open-contracting/standard/issues/1582". #1583 is more like a new feature, whereas #1160 is a major change (#1582 is also more like a new feature). The major change should lead, before related features are implemented.

I prefer to close #1160 on its own. #1582 and #1583 could potentially be combined.

duncandewhurst:
Sounds good! I'll pause work on this PR and prepare a PR for #1160 (PR: https://github.com/open-contracting/standard/pull/1648)

jpmckinney:
@duncandewhurst This is now unblocked.

duncandewhurst:
I think this is ready for review, unless I missed something.

jpmckinney (approved):
I aligned 'cancelled' with all other 'cancelled' in other codelists (basically, it's a decision by the buyer, always for the same potential reasons).

I also aligned 'unsuccesful' in terms of structure, and using a less specific example. A specific example, is e.g. a building application where residents might reject it.

odscjen (approved):

Issue #557, opened 2017-08-31
partyRole: Add 'interestedParty' code


duncandewhurst:
From @dwasyl:

> `interestedParty` is a perfect format of what I’m looking for with the register concept. I see where it would be useful in the P3 (public private partnerships) add-on, although I would suggest that it is a decent code to bump up to the main schema since it could easily apply to any tendering process or separate out on it's own. For me it could be linked to the question/answer data model as well.

This code exists in the PPP profile with the following definition:

> A party which has expressed an interest in the project, for example by purchasing tender documents or submitted clarification questions, but has not yet formally submitted a bid.

mireille-raad:
would it make sense to call it something more specific like interestedBidder ? an interestedParty can be anyone like an NGO, a bank, a news media outlet, a developer poking around with open data. I didn't get what it really meant until I read the email....so an open data users can get confused with what it means to be an interested party.

duncandewhurst:
> an interestedParty can be anyone like an NGO, a bank, a news media outlet, a developer poking around with open data

These are all valid uses of the code

`interestedParty` was added to the PPP profile to represent the different types of organization which had purchased bidding documents.

In the pilot PPP we worked with, it wasn't possible to know whether an organization which purchased bidding documents was interested in submitting a bid, or just interested in the process and wanted to submit questions, until they submitted a bid (or the deadline for submissions passed).

Once an `interestedParty` submits a bid they become a `tenderer`, in which case `interestedBidder` is redundant.

dwasyl:
I'd concur with @duncandewhurst. Interest parties in a tender process can be anyone, no one knows who will actually submit a bid until bidding is closed.

Some with the enquiries addon, anyone can ask question, since a proponent hasn't necessarily decided if they will bid or not.

Often the `interestedParty` list would be used to send updates/addendums for the tender.

mireille-raad:
ha! thanks for clarification :D makes much more sense now :)

jpmckinney:
> A party which has expressed an interest in the project, for example by purchasing tender documents or submitting clarification questions, but has not yet formally submitted a bid.

@duncandewhurst Should this not refer to 'contracting process' rather than project? I propose:

> A party that has expressed an interest in the contracting process: for example, by purchasing tender documents or submitting clarification questions.

[...]

PR #1644, opened 2023-10-09
schema/codelists/partyRole.csv: Clarify description of interestedParty

### LINKED ISSUE #1522: `partyRole` codelist: clarify description of interestedParty
Spin-off from https://github.com/open-contracting/standard/pull/1513#discussion_r878136484. 

'enquirer' in `partyRole` is defined as

> An organization that has made an enquiry during the enquiry phase of a contracting process.

There are several grey areas: 
1) Organizations also ask questions at other stages and currently we don't have any codes for that: 
    - planning ("could you clarify vague paragraph XYZ in the PIN", "are you aware of solution X that would solve your need but might end up not being applicable in your future contracting process because of thing ABC that you mention in your planning process" or perhaps even "do you have any news on when the contracting process will be launched")
    - award ("when will you publish the evaluation results") 
    - contract ("is the contract published anywhere?") 

    The two obvious solutions would be to Broaden the scope of 'enquirer' by removing "the enquiry phase" or adding codes for questions asked in other phases. (I note that "the enquiry phase" is not something I would a priori know, or know where it starts). 

2) 'interestedParty' is defined as 

    > "An organization that has expressed an interest in the contracting process: for example, by purchasing tender documents or submitting clarification questions." 

    where the second condition overlaps with 'enquirer', so it might be better to reduce the scope of 'interestedParty' to avoid the overlap (i.e. remove "or submitting clarificiation questions" from 'enquirer'.). (However, this is isn't a backward compatible change...)

3) There is no code for organizations that submit replies (or ask questions?) when doing a market consultation doing a planning phase (market consultations might sometimes be a bit of a dialogu, so this might overlap a bit). We could have a new code for that.

I would have a tendency to enlarge the scope of 'enquirer', remove the overlap with 'interestedParty', and add a code for 'marketConsultationRespondent' or something like that and do the rest only if user demand requires it. (One question could be whether people want to be distinguishing tenderers that did and didn't ask a question, but if went down that route, then we should just remove 'enquirer', because it would be subsumed already under 'interestedParty'...)

duncandewhurst:
* Closes #1522


schema/codelists/partyRole.csv
```diff
@@ -8,7 +8,7 @@ enquirer,Enquirer,An organization that has made an enquiry during the enquiry ph
 payer,Payer,An organization making a payment from a transaction.,,
 payee,Payee,An organization in receipt of a payment from a transaction.,,
 reviewBody,Review body,An organization responsible for the review of this procurement process. This organization often has a role in any challenges made to the contract award.,,
-interestedParty,Interested party,"An organization that has expressed an interest in the contracting process: for example, by purchasing tender documents or submitting clarification questions.",,
+interestedParty,Interested party,"An organization that has expressed an interest in the contracting process: for example, signing up for notifications about the process, submitting questions or downloading or purchasing procurement documents.",,
```
odscjen:
```suggestion
interestedParty,Interested party,"An organization that has expressed an interest in the contracting process: for example, by signing up for notifications about the process, submitting questions or downloading or purchasing procurement documents.",,
```

odscjen (changes requested):

schema/codelists/partyRole.csv
```diff
@@ -4,11 +4,11 @@ procuringEntity,Procuring entity,"The organization managing the contracting (or
...
 tenderer,Tenderer,An organization that submitted a bid.,,
 funder,Funder,The funder is an organization providing money or finance for this contracting process.,,1.2
-enquirer,Enquirer,An organization that has made an enquiry during the enquiry phase of a contracting process.,,
+enquirer,Enquirer,An organization that submitted an enquiry during the enquiry period of the contracting process.,,
 payer,Payer,An organization making a payment from a transaction.,,
 payee,Payee,An organization in receipt of a payment from a transaction.,,
 reviewBody,Review body,An organization responsible for the review of this procurement process. This organization often has a role in any challenges made to the contract award.,,
-interestedParty,Interested party,"An organization that has expressed an interest in the contracting process: for example, by purchasing tender documents or submitting clarification questions.",,
+interestedParty,Interested party,"An organization that has expressed an interest in the contracting process: for example, by signing up for notifications about the process, submitting questions or downloading or purchasing procurement documents.",,
```
jpmckinney: The examples suggest that the party is a bidder, but in #557 we clarified that it can be anything – so adding a sentence for further clarity here.

Using "enquiries" consistently to avoid longer "questions or requests for clarification".

Using "requesting" instead of two words.

```suggestion
interestedParty,Interested party,"An organization that expressed interest in the contracting process: for example, by signing up for notifications, submitting enquiries, or requesting procurement documents. The organization can be a potential supplier, news organization, investor, etc.",,
```

jpmckinney (commented):
I also added commits to:

* clarify enquirer (use "period" not "phase")
* similarly, avoid "phase" except when referring to OCDS implementation phases

Issue #364, opened 2016-08-04
Rules and guidance for re-publishers of data


timgdavies:
Linked to the discussion in #130 and exploration of the technical requirements for an OCID registration tool, we have been exploring whether any special fields are required to indicate that a publisher is 're-publishing' open contracting data.

E.g. when a publisher is not the original source of information, what should they do to indicate this. 

This might apply to platforms like OpenOpps.com which are aggregating from many sources.

timgdavies:
See also #325 which handles related issues for cases of systems republishing OCDS from multiple sources.

jpmckinney:
1. I recommend that the `publisher` be part of each release in #325, to make it clear who published the release. Releases should be treated as immutable. Re-publishers can add their own metadata in the package – though, what is a use case that relies on information about the re-publisher?
1. The relevant comment from #130 is https://github.com/open-contracting/standard/issues/130#issuecomment-59898216. However, I think the authority that registered the OCID prefix should be the only publisher authorized to either mint new OCIDs or delegate such responsibility. Others should not be minting new OCIDs for which they have no authority. If delegated, all minters should agree on a methodology for what suffix to use https://github.com/open-contracting/standard/issues/461#issuecomment-304083078
1. Further to (2), we should clarify that OCID prefixes establish an identifier series, for which there is one publisher with the authority to mint OCIDs (and to allow others to do so). Anyone can publish a release using an existing OCID. Having `publisher` on the release as in (1) would allow users to exclude publishers they don't trust.

This can be written up on a new guidance page, and also integrated into the OCID prefixes page.

yolile:
Observatorio Fiscal (a CSO) from Chile is adding more information to the existing Chile Compra OCDS publication with data from the Ministry of Works. For processes that already exist in Chile Compra, they are adding new releases (planning and contract implementation in this case), using the same ocid as Chile Compra, and using the `release/publisher` object to identify each release's publisher. For contracting processes that are not in Chile Compra, they are publishing their information using a new ocid and prefix created for them.

duncandewhurst:
Some content related to OCID assignment from https://github.com/open-contracting/standard/pull/1646#discussion_r1363187686:

[...]

PR #1646, opened 2023-10-10
Add guidance on linking releases from different publications

### LINKED ISSUE #1430: Add guidance on linking releases from different publications
We should provide guidance since this issue has come up a couple of times recently

* In the UK, where Contracts Finder, Find a Tender Service and Public Contracts Scotland all publish releases about the same contracting processes (see CRM-7250, CRM-7262, CRM-7304 and CRM-7512)
* In Colombia, where Colombia Compra and ANI publish releases about the same contracting processes (see CRM-7263)

Unless the publications use consistent mappings, reusing the same OCIDs across publications could result in unpredictable and nonsensical results. For example, unless they use an identical approach to assigning local identifiers to objects in arrays, then awards, contracts etc. will overwrite and/or duplicate each other in nonsensical ways.

Roughly, the guidance can be:

* If the publications are distinct (e.g. they cover disjoint sets of fields, different stages, different procedure types etc.) then they should use the same OCID so that users can merge releases.
* If the publications use a consistent mapping, including the approach to assigning local identifiers to objects in arrays, they should use the same OCID so that users can merge releases.
* If the publications don't use a consistent mapping, they should use the `Links` field (added in https://github.com/open-contracting/standard/issues/928) so that users can identify that contracting processes are the same, whilst avoiding incorrect merging of releases.

We can also include some of the content from CRM-7250 around the governance and technical considerations for coordinating publications.

cc @yolile

duncandewhurst:
* Closes #1430


duncandewhurst:
I suggest reviewing the [HTML version of the new page](https://standard.open-contracting.org/staging/1430-linking-releases-guidance/en/guidance/map/connecting_publications/) as the examples are easier to parse that way.

docs/guidance/map/connecting_publications.md
```diff
@@ -0,0 +1,429 @@
...
+    }
+  ]
+}
+```
+````
+`````
+
+## How to connect releases from different publications without reusing ocids
+
+If publications have overlapping fields and inconsistent mappings, then you ought to use [links](../../schema/reference.md#link) to connect releases across the publications. If, like FTS, a publication's releases are derived from another publication's data, the derivative publication ought to use the 'canonical' [link relation type](../../schema/codelists.md#link-relation-type) to refer to the releases in the authoritative publication. If a publication's releases follow another publication's releases, the subsequent publication ought to use the 'prev' link relation type to refer to the earlier releases.
```
odscjen: Just one thought here. In the case of 'canonical' releases it's clear which publisher should include the `links` object. However in the case of 'prev' releases it's less clear. I can imagine the following sort of case (in chronological order):

1. publisher A publishes a tender release
2. publisher B then publishes a tender release for the same process and `links` to publisher A's release as 'prev'
3. publisher A then publishes an award release for the same process

Does publisher A now need to reference publisher B's release as it was published before publisher A's award release? Maybe I'm over thinking this but from this advice I think if I was a publisher this is something I might not be clear on.

odscjen (commented):
Mostly all good just the one comment on when to use the link with a 'prev' relationship

docs/guidance/map/connecting_publications.md
duncandewhurst: In that scenario, I would expect publisher B's tender release to refer to publisher A's original tender release using the 'canonical' link relation type. I don't think publisher's A's award release would need to link to publisher B's tender release as that would require publisher A's system to be aware of publisher B's system.

I'll leave this comment thread open in case @jpmckinney thinks differently.

duncandewhurst:
@jpmckinney let me know if you'd prefer to review in a Google Doc.

docs/guidance/map/connecting_publications.md
jpmckinney: I haven't read the PR yet, but the 'prev' code was described in #928 as:

> The linked resource is a previous publication about this contracting process, that is not in OCDS format or is using a different OCID. This code is intended for use when transitioning from an old system without converting its data, which might occur in cases of jurisdictional changes like municipal mergers.

#1326 landed on the following, based on https://www.iana.org/assignments/link-relations/link-relations.xhtml:

> Indicates that a release or compiled release is a part of a series, and that the previous in the series is the link target.

That description does allow for extremely complicated implementations – but, really, it was designed for only the simple case where a jurisdiction has transitioned from an old system to a new system, and it needs to link to the previous "release" of information about the contracting process.

As I said, I haven't read the PR. If the PR doesn't make this specific use case clear (or describes broader uses of 'prev'), then it should be simplified to only describe this use case for which we have evidence.

docs/guidance/map.md
```diff
@@ -94,6 +94,7 @@ Mapping data to OCDS is not always obvious. Please refer to our how-to guides an
 :maxdepth: 2
 :titlesonly:
 
+map/connecting_publications
```
jpmckinney: I don't think this is the first "hard case" that readers should see – more likely the last?

docs/guidance/map/connecting_publications.md
```diff
@@ -0,0 +1,429 @@
+```{workedexample} Connecting releases from different publications
+:tags: release
+```
+
+# Connecting releases from different publications
+
+Different OCDS publications can contain releases describing the same contracting process. In OCDS, each contracting or planning process is assigned a globally unique identifier, the [ocid](../../schema/identifiers.md#open-contracting-process-identifier-ocid). Ideally, releases describing the same contracting or planning process in different publications would reuse the same ocid so that users can [merge releases](../../schema/merging.md).
```
jpmckinney: I think this style rule is out of date:

> “ocid”, not “OCID”. Although abbreviations in prose are uppercase, the mixing of upper- and lowercase in the documentation (when referring to the concept versus the field) may cause confusion.

In the Primer we went with OCID. When referring to the field, we use backticks. When we're discussing the concept, we use the acronym. I think this is a solution that reads better.

I think similar changes can be made in this PR.

jpmckinney: We also use "contracting (or planning) process" unless this noun phrase appears in parentheses (because double parentheses are hard to read).

docs/guidance/map/connecting_publications.md
```diff
@@ -0,0 +1,429 @@
...
+
+The examples on this page are based on Scotland's Public Contracts Scotland (PCS) publication and the UK's Find a Tender Service (FTS) publication, both of which include releases about the same contracting processes, covering the same contracting process stages and having many fields in common.
+
+## Problems with reusing ocids across publications
+
+Unless publications have disjoint coverage of OCDS fields or implement consistent mappings, reusing the ocids across publications can result in unpredictable and nonsensical results when releases are merged to create a compiled release. For example, unless publications use an identical approach to assigning [local identifiers](../../schema/identifiers.md#local-identifiers) to objects in arrays, then awards, contracts and other objects in arrays can overwrite and/or duplicate each other in nonsensical ways.
+
+### Example: Reusing ocids across Public Contracts Scotland and Find a Tender Service
+
+The PCS and FTS publications have many overlapping fields, but implement slightly different mappings. Therefore, if FTS were to reuse the ocids minted by PCS, merging releases would result in nonsensical data
```
jpmckinney: We've used "minted" in issues, but I think "assigned" is better for non-native speakers.

docs/guidance/map/connecting_publications.md
```diff
@@ -0,0 +1,429 @@
...
+
+### How to reuse ocids across publications
+
+There are two scenarios in which reusing ocids across publications might be possible: publications with distinct coverage of OCDS fields and publications that implement consistent OCDS mappings.
+
+### Publications with distinct coverage
+
+If two publications cover disjoint sets of OCDS fields, with the exception of the release-level required fields (`ocid`, `id`, `date` and `tag`), then the publications can reuse ocids so that users can merge releases.
+
+The publishers would need to agree the process and responsibilities for minting ocids. For example: who can mint ocids; if an ocid cannot be calculated deterministically from the input data, then how to lookup whether an ocid for a given process already exists, and how to share the new ocid for others to find. They would also need to implement the agreed approach, whether this means reporting a new ocid to a central registry, or requesting the ocid for a given procedure from a shared "ocid issuer" service, etc.
```
jpmckinney: It might be simpler to discuss the assignment of new OCIDs as a separate subject. In the PR for `identifiers.md`, it's clarified that there's a delegation process. It might be too much to cover that in this example.

docs/guidance/map/connecting_publications.md
```diff
@@ -0,0 +1,429 @@
...
+        "currency": "GBP"
+      }
+    }
+  ]
+}
+```
+````
+`````
+
+However, this scenario is unlikely to occur since a publication that includes the `awards` array ought to also include details of the suppliers in the `parties` array. In which case, in order to avoid the risk of clashing identifiers, the publications would need to coordinate their approaches to assigning local identifiers to objects in the `parties` array.
```
jpmckinney: This paragraph is kind of saying  "gotcha!". Should we either:

- State upfront that they'll need to coordinate party IDs, before the example
- Use an example that requires no coordination (e.g. planning and tender)

jpmckinney (commented):
I reviewed at a mid-level. That said, I don't expect I'll have much more when I read more slowly.

I haven't reviewed the Colombia case from the issue. Which example to they fall into?

docs/guidance/map/connecting_publications.md
jpmckinney: Not sure how much we want to cover here (there are already many examples), versus in a solution to https://github.com/open-contracting/standard/issues/364

duncandewhurst:
> I haven't reviewed the Colombia case from the issue. Which example to they fall into?

The CRM issue was closed before we learned whether ANI and Colombia Compra use consistent mappings, but that seems very unlikely so I think they fall under "how to connect releases from different publications without reusing ocids".

docs/guidance/map.md
duncandewhurst: Good point. Done in 4d97e3b78c7b045e7fdd65b0d5955c9474b7cf9b

docs/guidance/map/connecting_publications.md
duncandewhurst: Upper case OCID sounds good. Done in e4312363c6d94793a232f4bb7af46f18f03c04cb

Second change made in 766a6373542387dfb01b5863232ef1e7757ba52d

docs/guidance/map/connecting_publications.md
duncandewhurst: Done in 1d2ac3fd1c7b08715083f49dcf4f68b3a9ec1dc5

docs/guidance/map/connecting_publications.md
duncandewhurst: Haha, it is a bit like that!

If they need to coordinate party IDs, then it'll become like the next example (publications with consistent mappings of shared fields) so I don't think we should do that.

I was struggling to come up with a realistic example for publications with disjoint coverage, but I had misremembered that `planning` and `tender` could no longer co-exist in the same process. On re-reading the [contracting processes and planning processes example](https://standard.open-contracting.org/staging/1430-linking-releases-guidance/en/guidance/map/contracting_planning_processes/), I see that `planning` and `tender` can co-exist in the same planning process so I think we have two options for the 'publications with distinct coverage' example:

1. Update it to be about a planning process with two publishers: FTS publishes details of planned tenders in `tender` and HM Treasury publishes budget data in `planning.budget`. I'm still not sure this is a very realistic example
2. Replace it with a sentence like "However, this scenario is unlikely to occur since most publications are likely to have some fields in common, even when publishing data about different stages of the contracting (or planning) process."

duncandewhurst: Hmm, perhaps the example given in https://github.com/open-contracting/standard/issues/364#issuecomment-1182420032 would be a better one to use, though we might need to adapt it to use the same context as the other examples on the page.

docs/guidance/map/connecting_publications.md
duncandewhurst: Yes, the second paragraph does feel very information-heavy. I think the information probably fits better with the solution to #364, although that it is of broader scope than simply republishing.

docs/guidance/map/connecting_publications.md
jpmckinney: @yolile Since you shared the example in the cited comment, do you have an opinion on what example is best here? 

We're looking for an example where two publications cover distinct fields, such that they can reuse OCIDs without e.g. coordination around IDs values in arrays.

If we can't think of a good example, I think we can put the "However," caveat in a note admonition at the start of the example, and encourage people to read the next example.

yolile: From my example, there are some cases where the Chile Compra publication indeed doesn't have the contract data, because the buyers fill out that information using the Ministry of Works systems and not Chile Compra. There are other cases where contract info exists in both and they need to keep IDs in sync. But I guess we could use the first case for the example purposes. They are not publishing this yet (they should by the end of this month) but I think I  could find some test data if it is useful

duncandewhurst: @yolile Assuming that Chile Compra has the award data, I would expect that there would still be some overlapping fields in the first case, e.g. `Contracts.awardID` and supplier data in `parties`, so it would be good to see some test data and to understand any overlapping fields are coordinated.

duncandewhurst: @yolile just checking that you saw my previous comment - if you can share test data, it would be good to see it in order to understand a little more about the Chile example. If not, we can go with the approach suggested by James.

yolile: Ah sorry, @duncandewhurst this one was lost in the multiple updates I'm getting from this and OC4IDS repos :) 

Now we have examples from a test server, for example: https://obrapublica.azurewebsites.net/Contratos/StandarOCDS?Id=346979 (you can check the publisher field in each release to check whether the releases are generated by Chile Compra ("Dirección de Compras y Contratación Pública") or by the Ministry of Public Works ("Observatorio del Gasto Fiscal - Dirección General de Obras Públicas"))

Let me know if this helps.

duncandewhurst: No worries :-) I know that feeling!

The example is very helpful, but as I expected there are overlapping fields and it demonstrates coordination since the values of the fields are the same. For example, the award release from Chile Compra and the contract release from the Ministry of Public Works list the same organizations in the `parties` array ([notebook](https://colab.research.google.com/drive/10Woxkm4vN2p23dSCrj0xaqjxcPHiPdn3#scrollTo=E9wK0k6QAfLd)).

As such, I'll go with James's suggestion from https://github.com/open-contracting/standard/pull/1646#discussion_r1381176507

docs/guidance/map/connecting_publications.md
duncandewhurst: I've removed the content in 413c39012e15236ef8bb347b3b41df947e6e4c71 and added it to #364

docs/guidance/map/connecting_publications.md
duncandewhurst: Done in https://github.com/open-contracting/standard/pull/1646/commits/413c39012e15236ef8bb347b3b41df947e6e4c71

docs/guidance/map/connecting_publications.md
duncandewhurst: @jpmckinney apologies, I had somehow missed replying to this comment. 

The PR doesn't describe the specific use case of referring back to a non-OCDS publication because the scope of #1430 is about linking releases from different OCDS publications.

It has the following to say about 'prev':

> If a publication's releases follow another publication's releases, the subsequent publication ought to use the 'prev' link relation type to refer to the earlier releases.

Given the above and having reviewed the PR, does your comment still stand?

If so, I guess we could broaden the scope of the page and a add section along the lines of "How to refer to a non-OCDS publication".

jpmckinney: Okay, I'll review in full.

duncandewhurst:
@jpmckinney I think this PR is awaiting your review per your comment in https://github.com/open-contracting/standard/pull/1646#discussion_r1408548980. I'll re-request a review so that it is marked as such.

docs/guidance/map/connecting_publications.md
jpmckinney: I've reviewed, and I just added a note admonition about the use of 'prev' in the context of non-OCDS admonitions.

jpmckinney (approved):

PR #1655, opened 2023-10-23
Update identifier reference

### LINKED ISSUE #1524: Clarify reference on identifiers and registered prefixes
From https://github.com/open-contracting/standard/issues/1409

@JachymHercher: 

> I don't understand the following sentence in https://standard.open-contracting.org/staging/1.2-dev/en/schema/identifiers/#registered-prefixes "A prefix is assigned to each organization that holds the existing internal identifier for a Contracting Processes.". Prefixes are assigned to publishers, not buyers, no? How is that linked to whether an organization "holds" (i.e. knows? sets?) an "internal identifier" (why not a public one?)

@jpmckinney 

> Gosh that section is badly worded. https://standard.open-contracting.org/staging/1.2-dev/en/guidance/build/#register-an-ocid-prefix is where we refer to the prefix in the guidance. The only complexity around prefixes is deciding whether a given publisher can use the same prefix as another. There is some public discussion in https://github.com/open-contracting/standard/issues/364#issuecomment-445893561. In short, other publishers need to coordinate with the original registrant if they want to use the same prefix.

Let's reformulate. (We might want to wait for clarifications on purpose of the the whole page in https://github.com/open-contracting/standard/issues/1523 though.)

### LINKED ISSUE #1595: Rewrite examples on Identifiers page to use a common thread, remove types of identifiers section
We can consider linking the examples on [this page](https://standard.open-contracting.org/latest/en/schema/identifiers/) together to build on one another, i.e. instead of Anytown, Mexico, and the UK, we choose one scenario/context and use it across all three.

### LINKED ISSUE #1653: Merge "Publisher namespace" section into instructions on selecting internal identifiers
Follow-up to #1524.

Right now there is mysterious language of:

> **Earlier versions of this documentation imposed a stricter pattern on how internal identifiers ought to be combined with the ocid prefix, including a requirement for local namespaces. This requirement has been relaxed in practice and can be considered deprecated.**
>
> However, publishers are encouraged to consider whether there are any risks of clashes in local identifiers (e.g. the possibility that two parts of the publishing body might use the same identifier for different contracting or planning processes) and to plan to mitigate this when establishing their own patterns to generate their `ocid`

This language was introduced in 2016 in [this commit](https://github.com/open-contracting/standard/commit/42b42089ca968e004359c41cd5a743f7cf0b051f#diff-bfcc5e7c96c1fdb89762c132ef58874889f3e48df7039fb88ee8a56da0cf973d), without a PR or issue or any discussion that I can discover. It appeared in [1.0.1](https://github.com/open-contracting/standard/blob/1__0__1/standard/docs/en/schema/identifiers.md). There is no changelog entry, as 1.0.1 didn't have any entries until 1.1.2 (presumably, the very large commit with the message "Updated documentation structure" made it difficult to discover this change).

The original language was drafted in #73 and [appeared in 1.0.0](https://github.com/open-contracting/standard/blob/1__0__0/standard/docs/en/02_key_concepts/04_identifiers.md#publisher-namespace).

The deprecated content is so ancient – and has only ever been indirectly referenced – that I think we can pretend it never existed and just redraft the second paragraph into the guidance on this page on selecting identifiers.

----

Noting that this was also added in that commit:

> The registered prefixes are dumb identifiers. They are not intended to carry any semantics, and their sole purpose is to turn internal identifiers into globally unique identifiers which can be cross-referenced between systems.

duncandewhurst:
* Closes #1524 
* Closes #1653
* Closes #1595


docs/schema/identifiers.md
```diff
@@ -6,105 +6,107 @@ Consistent identifiers are essential to help join up open contracting data.
...
+1. An internal identifier for the contracting (or planning) process, which is unique within the series
 
-You can read more about the OCDS approach to identify organizations below.
+It is encouraged to separate the OCID prefix and the internal identifier with a hyphen (`-`).
 
-### Local identifiers
+The `ocid` is case-sensitive; in other words, the letter case of an ocid must be consistent.
 
-Not all the identifiers in OCDS need to be globally unique. Most only need to be unique among the identifiers used for the same type of object within the same scope. For example:
+To assign an `ocid` to a contracting (or planning) process, you need to register an OCID prefix and choose an internal identifier.
```
odscjen: Add a link to the guidance on registering an OCID prefix to "register an OCID prefix", or to the section later in the page "OCID prefix"

odscjen (changes requested):
Just one small change

docs/schema/identifiers.md
duncandewhurst: Done in https://github.com/open-contracting/standard/pull/1655/commits/eda23e89bd4de753a8d977ca767dd3dcde40e591

docs/history/changelog.md
```diff
@@ -306,6 +306,7 @@ Per the [normative and non-normative content and changes policy](../governance/n
 * [#1618](https://github.com/open-contracting/standard/pull/1618) Remove validator and application conformance rules.
 * [#1618](https://github.com/open-contracting/standard/pull/1618) Move governance policies from Google Docs, updating references for OCDS 1.1.5 and OCDS 1.2.0, and removing references to GitHub issues.
 * [#1643](https://github.com/open-contracting/standard/pull/1643) Update identifier section in release reference.
+* [#1655](https://github.com/open-contracting/standard/pull/1655) Rewrite identifiers reference and examples.
```
```suggestion
* [#1655](https://github.com/open-contracting/standard/pull/1655) Rewrite identifiers reference and examples for clarity.
```

docs/schema/identifiers.md
```diff
@@ -6,105 +6,107 @@ Consistent identifiers are essential to help join up open contracting data.
...
-### Local identifiers
+The `ocid` is case-sensitive; in other words, the letter case of an ocid must be consistent.
 
-Not all the identifiers in OCDS need to be globally unique. Most only need to be unique among the identifiers used for the same type of object within the same scope. For example:
+To assign an `ocid` to a contracting (or planning) process, you need to [register an OCID prefix](../guidance/build.md#register-an-ocid-prefix) and choose an internal identifier.
 
-* A release ID must be unique within the scope of the process of which it is a part;
-* Award and contract identifiers must be unique within the scope of the process of which they are a part;
-* An item, milestone or document ID must be unique within the array it is part of.
+````{admonition} Worked Example
```
jpmckinney: Change in all cases to:

```suggestion
````{admonition} Example
```

docs/schema/identifiers.md
```diff
@@ -6,105 +6,107 @@ Consistent identifiers are essential to help join up open contracting data.
...
+:class: hint
 
-Local identifiers must be used consistently. For example, if the `id` of an award is "22" in one release, then the `id` of the same award in another release must also be "22".
+Two publishers, the UK Atomic Energy Authority and Health Canada, use sequential numbers as internal identifiers for their contracting processes.
 
-## Open contracting process identifier (ocid)
+The UK Atomic Energy Authority initiates a contracting process to purchase productivity software and assigns it the internal identifier '0005'. Health Canada initiates a contracting process to purchase office furniture and also assigns it the internal identifier '0005'.
```
jpmckinney: Use double quotes instead of single quotes in all examples. Only use single quotes for codes.

docs/schema/identifiers.md
```diff
@@ -6,105 +6,107 @@ Consistent identifiers are essential to help join up open contracting data.
...
 
-<img src="../../_static/svg/green_compilation.svg" width="150" align="right"/>
+To create a globally unique `ocid`, each publisher prepends their internal identifier with their OCID prefix: 'ocds-fh349f' for the UK Atomic Energy Authority and 'ocds-twb234 for Health Canada.
```
jpmckinney: Missing a closing quote on the second prefix.

Also, these OCID prefixes don’t actually exist. Maybe use real publishers, with corresponding prefixes?

docs/schema/identifiers.md
```diff
@@ -6,105 +6,107 @@ Consistent identifiers are essential to help join up open contracting data.
...
+  },
+  "tender": {
+    "id": "0005",
+    "title": "Office furniture"
+  }
+}
+```
 
-In Mexico City, each time a tender or direct contract award process is initiated, the department responsible assigns an identifier.
+As a result, users and tools that work with data from both publishers are less likely to confuse the two contracting processes. If another publisher, such as an independent monitor, wants to publish more data about one of the contracting processes, they can reference it using its unique `ocid`.
```
jpmckinney: Removed content that overlaps with #364

```suggestion
As such, users and tools that work with both publishers' data will not confuse the two contracting processes.
```

The original content was:

> And, if an independent civil society contract monitoring group want to publish a report about implementation of Town A's road project, or Town B's text-book procurement, they have distinct identifiers they can use in their own data to refer to these.

But this is about references from a report or dataset, not from an OCDS dataset specifically.

docs/schema/identifiers.md
```diff
@@ -119,23 +121,29 @@ There are two parts to expressing an **organization identifier** in open contrac
...
-  "id": "06368740",
-  "uri": "http://data.companieshouse.gov.uk/doc/company/06368740",
-  "legalName": "Development Initiatives Poverty Research Limited"
+  "parties": [
+    {
+      "identifier": {
+        "scheme": "GB-COH",
+        "id": "01624297",
+        "legalName": "MICROSOFT LIMITED",
+        "uri": "http://data.companieshouse.gov.uk/doc/company/01624297"
```
```suggestion
        "uri": "https://data.companieshouse.gov.uk/doc/company/01624297"
```

docs/schema/identifiers.md
```diff
@@ -119,23 +121,29 @@ There are two parts to expressing an **organization identifier** in open contrac
 * A prefix for the organization list (also known as registry or register) from which the identifier is drawn: for example, a company register.
 * The existing identifier for the organization within that list.
 
+In OCDS, the list's prefix is disclosed via the `scheme` field of an identifier object, with the existing identifier in the `id` field. If there is a recognized public URL that uniquely identifies the organization, this can be disclosed via the `uri` field.
+
 ````{admonition} Worked Example
 :class: hint
 
-The prefix for the UK's Companies House is "GB-COH". The organization Development Initiatives has been assigned the company number ‘06368740’ by Companies House. The globally unique organization identifier for Development Initiatives can be expressed as below:
+Companies in the UK are assigned identifiers by [Companies House](https://www.gov.uk/government/organisations/companies-house). MICROSOFT LIMITED's identifier is '01624297' and the prefix for Companies House is "GB-COH". Therefore, MICROSOFT LIMITED's organization identifier can be expressed as follows:
```
```suggestion
[Companies House](https://www.gov.uk/government/organisations/companies-house) registers company information in the United Kingdom. The prefix for Companies House is "GB-COH”. Companies House registered a company with the name "MICROSOFT LIMITED" and the number “01624297”. The company's organization identifier can be expressed as:
```

docs/schema/identifiers.md
duncandewhurst: Done in 36911186b9da46fa4a6aec246405596ae6b966ff

docs/schema/identifiers.md
duncandewhurst: Done in 36911186b9da46fa4a6aec246405596ae6b966ff

docs/schema/identifiers.md
duncandewhurst: Missing quote fixed in 36911186b9da46fa4a6aec246405596ae6b966ff. I also corrected instances of smart quotation marks.

There's some background on the choice of publishers in https://github.com/open-contracting/standard/issues/1595#issuecomment-1752226485, in addition to which I thought that using organisations with the names of their countries in their names would clearly communicate that the publications are distinct.

Do you think that using real prefixes combined with fictional internal identifiers is preferable to using fictional prefixes with fictional identifiers? I figured that combining real and fictional might be confusing for readers familiar with the real publisher.

If you prefer to use a real prefix with a fictional internal identifier, we could replace the UK publisher with Find a Tender Service and the Canada publisher with Buy and Sell.

jpmckinney: Hmm, maybe just change the prefixes to def456 and ghi789?

duncandewhurst: Sounds good. Done in 3931a15ef839e42b5398a67c3bf4eaf7a4b26d8a

jpmckinney (approved):

Issue #1064, opened 2020-09-09
Guidance: Publishing OCDS in your language


romifz:
Group existing content and add instructions on flattening commands and in-file translations (see also https://github.com/open-contracting/standard/issues/229).

----

How to include translations for schema files

I'm not sure if this is the correct repository for the question, please let me know if this does not belong here.

I'm wondering how would be the best way to produce and save translated schema files. I'm asking because we would like to show Vialidad Argentina how to use the Flatten Tool to convert their OCDS data to a tabular format ([CRM issue](https://crm.open-contracting.org/issues/5449#note-41)), and we also know of a potential user for this format for which is important to have Spanish titles if possible. The issue here is that they would need to translate their local extensions to Spanish, to produce a patched version of the OCDS Schema in Spanish, and then use it with the Flatten Tool to get the translated field titles.

What would be the best way to do that? To have a second repository for each extension? Or is there a way for the files to be in a single repository?

jpmckinney:
@romifz For registered extensions, there's an open issue to host the translated schema and codelist files in the Extension Explorer: https://github.com/open-contracting/extension-explorer/issues/62 That approach (like for the standard) uses OCDS Babel to extract strings to translate from those files (just the titles and descriptions), using the `pybabel extract` command. These are then pushed to Transifex for translation. OCDS Babel is then also used to substitute the translated text (by using its `translate` method).

For local extensions, even if we make the above steps simpler with good instructions, it's probably still too much work.

So, I would recommend creating an `es` branch of each local extension, where the titles (and descriptions, if needed) are translated. That way, over time, it will be easy to see if the `es` branch is behind the master branch, so that any changes or fixes can be merged. The new branch would behave the same as the master branch, in terms of referencing it from the `extensions` array, providing it to tools like the Extension Registry Python Package to build an extended schema, etc.

How does that sound?

romifz:
@jpmckinney I think it sounds reasonable. I will prepare a recommendation based on this.

Thanks!

jpmckinney:
Noting that translations of registered extensions can be accessed at URLs like https://extensions.open-contracting.org/es/extensions/bids/master/extension.json

jpmckinney:
Although this can be done in a 1.1.6, I think we will freeze 1.1, and only make changes in 1.2.0.

PR #1665, opened 2023-11-29
Abandon in-file translations, document full-file translation guidance

### LINKED ISSUE #229: Abandon in-file translations, document full-file translation guidance
The [current schema](http://ocds.open-contracting.org/standard/r/1__0__0/en/schema/reference/#language) uses an approach to indicate language based on suffixes to property names. 

``` json
{
    "language": "en",
    "tender": {
        "item": {
                "description":"Software consultancy services",
                "description_es":"Servicios de consultoria en software",
                "description_fr":"Services de conseil en logiciels"
        }
    }
}
```

This was discussed in #21 during the beta, and was chosen on the understanding that we wanted to avoid depth where we could (for easier flat renderings of the the data). However, on reflection, the approach we have adopted may not have been the most appropriate.

In particular, @elf-pavlik has pointed out [here](https://github.com/open-contracting/standard/issues/40#issuecomment-134295223) in #40 that [JSON-LD](http://www.w3.org/TR/json-ld/#string-internationalization) uses language maps, and it would be much easier to create a JSON LD rendering of our data this way. 

``` json
{
    "@context": { "@language": "en" },
    "tender": {
        "item": {
              "description": {
                "en" :"Software consultancy services",
                "es": "Servicios de consultoria en software",
                "fr": "Services de conseil en logiciels"
        }
    }
}
```

Our flattening approach would also just render this to 'description/en', 'description/es' and so-on, which seems fairly intuitive, and our earlier fears about over-use of objects does not appear a major one.

However, to change to language maps would potentially be a backwards incompatible update. 

It might make sense to look at doing it early: but would definitely need wide discussion.

duncandewhurst:
* Closes #229


odscjen (approved):

docs/guidance/map.md
```diff
@@ -110,6 +110,7 @@ map/organization_personal_identifiers
 map/organizational_units
 map/organization_classifications
 map/beneficial_ownership
+map/translations
```
jpmckinney: I think we instead want a paragraph with a link to this page under "Localize OCDS to your context".

docs/schema/reference.md
```diff
@@ -34,35 +34,11 @@ There can be cases where a publisher needs to remove, rather than update, a valu
...
-
-A contract for ‘Software consultancy services’ is published in a release with the default language sent to ‘en’ (the ISO639-1 code for English). The following examples give the description of an item as English, French and Spanish.
-
-**json**
-
-```{jsoninclude} ../examples/language_localization/language.json
-:jsonpointer:
-:expand: tender,item
+```{deprecated} 1.2
+  Use [full-file translations](../guidance/map/translations.md) instead.
```
jpmckinney: This entire section is deprecated, so I think instead delete the "Many publishers..." paragraph and:
```suggestion
OCDS 1.1 allowed data to be published in multiple languages by suffixing a language code to a field name: for example, `title` for the default language and `title_es` for Spanish. OCDS 1.2 uses [full-file translations](../guidance/map/translations), instead.
```

jpmckinney (commented):
I don't think this closes #1064, which wants to cover translations of extensions.

I'll make translations.md copy-edits directly.

We can delete examples/language_localization/language.csv and its language.json

----

I updated manage.py to make these nontranslatable: postalCode, email, telephone, faxNumber, milestone code. These have publisher-specific formats or codelists, or are missing a validation property like `"format": "email"` #881.

Comparing `i18n.csv` to the patternProperties, we've removed deprecated fields (good) and added:

* These 3 address components can indeed be translated
  * Address.locality
  * Address.region
  * Address.streetAddress
* I suppose we just missed these 3 originally:
  * Amendment.description
  * RelatedProcess.title
  * Tender.procurementMethodDetails
* These fields are new:
  * Award.statusDetails
  * Contract.statusDetails
  * Tender.statusDetails
  * Tender.exclusionGrounds

Also, in OCDS 1.0, `Organization.name` was translatable. In OCDS 1.1, `OrganizationReference.name` was not translatable. In OCDS 1.2, both are translatable. In other words, both `buyer/name` and the `.name` of the corresponding entry in the `parties` array are translatable. This is good.

jpmckinney:
I've pushed my edits to the new guidance page, for review.

docs/schema/reference.md
duncandewhurst: Paragraph removed in cb37d9935cab44a24c293647f520554ad6a2ee6d.

docs/guidance/map.md
duncandewhurst: Done in e88f438da8c611674d9cb7a448950d2aa4db6776

duncandewhurst:
> We can delete examples/language_localization/language.csv and its language.json

Done in 37a024addef69ca648d720a0433f1ba68f524ead

The revised guidance page and the changes to `manage.py` look good to me.

jpmckinney (approved):

PR #1669, opened 2024-01-16
release-schema: add tender.expressionOfInterestDeadline, update docs

### LINKED ISSUE #1649: Two-stage processes: Review tenderPeriod and add a field for bid submission deadline
From: https://github.com/open-contracting/standard/issues/1160#issuecomment-1596302881:

> We might need to review the definition of tender/tenderPeriod. In a two-stage process, I think this is usually the deadline for requests to participate (first stage). However, in terms of 'active', I think the relevant deadline is the one for bids (second stage). This second deadline is not actually published in any field, so maybe it's not possible to determine the tender status without a status code.

The description of `tender.tenderPeriod` on `1.2-dev` is clear that the end date is the bid deadline:

> The period when the tender is open for submissions. The end date is the closing date for bid submissions.

Prior to #1353, the semantics were ambiguous ("tender submission" could be interpreted as "bid submissions" or as submissions for the `tender` object, which typically represents the first stage of a multi-stage procedure):

> The period when the tender is open for submissions. The end date is the closing date for tender submissions. 

I think that there are two options:

1. Update the description of `tenderPeriod` to reflect how the field is used in practice, i.e. bid deadline for single-stage processes and first-stage deadline for multi-stage processes, and add a new field for the bid submission deadline. This option is backward compatible but it introduces conditional semantics and requires users to look in two different places for the bid deadline, depending on the type of process.
2. Keep the description of `tenderPeriod` as-is and add a new field for the first-stage submission deadline. I prefer this option, however, I'm not sure if the change already made from "tender submissions" to "bid submissions" was actually backward compatible given how the field is being used in multi-stage procedures.

@jpmckinney what do you think?

odscjen:
closes #1649 (branched the extensions out into a new issue)

found no additional documentation changes other than those already discussed in the issue thread


jpmckinney (approved):
I did some other minor edits. It is easier to read my individual commits.

odscjen:
thanks, the edits look good, as you've approved I'll go ahead and merge

PR #1645, opened 2023-10-09
Simplify, align and clarify related processes

duncandewhurst:
* Closes #1520

PR #1670, opened 2024-01-16
release-schema.json: deprecate Contract.relatedProcesses and 'parent'…

### LINKED ISSUE #1650: Deprecate `contracts.relatedProcesses` field and 'parent' relatedProcess code
Follow-up to #1520 

## Background

The field and code were originally discussed in https://github.com/open-contracting/standard/issues/371 and committed in 1727ae4 and 18de84bfc

#1520 deprecates "parent-to-child" codes. Three parent-to-child codes referred to contracts, rather than contracting processes:

Code | Title | Description
-- | -- | --
subContract | Sub-contract | The related process might result in a sub-contract of this contract.
replacementProcess | Replacement process | The related process might result in the replacement of this contract.
renewalProcess | Renewal process | The related process might result in the renewal of this contract.

#1520 replaces 'replacementProcess' and 'renewalProcess' with "child-to-parent" codes. 'subContract' already had a parent code, 'parent'. These "child-to-parent" codes refer to contracting processes, not individual contracts.

## Usage

The field is used by:

* Argentina's Direccion Nacional de Vialidad, to link contracts to planning processes. It never populates the top-level relatedProcesses, and there is only ever one contract. So, this publisher can just use the top-level field (it seems like it is probably the better mapping).
* Secretaría Ejecutiva del Sistema Estatal Anticorrupción de Aguascalientes (SESEA) to link to parent contracts. This publisher has only 6 contracting processes. The `id` of the relatedProcess seem to always match the `id` of the contract, so this is probably a mapping error.

The 'parent' and 'subContract' codes are used in the top-level `relatedProcesses` array by:

* United Kingdom: Contracts Finder: 'parent' in two contracting processes. However, only the `relationship` and the local `id` (1) is provided, so this information has no value.
* Honduras: Sistema de Información y Seguimiento de Obras y Contratos (SISOCS): 'subContract' in one contracting process. The dataset has 22 contracting processes. The contracting process `ocds-mfx54g-11` has 3 contracts and it refers *forward* to `ocds-mfx54g-13`, which has 2 contracts. It's unclear which contract is a subcontract of which.

<details>

```
rg -z '"relatedProcesses"' /data/storage/exporter_dumps/ > 1520-contracts.relatedProcesses.jsonl
cut -d: -f2- 1520-contracts.relatedProcesses.jsonl | jq -c 'select(.contracts[]?.relatedProcesses) | [.ocid, (.contracts|length), .contracts[].relatedProcesses, .relatedProcesses]'
```

```
rg -z '"replacementProcess"|"renewalProcess"|"subContract"' /data/storage/exporter_dumps/ > 1520-relatedProcess-codes.jsonl
cut -d: -f2- 1520-relatedProcess-codes.jsonl | jq -c 'select(.relatedProcesses or .contracts[]?.relatedProcesses) | [.ocid, .relatedProcesses, .contracts[]?.relatedProcesses]'
```

</details>

## Discussion

In theory, it's possible that two different contracts from one same contracting process are the 'replacement', 'renewal', 'repetition', or 'additional' contracts to original contracts from different contracting processes. Or, that only one of several contracts has a relationship to a 'prior' contracting process. In practice, this hasn't been observed. So, the change of scope from contract to contracting process in #1520 for 'replacement' and 'renewal' is fine.

For the remaining codes:

* planning: It's possible that multiple planning processes yield one contracting process, and that different items were planned separately, such that different contracts containing those items could meaningfully refer to different planning processes. However, if this detail were available, we would want to be able to disclose it without a contract necessarily being concluded.
* unsuccessful: In theory, it's possible that one contracting process attempts to award contracts that were originally attempted in different contracting processes. However, if this detail were available, we would want to be able to disclose it without a contract necessarily being concluded.
* framework: An individual call-off can only ever refer to one contracting process as having established its framework agreement, so this can't be used in `contracts.relatedProcesses`.

This leaves 'parent'. The result of the discussion in #392, and the approach in OCDS for eForms and the subcontracting extension, is to include information about subcontracts as part of the main contract in the original contracting process.

When #371 was discussed, it was maybe anticipated that a different system (e.g. a supplier's procurement system) would disclose information about the contracts they award. In this case, their contracting process would be awarding a contract that would be in a sub-contract relationship to a contract from another contracting process (e.g. in a government system). However, OCDS has not achieved this level of universal adoption (and won't in the foreseeable future), so we don't need to plan for this use case.

## Proposal

Deprecate `contracts.relatedProcesses` field and 'parent' relatedProcess code

odscjen:
… in relatedProcess.csv

closes #1650 

also fixed a mistake left over from https://github.com/open-contracting/standard/pull/1645 that left in references to the code 'successfulProcess' that was changed back to 'prior'

### DIFF
--- a/schema/release-schema.json
+++ b/schema/release-schema.json
@@ -808,7 +808,7 @@
           "title": "Related processes",
           "type": "array",
           "deprecated": {
-            "description": "This field is deprecated in favor of `Tender.relatedProcesses`, `Award.relatedProcesses` and the Subcontracting extension, to reflect the updated modelling that removes the 'parent refers to child' approach to linking related processes, and records subcontracts as part of the main contract in the original contracting process.",
+            "description": "This field is deprecated, because all relevant codes from the relatedProcess codelist are deprecated.",
             "deprecatedVersion": "1.2"
           }
         },
--- a/schema/release-schema.json
+++ b/schema/release-schema.json
@@ -806,7 +806,11 @@
           },
           "description": "The details of related processes: for example, if this process is followed by one or more contracting processes, represented under a separate ocid. This is commonly used to refer to subcontracts and to renewal or replacement processes for this contract.",
           "title": "Related processes",
-          "type": "array"
+          "type": "array",
+          "deprecated": {
+            "description": "This field is deprecated in favor of `Tender.relatedProcesses`, `Award.relatedProcesses` and the Subcontracting extension, to reflect the updated modelling that removes the 'parent refers to child' approach to linking related processes, and records subcontracts as part of the main contract in the original contracting process.",
+            "deprecatedVersion": "1.2"
+          }
         },
         "milestones": {
           "title": "Contract milestones",
--- a/schema/release-schema.json
+++ b/schema/release-schema.json
@@ -244,7 +244,7 @@
             null
           ],
           "deprecated": {
-            "description": "Deprecated because the same information can be provided by filling in the relevant date fields.",
+            "description": "This field is deprecated, because the same information can be provided by filling in the relevant date fields.",
             "deprecatedVersion": "1.2"
           }
         },
@@ -567,7 +567,7 @@
           "codelist": "awardStatus.csv",
           "openCodelist": false,
           "deprecated": {
-            "description": "Deprecated because the same information can be provided by filling in the relevant date fields.",
+            "description": "This field is deprecated, because the same information can be provided by filling in the relevant date fields.",
             "deprecatedVersion": "1.2"
           }
         },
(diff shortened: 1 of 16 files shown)


schema/release-schema.json
```diff
@@ -806,7 +806,11 @@
           },
           "description": "The details of related processes: for example, if this process is followed by one or more contracting processes, represented under a separate ocid. This is commonly used to refer to subcontracts and to renewal or replacement processes for this contract.",
           "title": "Related processes",
-          "type": "array"
+          "type": "array",
+          "deprecated": {
+            "description": "This field is deprecated in favor of `Tender.relatedProcesses`, `Award.relatedProcesses` and the Subcontracting extension, to reflect the updated modelling that removes the 'parent refers to child' approach to linking related processes, and records subcontracts as part of the main contract in the original contracting process.",
```
jpmckinney: This doesn't look correct... There are no `relatedProcesses` fields in Tender or Award.

I'll make a commit to fix this and other smaller edits.

jpmckinney (approved):

schema/release-schema.json
odscjen: oops, sorry I must have gotten confused at some point making this PR!

odscjen:
thanks for fixing the errors, I'll merge this now as you've approved it

PR #1671, opened 2024-01-23
framework_agreements.md: add guidance to linking second stage to firs…

### LINKED ISSUE #29: Lots: Framework agreement call-offs with lots
From discussion with @tomasposepny in Georgia:

Framework agreements are often divided into lots and mini-competitions for call-offs from the framework relate to a particular lot.

Based on the [framework guidance](http://standard.open-contracting.org/latest/en/implementation/related_processes/#framework-types), this would be represented in OCDS as multiple contracting processes using [relatedProcess](http://standard.open-contracting.org/latest/en/schema/reference/#relatedprocess) to link the processes for each mini-competition to the process for the initial setup of the framework.

Currently we lack a way to specify which lot in the framework the mini-competition relates to.

We discussed in Georgia adding a `relatedLot` property to the `relatedProcess` object to capture the `lot/id` from the framework which the mini-competition relates to.

odscjen:
…t stage lots

closes https://github.com/open-contracting/ocds-extensions/issues/29

### DIFF
--- a/docs/examples/frameworks/open_multiple_award.json
+++ b/docs/examples/frameworks/open_multiple_award.json
@@ -68,7 +68,10 @@
           "title": "CM Articulos de Escritorio y Papeleria",
           "scheme": "ocid",
           "identifier": "ocds-70d2nz-2239-4-LP14",
-          "uri": "https://example.org/records/ocds-70d2nz-2239-4-LP14"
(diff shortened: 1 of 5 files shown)


jpmckinney (approved):
Made some small copy-edits.

odscjen:
thanks for the edits, as you've approved it I'll merge this now

PR #1673, opened 2024-02-01
Fix procurementMethod explanation in framework worked example 1

johansenja:
Currently in the [framework documentation](https://standard.open-contracting.org/latest/en/guidance/map/framework_agreements/#example-1-closed-framework-agreement-with-single-buyer-single-supplier-and-without-second-stage-competition), the worked example 1 explanation lists the `procurementMethod` as `selective`, but the example give actually lists the `procurementMethod` as `open`:

Screenshot 2024-02-01 at 15 05 39


johansenja:
It seems like the linting is failing because it's trying to fetch https://raw.githubusercontent.com/open-contracting/standard/patch-1/schema/record-schema.json (which should perhaps be https://raw.githubusercontent.com/johansenja/standard/patch-1/schema/record-schema.json?)

jpmckinney:
Thank you, @johansenja ! Indeed, that test wasn't configured for PRs outside the main organization. Fixed now.

Which OCDS publication are you working on, by the way?

jpmckinney:
The word was changed to selective in https://github.com/open-contracting/standard/pull/1461, but the JSON was not updated to match. That PR fixes https://github.com/open-contracting/standard/issues/1301, but this word change was not discussed there (at least not directly from my quick review).

The preamble to the example has "The framework agreement is not open, since the tender notice does not specify that the procurement involves the setting up of a dynamic purchasing system." This was present since the [first commit](https://github.com/open-contracting/standard/blob/4fe0e123e7175000195c5aa678bee68cdefebff7/docs/guidance/map/related_processes.md) of this guidance, though the bullet was different:

> * Since the framework agreement will be concluded with a single supplier and since any supplier is able to submit a response to the invitation to participate, we set the `procurementMethod` to ‘open’.

The wording was changed in https://github.com/open-contracting/standard/commit/058a9ea8593ae8805f17149681b1db2015c2a573 to:

> * Since any supplier is able to submit a response to the invitation to participate, `tender.procurementMethod` is set to 'selective'.

I think it should be 'selective' – but we can maybe fix the bullet point for clarity.

johansenja:
Ok, thanks! I've updated the json now and reverted the bullet point, though it feels to me like the bullet could still use a bit of clarification.

> Since any supplier is able to submit a response to the invitation to participate, `tender.procurementMethod` is set to 'selective'

"Any supplier" gives me the notion that the submission process is completely free/open, whereas "selective" suggests a degree of choosiness which could prohibit certain suppliers from applying.
 
Screenshot 2024-02-13 at 14 06 48

Though I'm not sure if I'm well-versed enough in the differences between `open` and `selective` to make an accurate or valuable suggestion!

jpmckinney:
'selective' has a very specific meaning in procurement. The "request to participate" in the first step in a two-step process. In this first step, the buyer determines which suppliers are qualified and which are not. This step is "selection" or "qualification", in which "selection criteria" (and exclusion grounds) are used to determine which suppliers are qualified. In the second step, only qualified suppliers can bid.

johansenja:
Ok thanks for the clarification, that's helpful. Do you think something like this captures the sense any better?
> Since requests to participate are open to all suppliers prior to a selection process conducted by the buyer, `tender.procurementMethod` is set to 'selective'

I'm trying to capture in equal measures the fact that the process is open to all suppliers, but the sense of where the "selection" comes in (through the buyer applying the selection criteria) - while not being too wordy.

Happy to update the PR again if it's worth it!

jpmckinney:
Noting that #1680 edits this page, so I'll review this PR after merging that other one.

jpmckinney:
#1680 corrects the JSON to 'selective' and changes the wording to:

> Although any supplier can submit a request to participate, only qualified suppliers can submit a proposal, so `tender.procurementMethod` is set to 'selective'.

So, I think that will also resolve this PR.

Issue #1571, opened 2022-12-06
Extensions: Consider normative content about how extensions are published


jpmckinney:
We presently assume that extensions are published in a web-accessible directory (or API that behaves like a directory), such that `extension.json` can be trimmed from the end of a URL and replaced with a schema or codelist file.

Some publishers publish via APIs (e.g. Visual Studio) where the file path is a query string parameter.

The available documentation on creating extensions is:

* https://standard.open-contracting.org/latest/en/guidance/map/extensions/
* https://github.com/open-contracting/standard_extension_template
* https://extensions.open-contracting.org/en/publishers/

Options:

1. Add normative content establishing our assumption as the rule.
1. Change the publication format to be a ZIP file (GitHub, GitLab, Visual Studio, etc. auto-generate a ZIP). The `extensions` array would also have to be links to ZIP files.
1. Change the `schemas` and `codelists` keys in `extension.json`) to be absolute URLs and/or relative URLs.
1. Relax our assumption so that just replacing `extension.json` in the URL with the desired path should work.

Considerations:

1. This will be fine for the vast majority of publishers, but might frustrate the 1-2 exceptions.
2. This will require a change to all publications that use extensions. Otherwise, it should be fine for all publishers, as all known extensions (as far as I remember) are published on platforms that provide ZIP files automatically.
3. This will require a change to all publications that use extensions. This also has its own complexities, e.g. resolving relative URLs, testing extensions (absolute URLs are likely to change between development and production), etc.
4. This should be fine for all publishers. A challenge is that the rule might be hard to word. We can maybe word it the same as (1), and then add the substitution case as an alternative.

jpmckinney:
Closed #1291 as duplicate. (Please read in case anything additional to cover.)

#606 is also related and can maybe be closed at the same time.

odscjen:
Considering all of those options:

1. Checking through the publishers in the Data Registry who aren't marked as no longer publishing, 42/56 use OCP maintained extensions, and 25/56 use local extensions. Of the local extensions that are linked to from the Data Registry pages **all but one** are hosted on Github, Gitlab or Bitbucket and so meet the need that  
>    `extension.json` can be trimmed from the end of a URL and replaced with a schema or codelist file.

2. Given how many publishers use extensions (local or OCP maintained, see above) I don't think this is a desirable option.

3.  as for 2

[...]

PR #1682, opened 2024-03-29
conformance_and_extensions.md: add normative guidance about extension…

odscjen:
… paths

partially addresses #1571 

The other page in this repo that references Extensions is https://standard.open-contracting.org/latest/en/guidance/map/extensions/ but all it says is of relevant here is "You can find the [extension template](https://github.com/open-contracting/standard_extension_template) on GitHub. It contains guidance on creating an extension." and while the guidance will be moved there will still be a link to it from the extension template so I don't think we need to alter this sentence.

This PR does not close the linked issue as it still requires updating the extension_explorer repo

### DIFF
--- a/.github/workflows/shell.yml
+++ b/.github/workflows/shell.yml
@@ -6,10 +6,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - name: Install shellcheck
-        run: |
+      - run: |
           sudo apt update
-          sudo apt install shellcheck
-      - run: sudo snap install shfmt
+          sudo apt install shellcheck shfmt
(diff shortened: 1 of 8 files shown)


jpmckinney:
I added the extension template's two issues to the OCDS 1.2 project, so I think this would now close #1571, correct?

jpmckinney (approved):
I added a few other unrelated commits, instead of making an entire new PR.

odscjen:
those extra commits all look good, I'll go ahead and merge this

odscjen:
@jpmckinney hmm, despite the tests passing in the PR when I merged it's created a lint error, I'm not sure how to fix that!

jpmckinney:
Don’t worry, it’s a false positive

Issue #330, opened 2016-05-18
Update release schema to be coherent with merging specification


kindly:
As mentioned in #284 a compiledRelease in an OCDS record has the same schema as a standard OCDS release.
However, some of the fields, most notably the release ID and date, make no sense once compiled and they are also mandatory.  This means it is not possible to make a compiled release that makes sense and validates.

The best approach would be to make a new sightly modified version of the OCDS release to take this into account.  Naming possibly merged-release-scema.json

duncandewhurst:
It's also worth noting that whilst the [merging rules guidance](http://standard.open-contracting.org/latest/en/schema/merging/#merging-rules) states that the `id` and `date` fields should be omitted from a compiled release, the example data on the same page of the docs shows a compiled release with both of these fields included.

We should fix this as when we update the docs as part of the upgrade

Bjwebb:
@kindly Your current code drops `tag` from the versioned release. I'm not quite clear that this is the correct behaviour.

Bjwebb:
Just discussed with @kindly why dropping `tag` does make sense, because every versioned block has the associated releaseTag.

jpmckinney:
The issues that @kindly and @duncandewhurst reported haven't been fixed.

jpmckinney:
Starting to document differences between release-schema.json and proposed compiled-release-schema.json:

* `id` and `date` should be omitted from the schema (backwards-compatible)
* `tag` should be `["compiled"]`, or it should be omitted from the schema, and 'compiled' should be removed from the `releaseTag.csv` codelist. I prefer the latter. Making both those changes at once would be backwards-compatible, as only compiled releases should be using the 'compiled' tag.
* No fields should have `null` values (debatably backwards-incompatible)
* See discussion on identifying (or at minimum dating) compiled releases in #455 
* If `publisher` is added to the release schemas, as in #325, we might want its semantics to be different in the compiled release, i.e. to be the publisher of the compiled release, instead of following the regular merge rules (or, we might want to use a different field name for this).

JachymHercher:
> tag should be ["compiled"], or it should be omitted from the schema, and 'compiled' should be removed from the releaseTag.csv codelist. I prefer the latter. Making both those changes at once would be backwards-compatible, as only compiled releases should be using the 'compiled' tag.

[...]

PR #1705, opened 2024-07-17
compiled-release-schema.json manual draft

odscjen:
related to #330


jpmckinney:
Closing as per latest update to #330 we won't be creating a separate schema.

PR #1708, opened 2024-10-23
docs/schema/identifiers.md: Explain who can register an OCID prefix

### LINKED ISSUE #1325: Clarify "globally unique identifiers" example
Two publishers have recently interpreted the [globally unique identifiers](https://standard.open-contracting.org/latest/en/schema/identifiers/#globally-unique-identifiers) worked example that uses "Two government publishers (Town A and Town B)" to mean that each *procuring entity* should have a different OCDS prefix because it refers to town-level publication.

We should update the example to clarify, e.g. to use "Country A" and "Country B" instead of "Town A" and "Town B".

See [CRM#6950](https://crm.open-contracting.org/issues/6950#note-44) for context.

duncandewhurst:
Closes #1325


duncandewhurst:
@jpmckinney the lint check is failing on a bunch of ruff errors across various Python files. Should I go through and fix them or do you want to do that in a separate PR?

jpmckinney:
I already have them fixed locally. I'll commit when reviewing.

jpmckinney (approved):

jpmckinney:
Btw, you can now `pip install pre-commit` and run `pre-commit install`. The Python and requirements checks will then run if those files are changed.

duncandewhurst:
> Btw, you can now `pip install pre-commit` and run `pre-commit install`. The Python and requirements checks will then run if those files are changed.

Great! So to reproduce the GitHub Actions checks locally, the only things I need to run manually are the tests in the standard repo and `test_csv.py`,  `test_json.py` and `test_readme.py` from standard-maintenance-scripts?

duncandewhurst:
@jpmckinney not sure if you were notified of my comment above

duncandewhurst:
@jpmckinney also, the ruff-format hook makes extensive changes to quotation marks in `.manage.py` (not sure if anything else) when I change that file.

jpmckinney:
Ah, yeah, I forgot to run `ruff format`. We can do that in another PR.

Yes, those are the only tests that are not run automatically by pre-commit.

Issue #582, opened 2017-09-28
Review Extensions conformance section


jpmckinney:
Possible new requirements for extensions:

* All definitions and properties must set a `title`, `description` and `type`, unless they are originally defined in the core schema or in another extension. https://github.com/open-contracting/standard/issues/603
* If `openCodelist` is `true`, `enum` must not be set.
* If `openCodelist` is `false`, `enum` must be set. https://github.com/open-contracting/ocds-extensions/issues/32
* If `openCodelist` is `false` and `enum` is set, `enum` must match the codes in the `codelist`.
* If a field has an `enum`, should we require that it be expressed as a codelist?

We can offer tool support for this: both for validating extensions, and for automatically putting the codes from the codelists into the `enum`. (Both tools now exist.)

http://standard.open-contracting.org/latest/en/schema/conformance_and_extensions/#extensions

jpmckinney:
Possible new requirements (related to https://github.com/open-contracting/standard-maintenance-scripts/issues/48):

* Conformant extensions should not delete fields from the standard's schema.
* Conformant extensions should not change the properties of fields from the standard's schema. Note: api_extension changes `required` to `[]` and ocds_milestone_documents_extension changes `deprecated` to `null`.

This is to avoid an extension changing the semantics of the standard's schema and thereby breaking interoperability.

If an extension desires to further document the usage of a field, it should do so through documentation, rather than through changing a field's `description` property.

jpmckinney:
Possible new requirements (related https://github.com/open-contracting/standard-maintenance-scripts/issues/45):

* Definition names should be UpperCamelCase.
* Field names should be lowerCamelCase.
* Both names should contain only ASCII alphabetical letters.
* If an acronym is used within a name, the acronym should be all uppercase, unless it is at the beginning of the name, in which case it should be all lowercase.

duncandewhurst:
I'm in favour of all of the proposed requirements. However, there is some cross-over with parts of the [schema style guide](https://ocds-standard-development-handbook.readthedocs.io/en/latest/meta/schema_style_guide.html#), so we should consider how to keep those in sync, or whether the style guide should defer to the extensions conformance section for some points.

> If a field has an `enum`, should we require that it be expressed as a codelist?

I think so, so that titles and descriptions for the enum values can be translated.

jpmckinney:
Yes, once this documentation exists, we can remove some content from the style guide and link to these docs.

[...]

PR #1715, opened 2024-11-07
conformance_and_extensions.md:update extension guidance

odscjen:
addresses https://github.com/open-contracting/standard/issues/582 (not sure if it will close it as an additional PR to ensure this is consistent with the style guide may be needed)

* https://standard.open-contracting.org/staging/1.2-dev/en/governance/normative/ states that all .md files in the docs/schema folder are normative so have used normative language throughout (and noticed a few non-normative uses in the rest of the page)
* re. the names rules from https://github.com/open-contracting/standard/issues/582#issuecomment-354357410, I've changed these from **should** to **must** to reflect the wording in the [tests](https://jscc.readthedocs.io/en/latest/api/testing/checks.html)

### DIFF
--- a/docs/schema/conformance_and_extensions.md
+++ b/docs/schema/conformance_and_extensions.md
@@ -27,29 +27,39 @@ The schema for the standard by default allows for new fields, and does not fail 
 
 Conformant extensions should respect the following guidelines:
 
-### Core fields
+### Changes to existing fields
 
-* Extensions *should not* delete fields from the core schema
-* Extensions *should not* change the properties of fields from the core schema. If an extension desires to document further usage of a core standard field, it should do so through documentation, rather than changing the field's `description` property.
+A conformant extension *should not*:
 
-### Structure
+* Delete properties or definitions from the OCDS schema.
+* Change the value of keywords in the OCDS schema. If an extension desires to document further usage of a field in the OCDS, it should do so through documentation, rather than changing the field's `description` keyword.
 
-* All definitions and properties *must* set a `title`, `description` and `type`, unless they are originally defined in the core schema or in another extension in which case they *must* set a `$ref` to the existing object
-* If a field's `type` is "array", `items` *must* be set
-* If using `items`, its `type` *must* only include "array", "number" and/or "string"
-* If an array field's `wholeListMerge` and `omitWhenMerged` properties are not used or are set to `false`, the object fields under it *must* have an `id` field and this `id` field *must* be required
+### New fields
+
+A conformant extension *must*:
+
+* Set the `title`, `description` and `type` keywords for all new definitions and properties
+* Set `$ref` for fields that reuse a definition in the OCDS schema.
+* Set the `items` keyword for properties of `type` "array"
+* Not include types other than "array", "number" and/or "string" in the `types` keyword under an `items` keyword.
+* Ensure that any definition referenced in a property of `type` array, whose `wholeListMerge` and `omitWhenMerged` keywords are unset or set to `false`, has a required `id` property.
 
 ### Codelists
 
-* If `openCodelist` is `true`, `enum` *must not* be set
-* If `openCodelist` is `false`, `enum` *must* be set
-* If a field has an `enum`, this *must* be expressed as a csv codelist
-* If a field has an `enum`, `codelist` and `openCodelist` *must* be set
-* If adding codes to an existing codelist the codelist filename *must* append `+` to the start of the core codelist filename, for example `+documentType.csv`
+A conformant extension *must*:
 
-### Field and code names
+* Set `enum` for properties with `openCodelist` set to `false`
+* Not set `enum` for properties with `openCodelist` set to `true`
+* Set `codelist` and `openCodelist` for properties that set `enum`.
+* Document titles and descriptions for `enum` values as a CSV codelist.
+* Append "+" to the start of the codelist filename when adding codes to an existing codelist, e.g. `+documentType.csv`.
 
-* Definition names *must* be UpperCamelCase
-* Field names *must* be lowerCamelCase
-* Definition and field names *must* contain only ASCII alphabetical letters
-* If an acronym is used within a field or definition name, the acronym *should* be all UPPERCASE, unless it is at the beginning of the name, in which case it *should* be all lowercase.
+### Naming conventions
+
+A conformant extension *must*:
+
+* Use lowerCamelCase for property names, e.g. `tenderPeriod`
+* Use UpperCamelCase for definition names, e.g. `RelatedProcess`
+* Use only ASCII alphabetical letters for property and definition names
+
+A conformant extension *should* use UPPERCASE for acronyms within property or definition names, unless the acronym is at the beginning of the name, in which case it *should* be all lowercase.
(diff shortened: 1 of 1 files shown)


docs/schema/conformance_and_extensions.md
```diff
@@ -13,14 +13,43 @@ To maximize the interoperability of data published using the Open Contracting Da
...
+Extensions to the standard may add new objects and fields to accommodate specific local requirements. An extension must not be created if it is possible to use existing terms from the standard. It must be possible to access an extension's schema and codelist files by replacing `extension.json` in the extension's URL with a file's path, e.g. `release-schema.json` or `codelists/codelist.csv`. For more information, refer to the Extension Explorer's [guidance on documenting extensions](https://extensions.open-contracting.org/en/publishers/#document).
 
 The schema for the standard by default allows for new fields, and does not fail validation of a file which contains unknown fields.
+
+Conformant extensions should respect the following guidelines:
```
duncandewhurst: Prefer singular as in the first list item under the publication conformance heading. I think we should also update that sentence to "A conformant publication..." since this page describes the 'static' rules that a publication must meet to conform to the standard, rather than describing a process of conforming to the standard.

```suggestion
A conformant extension should respect the following guidelines:
```

docs/schema/conformance_and_extensions.md
```diff
@@ -13,14 +13,43 @@ To maximize the interoperability of data published using the Open Contracting Da
...
+* If a field has an `enum`, this *must* be expressed as a csv codelist
+* If a field has an `enum`, `codelist` and `openCodelist` *must* be set
+* If adding codes to an existing codelist the codelist filename *must* append `+` to the start of the core codelist filename, for example `+documentType.csv`
+
+### Field and code names
+
+* Definition names *must* be UpperCamelCase
+* Field names *must* be lowerCamelCase
+* Definition and field names *must* contain only ASCII alphabetical letters
+* If an acronym is used within a field or definition name, the acronym *should* be all UPPERCASE, unless it is at the beginning of the name, in which case it *should* be all lowercase.
```
duncandewhurst: I've taken a pass at rewording to use consistent imperative form and for strict use of the terms 'keyword', 'property' and 'definition' from JSON Schema. The previous draft used 'properties' to mean both properties and keywords in JSON Schema, and often referred to 'fields' and 'definitions', which mixes OCDS terminology with JSON Schema terminology, e.g. `tender.tenderPeriod.startDate` is a field in OCDS, but in JSON Schema `tender`, `tenderPeriod` and `startDate` are properties, and `tender` and `tenderPeriod` reference the `Tender` and `Period` definitions, respectively.

The second bullet point might need to be further refined depending on https://github.com/open-contracting/standard/issues/582#issuecomment-2484286699

```suggestion
### Changes to existing fields

A conformant extension *should not*:

* Delete properties or definitions from the OCDS schema.
* Change the value of keywords in the OCDS schema. If an extension desires to document further usage of a field in the OCDS, it should do so through documentation, rather than changing the field's `description` keyword.

### New fields

A conformant extension *must*:

* Set the `title`, `description` and `type` keywords for all new definitions and properties
* Set `$ref` for fields that reuse a definition in the OCDS schema.
* Set the `items` keyword for properties of `type` "array"
* Not include types other than "array", "number" and/or "string" in the `types` keyword under an `items` keyword.
* Ensure that any definition referenced in a property of `type` array, whose `wholeListMerge` and `omitWhenMerged` keywords are unset or set to `false`, has a required `id` property.

### Codelists

A conformant extension *must*:

* Set `enum` for properties with `openCodelist` set to `false`
* Not set `enum` for properties with `openCodelist` set to `true`
* Set `codelist` and `openCodelist` for properties that set `enum`.
* Document titles and descriptions for `enum` values as a CSV codelist.
* Append "+" to the start of the codelist filename when adding codes to an existing codelist, e.g. `+documentType.csv`.

### Naming conventions

A conformant extension *must*:

* Use lowerCamelCase for property names, e.g. `tenderPeriod`
* Use UpperCamelCase for definition names, e.g. `RelatedProcess`
* Use only ASCII alphabetical letters for property and definition names

A conformant extension *should* use UPPERCASE for acronyms within property or definition names, unless the acronym is at the beginning of the name, in which case it *should* be all lowercase.
```

duncandewhurst (changes requested):

docs/schema/conformance_and_extensions.md
duncandewhurst: If my suggestion below is accepted, I think this sentence can be removed as it's bit confusing to introduce a load of 'must' rules with a 'should'.

duncandewhurst (changes requested):

docs/schema/conformance_and_extensions.md
odscjen: agree about the confusion, but I do think some sort of introduction to the guidelines is needed so I've added a new heading and nested everything that follows under it

docs/schema/conformance_and_extensions.md
```diff
@@ -25,31 +25,41 @@ Extensions to the standard may add new objects and fields to accommodate specifi
 
 The schema for the standard by default allows for new fields, and does not fail validation of a file which contains unknown fields.
 
-Conformant extensions should respect the following guidelines:
+### Extension conformance guidelines
```
duncandewhurst: This section documents rules, not guidelines. Also, come to think of it, we should probably explain the terminology used:

```suggestion
### Extension conformance

This section documents the normative rules that OCDS extensions must conform to. It uses the following terms from JSON Schema to describe the rules for extension schemas:

* [property](https://json-schema.org/understanding-json-schema/reference/object#properties): A key-value pair belonging to an object, defined using the `properties` keyword in the schema or subschema for the object, e.g. `ocid` is a property of an OCDS release and it is defined under the top-level `properties` keyword in the OCDS release schema.
* [definition](https://json-schema.org/understanding-json-schema/structuring#defs): A subschema defined using the top-level `definitions` keyword, e.g. the `Award` definition contains the subschema for an OCDS award and it is defined under the `definitions` keyword in the OCDS release schema. 
* [keyword](https://json-schema.org/understanding-json-schema/keywords): A key within a JSON Schema used to define the structure of JSON data , e.g. the `type` keyword specifies the data type for a property.



```

duncandewhurst (changes requested):
Happy for you to request a review from James, if you accept my suggestion. It's probably worth flagging to him the reasons for using JSON Schema terminology expressed in https://github.com/open-contracting/standard/pull/1715#discussion_r1847567056.

docs/schema/conformance_and_extensions.md
odscjen: unresolved this just so that it's visible for James when he comes to review

jpmckinney:
> an additional PR to ensure this is consistent with the style guide may be needed

Can you describe in more detail?

odscjen:
> > an additional PR to ensure this is consistent with the style guide may be needed
> 
> Can you describe in more detail?

This is in reference to https://github.com/open-contracting/standard/issues/582#issuecomment-860866508 - some of the detail added in this PR is also covered in the style guide, e.g. 2 of the rules in the Naming Conventions section are repeated in https://ocds-standard-development-handbook.readthedocs.io/en/latest/meta/schema_style_guide.html#field-and-code-names, while the third rule doesn't appear to be in that part of the style guide at all.

I've not done a full review to check which bullet points are already in the style guide and which are new and therefore what might need to be done to ensure consistency across the 2 resources. I can do so as part of this PR but I feel it might be easier as a separate issue and PR so we can discuss the best way to do it

Issue #1713, opened 2024-11-05
Update schema reference for Contract.relatedProcesses


duncandewhurst:
The deprecated field is still mentioned in the Markdown content below the jsonschema directive: https://standard.open-contracting.org/staging/1.2-dev/en/schema/reference/#relatedprocess

duncandewhurst:
Per https://github.com/open-contracting/standard/issues/1714#issuecomment-2472264366, also need to explain that contracting processes should be linked to planning processes using the 'planning' code.

PR #1716, opened 2024-11-12
reference.md: add contracting and planning process definitions

### LINKED ISSUE #1714: Feature contracting process and planning process definitions in the reference documentation
Presently, the following definitions from the `description` field of the release schema only appear in the implementation guidance:

> **Contracting process**
> All the actions aimed at implementing one or more contracts. This covers tendering, awarding, contracting and implementation. It does not include actions linked to planning, as these are often less structured and may be linked to multiple contracting processes. In multi-stage procedures (e.g. framework agreements with reopening of competition), each round of competition is treated as a separate contracting process.
>
> Procedures that failed and were restarted are considered new processes.
>
> Boundaries between processes (e.g. whether two contracts result from a single process or from two processes) are set by buyers depending on their needs (e.g. efficient division of labor, clear communication with the market) and legislation (e.g. rules on using procedures and lots).

> **Planning process**
> All the actions aimed at planning one or more contracting processes. This covers, for example, need identification, budget planning, and market research.
>
> Planning processes are often less structured than contracting processes, so one or more planning processes may lead to one or more contracting processes.

The page on which they appear (`docs/guidance/map/contracting_planning_processes.md`) is only expected to be (optionally) read during the mapping phase of an OCDS implementation, but I think it would aid understanding if these definitions were featured prominently in the reference documentation. The primer does include a definition of a contracting process, but I think this kind of detail more properly belongs in the reference section, since it is something that implementers might need to refer back to when thinking about different kinds of contracting process.

odscjen:
closes #1714 

I deliberately included the definitions and nothing else other than an introductory sentence as otherwise it would essentially just be a copy of https://standard.open-contracting.org/staging/1.2-dev/en/guidance/map/contracting_planning_processes/


docs/schema/reference.md
```diff
@@ -14,9 +14,29 @@ Releases must be published within a [release package](packaging/release_package)
 This page presents the release schema in tables, with additional information in paragraphs. You can also download the canonical version of the release schema as [JSON Schema](../../build/current_lang/release-schema.json), download it as a [CSV spreadsheet](../../build/current_lang/release-schema.csv), view it in an [interactive browser](release), or access it through the [Field-Level Mapping Template](https://www.open-contracting.org/resources/ocds-field-level-mapping-template/).
 ```
 
+## Contracting and planning processes
+
+OCDS recognizes two types of processes: contracting processes and planning processes. In OCDS, a given process is uniquely identified by an [open contracting process identifier](identifiers.md#open-contracting-process-identifier-ocid) (`ocid`).
```
duncandewhurst, line 19: I think we can omit the part about the `ocid`. On the worked example page, it serves to recap the purpose of the `ocid` before the later note that discusses using separate `ocid`s for planning and contracting processes. The content added in this PR doesn't mention the `ocid` again so I don't think it needs to be introduced here.

```suggestion
OCDS recognizes two types of processes: contracting processes and planning processes.
```

duncandewhurst (commented):
I've made a few more suggestions in https://github.com/open-contracting/standard/issues/1714#issuecomment-2472264366. Let's hold fire on the PR until we've decided what to do.

docs/schema/reference.md
odscjen: applied in https://github.com/open-contracting/standard/pull/1716/commits/84e7a9ef40e8199e52cbf83006915d20a41395b8

odscjen:
suggestions in #1714 now applied. Haven't updated the `releaseProcess` reference section as I think that requires it's own PR (see #1713)

docs/schema/reference.md
```diff
@@ -16,7 +16,7 @@ This page presents the release schema in tables, with additional information in
 
 ## Contracting and planning processes
 
-OCDS recognizes two types of processes: contracting processes and planning processes. In OCDS, a given process is uniquely identified by an [open contracting process identifier](identifiers.md#open-contracting-process-identifier-ocid) (`ocid`).
+OCDS recognizes two types of processes: contracting processes and planning processes.
```
duncandewhurst: Now that we've integrated the notes about `ocid`s from the worked example, I think we need to reinstate this.

```suggestion
OCDS recognizes two types of processes: contracting processes and planning processes. In OCDS, a given process is uniquely identified by an [open contracting process identifier](identifiers.md#open-contracting-process-identifier-ocid) (`ocid`).
```

schema/release-schema.json
```diff
@@ -33,7 +33,7 @@
...
+      "description": "A tag labeling the release, using the the open [releaseTag](https://standard.open-contracting.org/{{version}}/{{lang}}/schema/codelists/#release-tag) codelist. Tags distinguish, for example, planning and contracting processes and the stages of contracting processes. Codes outside the releaseTag codelist may indicate, for example, the notice or form to which the release corresponds, or the event that triggered the publication of the release. A planning process must use either the 'planning' or 'planningUpdate' code, and must not use any other code from the releaseTag codelist. A contracting process should not use the 'planning' or 'planningUpdate' codes.",
```
duncandewhurst: Fix the "the the" repetition in the tag description.

Did we discuss changing 'might' to 'may' in the sentence about codes from outside the codelist?

I think using a normative keyword is inappropriate because the sentence doesn't set a normative rule about the permitted semantics of codes from outside the codelist, rather it provides some illustrative examples.

```suggestion
      "description": "A tag labeling the release, using the open [releaseTag](https://standard.open-contracting.org/{{version}}/{{lang}}/schema/codelists/#release-tag) codelist. Tags distinguish, for example, planning and contracting processes and the stages of contracting processes. Codes outside the releaseTag codelist might indicate, for example, the notice or form to which the release corresponds, or the event that triggered the publication of the release. A planning process must use either the 'planning' or 'planningUpdate' code, and must not use any other code from the releaseTag codelist. A contracting process should not use the 'planning' or 'planningUpdate' codes.",
```

duncandewhurst (changes requested):
Need to remove the worked example.

schema/release-schema.json
odscjen: I looked and looked and couldn't see the "the the" that was being referred to! thanks for finding it :)

odscjen: > Did we discuss changing 'might' to 'may' in the sentence about codes from outside the codelist?

we hadn't discussed it, I was just going by the guidance that everything in the docs/schema folder is normative therefore we should use normative language. I'll take your suggestion, James can change it back if he doesn't agree

odscjen:
worked example now removed, I'd just been waiting on confirmation in the issue that that was what was agreed upon

duncandewhurst (approved):
Looks good!

> Haven't updated the releaseProcess reference section as I think that requires it's own PR (see https://github.com/open-contracting/standard/issues/1713)

I've added a reminder to #1713

PR #1680, opened 2024-02-28
Make examples minimal

duncandewhurst:
closes #1449 

Closes #1666

Replaces https://github.com/open-contracting/standard/pull/1678

Implements proposed improvements from https://github.com/open-contracting/standard/issues/1449#issuecomment-1849353843

* [x] Replace `1666-make-examples-minimal` with `1.2-dev` in handbook

PR #1660, opened 2023-11-17
remove mention of building blocks and blocks from documentation

odscjen:
closes #1347 

Used primarily "objects" as per the section of the style guide that states 
> When describing JSON schema * “object”, not “block”

but did use "sub-schema" in a few instances.

There's 2 uses of "block" in the schema itself but they'll be covered by a PR for https://github.com/open-contracting/standard/issues/850

PR #1718, opened 2024-11-18
docs/schema/: create individual pages for all codelists

### LINKED ISSUE #1405: Change order of codelists in Reference
When going through OCDS, I often go directly to one of the codelists in https://standard.open-contracting.org/latest/en/schema/codelists/, but then I am unsure whether it is open or closed, so I scroll up or down to find which section it is in. This seems a bit clumsy.
1) Wouldn't it be better if each codelist has an [open] or [closed] attribute below/after it's title?
2) If we added such an attribute, we could also remove the open / closed subsections. I'm not sure how useful they are (do people want to read about "all closed codelists" in one go? Seareching for codelists probably happens via ctrl-f.) and we could then have the same sorting across the whole page (see below).
3) I'm not sure how the codelists are sorted. Doesn't seem to be alphabetical, by importance, nor chronological? I would go by importance/chronology. 

(This issue is about ordering codelists as whole, https://github.com/open-contracting/standard/issues/1198 is about ordering codes within codelists.)

odscjen:
closes #1405 

- [x] Move codelists to individual pages
- [x] Add a [badge](https://sphinx-design.readthedocs.io/en/latest/badges_buttons.html#badges) below each codelist page title to indicate whether the codelist is open or closed. The badges can link to the sections of the codelist landing page that explain open and closed codelists.
- [x] Order the codelist pages alphabetically in the navigation
- [x] Propose a grouping and ordering for the links from the codelist landing page to the individual codelist pages

Will wait for https://github.com/open-contracting/standard/pull/1660 and https://github.com/open-contracting/standard/pull/1680 to be merged prior to fixing links broken by this PR and then converting it from draft

### DIFF
--- a/docs/schema/codelists.md
+++ b/docs/schema/codelists.md
@@ -2,7 +2,10 @@
 
 Some schema fields refer to codelists, to limit and standardize the possible values of the fields, in order to promote data interoperability.
 
-Codelists can either be open or closed. **Closed codelists** are intended to be comprehensive; for example, the [currency](#currency) codelist covers all currencies in the world. **Open codelists** are intended to be representative, but not comprehensive.
+Codelists can either be open or closed;
+
+* **Closed codelists** are intended to be comprehensive; for example, the [currency](#currency) codelist covers all currencies in the world.
+* **Open codelists** are intended to be representative, but not comprehensive.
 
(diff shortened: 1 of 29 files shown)


odscjen:
Have used the 'primary' badge for closed codelists (blue with white text) and the 'secondary' badge for open codelists (grey with white text). Other colour options are available.

Used `danger` badge (red with white text) for deprecated badge

odscjen:
removed codelists as titles to show all the link locations that will need to be updated in the test output

PR #1739, opened 2025-08-26
Switch from Transifex to Crowdin

### LINKED ISSUE #1600: Remove extra translations from theme.po
Accidentally added in 64cc136558293f566bf35f4903776d2daf741d21

### LINKED ISSUE #1601: Typo: Change "contracting authorities" to "buyers" in Chile framework agreement example
https://standard.open-contracting.org/latest/en/guidance/map/framework_agreements/#example-2-open-framework-agreement-with-multiple-buyers-with-multiple-suppliers-and-with-second-stage-competition

"Contracting authorities" has different meanings in different jurisdictions (and no definition in OCDS). We use the term for the UK example, because we know the meaning of the English term in that context (in law and in practice).

jpmckinney:


jpmckinney:
Hmm, 1.1-dev is too far behind the profile template (e.g. it uses Python 3.8).

I could bring it up to date, but that will create merge conflicts with 1.2-dev – which are resolvable but not worth the effort unless we need to push 1.1-dev to the 1.1 branch.

I'll just close this and update 1.2-dev (which doesn't have a related remote translation project yet).

